PR #151, opened 2014-05-27
Extend / refactor docs

stefankoegl:
I'm currently working on extending and refactoring the docs. Quite some changes have already gone into the first batch, but there is more to come.

Please have a look at the changes, and let me know if you find any issues, or if you want me to cover any specific topics.


KennethNielsen:
Hi @stefankoegl

I was wondering what the status is of this. The reason I ask is that the Napoleon enabling commit went into this PR and I would really like to start moving all of our doc strings to Napoleon. So I was wondering whether you see this PR being prepared for pull in the near future or if we should move the Napoleon stuff out into a PR of its own?

stefankoegl:
I didn't really find enough time to work on this (and I won't have time for at least another 4 weeks), so feel free to disregard it for now - I will just catch up with whatever the current state is then.

KennethNielsen:
Ok, thanks.

lawrenceakka:
I think we should put Napoleon in now. IT works fine alongside our existing docs - see #282

KennethNielsen:
@lawrenceakka agreed, I already +1'ed it

KennethNielsen:
This has been superseded and included in #406, which is now merged. Thanks for all the work.

PR #225, opened 2014-09-14
Add is_playing_tv, is_playing_radio, is_playing_line_in properties with tests

dajobe:
Also update the docs to note that the switch/play tv are for playbar.


robwebset:
Hi @dajobe - did you manage to do the test you detailed here:
https://github.com/SoCo/SoCo/pull/225#issuecomment-57056437
To confirm it is different from the "Line-In"

If it is indeed different - then it would be good to get this merged in

dajobe:
@robwebset the playbar only has one input and intended for TV, so Line-In == TV.  I'm not sure what to do with this patch. 

"PLAYBAR connects to your TV using a single optical cable and plays all sources connected to the TV, including cable boxes and game consoles." -- http://www.sonos.com/shop/products/playbar

See http://www.sonos.com/shop/i/products/playbar.detail_big.jpg

robwebset:
@dajobe The question is - does your code is_playing_tv do the same as the check for line-in already in SoCo?

And if you have something that has a line-in - does this is_playing_tv method return "true" when it is using line in?

dajobe:
@robwebset I can check out the playbar part of it at home, but I don't have any other sonos device with line-in to check the opposite.

KennethNielsen:
If no one else steps up, I can probably do the test here. In any case, I don't think it should be pulled before it has been tested (and possibly fixed), because the name of the property would then be misleading.

dajobe:
@robwebset I was looking for "the check for line-in already in SoCo?" you mentioned but I can't see it.  There is no `is_playing_linein` method - maybe we should add that too?

I can see `switch_to_line_in` that uses `x-rincon-stream:` with `CurrentURI` arg to `avTransport.SetAVTransportURI`

Looks like if that's the difference then TV and Line-In are distinguishable since my patch here looks for TrackURI starting 'x-sonos-htastream:'.  I'm curious if my code would make TrackURI return `x-rincon-stream...` if playing line in.  I can't test that since I only have a Playbar.  @KennethNielsen can you help ?

robwebset:
@dajobe Sorry, yes you are correct - got confused - the question I meant to ask - is this the same as line-in, so will the same command check TV and Line in - in which case it should have a generic name to cover both cases - if they are different checks - then they should have different names :)

KennethNielsen:
Agree with @robwebset about that the names should reflect the functionality.

@dajobe Yes, I may be able to test it this evening or over the weekend. Can you supply the test code you wich me to run on this branch?

dajobe:
@KennethNielsen I added two new is_playing methods we can try out but more specifically see if is_playing_line_in and is_playing_tv work correctly.  I can only test playbar (TV in).

robwebset:
So I think once the is_playing_lin_in has been verified (unfortunately I can't - I don't have anything with a line-in) we can get this into the 0.11 release https://github.com/SoCo/SoCo/issues/278

+1

dajobe:
When playing tv through my playbar with this code:

    In [18]: coord.is_playing_tv
    Out[18]: True
    
    In [19]: coord.is_playing_radio
    Out[19]: False
    
    In [20]: coord.is_playing_line_in
    Out[20]: False

KennethNielsen:
@dajobe I'm very sorry for not reporting back after having promised it. I've been busy with my 5 months old, still that's no excuse for not writing here that I was too busy.

In any case, I finally found a few moments to test the branch. I played something on line-in on my play 5 and it works the way it is supposed to:

```python
>>> kontor.is_playing_tv
False
>>> kontor.is_playing_line_in
True
>>> kontor.is_playing_radio
False
```

KennethNielsen:
@robwebset Wow, this has been laying around for a while. I guess this is ready for merge after the test I did, correct?

KennethNielsen:
@dajobe ^^

robwebset:
+1 from me - think everything is OK to go in

dajobe:
@KennethNielsen thanks.  this took a while :)

KennethNielsen:
@dajobe You agree it is good to go in in current form corretc?

dajobe:
@KennethNielsen yes. :shipit:

KennethNielsen:
I'll merge. Please drop a line about the change in the issue with information for the release notes #311

KennethNielsen:
@dajobe never mind my comment about added line to the release notes, it seems that @stefankoegl already did it.

PR #242, opened 2014-10-10
scan all interfaces when discovering

trollkarlen:
Signed-off-by: Robert Marklund <robbelibobban@gmail.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 requests
+netifaces
--- a/soco/__init__.py
+++ b/soco/__init__.py
@@ -14,13 +14,14 @@ __website__ = 'https://github.com/SoCo/SoCo'
 __license__ = 'MIT License'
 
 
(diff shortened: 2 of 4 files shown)


robwebset:
I just thought I'd check where this one is going?  I'd like to see this get merged if possible - from my understanding it just needs updating for https://github.com/SoCo/SoCo/pull/242#issuecomment-64959163 and conflicts resolved.

@trollkarlen are you OK with updating this.

I think we should looks at getting this in soon - therwise it might miss the boat again for 0.11

lawrenceakka:
I have something ready to go on this. Just awaiting #273 which moves discover to its own module. I have removed the get interfaces code however, so that it will be up to the user to provide the interfaces to use

KennethNielsen:
I think @lawrenceakka arguments makes sense. So I would also vote for a version that does not include the interface discovery. Whether this PR should be modified or it can be included in another, I have no strong opinions on.

Issue #99, opened 2014-04-11
Text to speech attempt and metadata


rodhall:
All - similar to aProgrammer, I think I am running into metadata problems on a little project I have been working on using SoCo.  google.translate has an API that will return a TTS conversion as an MP3 for anything you throw at it.  My initial little test program is as follows:

import requests
from soco import SoCo
import sys
import os
lr_sonos = SoCo('192.168.0.148')
tts_string = input('Enter something to say:  ')
speech_uri = 'http://translate.google.com/translate_tts?q='+tts_string
lr_sonos.play_uri(uri=speech_uri,meta='')
print('Done!')

However, this isn't working and it seems like the problems are mainly associated with metadata (though I am not completely sure).  Error return is as follows:

Enter something to say:  hello
Traceback (most recent call last):
  File "/Users/rodhall/Dropbox/Python/speaksonos.py", line 9, in <module>
    lr_sonos.play_uri(uri=speech_uri,meta='')
  File "/Users/rodhall/Dropbox/Python/soco/core.py", line 253, in play_uri
    ('CurrentURIMetaData', meta)
  File "/Users/rodhall/Dropbox/Python/soco/services.py", line 153, in _dispatcher
    return self.send_command(action, args)
  File "/Users/rodhall/Dropbox/Python/soco/services.py", line 316, in send_command
    self.handle_upnp_error(response.text)
  File "/Users/rodhall/Dropbox/Python/soco/services.py", line 377, in handle_upnp_error
    error_xml=xml_error
soco.exceptions.SoCoUPnPException: UPnP Error 714 received: Illegal MIME-Type from 192.168.0.148

I just think it would be super cool to get my Sonos to talk when various things happen (cameras detect motion, doorbell, etc etc.) however I am a newbie to SoCo and need some advice.

kylewill:
I'm working on a similar project and trying @robwebset code. I made some updates for Python 3 but I get an error from the Sonos controller "Unable to play station Speech - station not found". 

Anyone have this working?

robwebset:
There was a change in the google link a while ago:

trans_URI = "x-rincon-mp3radio://translate.google.com/translate_tts?tl=%s&q=%s&client=t" % (glang, msg)

New location of full file:

https://github.com/robwebset/script.sonos/blob/master/resources/lib/speech.py

Rob

leanto:
I just tried this and, as far as I can tell, the Google API isn't working anymore.

leanto:
[...]

PR #267, opened 2014-12-27
Initial Commit:  Plugin works for my single speaker system. I attempted ...

jlmcgehee21:
I attempted to build this for multiple speaker systems, but I have no way to test/troubleshoot it.

I'm not sure if this fits your programming style, but I tried my best.  I would love for someone with multiple speakers to test/add to this.  It may be a little while before I add speakers to my system.

The main purpose of this for me is to link it with my DIY home automation system.  I wanted to contribute to what I feel is a great project instead of just branching out on my own.

-Jeff McGehee

### DIFF
--- a/soco/plugins/talk.py
+++ b/soco/plugins/talk.py
@@ -52,8 +52,11 @@ class TalkerPlugin(SoCoPlugin):
         if zoneNames:
             assert type(zoneNames) == list and all([zone in nameList for zone in zoneNames]), \
                 'Speaker object must be instantiated with a list of existing zone names on your network'
+            speakingSoCos = [zone for zone in zoneList if zone.player_name in zoneNames]
+        else:
+            speakingSoCos = zoneList
 
-        speakingSoCos = [zone for zone in zoneList if zone.player_name in zoneNames]
+
 
         self.masterSoCo = speakingSoCos[0]
         speakingSoCos.pop(0)
--- a/soco/plugins/talk.py
+++ b/soco/plugins/talk.py
@@ -1,0 +1,211 @@
+import sys
+import re
+import urllib, urllib2
+import time
+
+
+from ..plugins import SoCoPlugin
+
+__all__ = ['Talk']
+
+
+
+class TalkerPlugin(SoCoPlugin):
+    """
+        The main use of this plugin is to make your Sonos system speak text.  It works by sending a request to the
+        Google Text To Speech service, downloading an MP3 from the service, then playing the MP3 on the desired Sonos
+        players.  It will pause and resume playback properly if you are listening to music at the time the message is
+        sent.
+
+        SETUP REQUIREMENTS:  You must add the path to the Google Text To Speech MP3 to your Sonos music library in order
+        to obtain the URI for that file.  Once this is done, you can find the URI "get_music_library_information()"
+        method in the soco package.
+    """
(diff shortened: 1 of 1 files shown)


DPH:
Version 0.10 of SoCo has all the stuff you need (snapshot, whole lot of fixes etc.)
It can be installed as any other Python package: https://pypi.python.org/pypi/soco/

The speaking bit needs to  either be in you own code, based on the examples out there, or be made into a plugin for general use. 

Hope this helps, Cheers David

KennethNielsen:
Hi @jlmcgehee21 and thanks for putting the stuff out here.

I'm a little in doubt about the status here, and the text-to-speech stuff is not my strong suit. Does this plugin represent functionality not already in SoCo (@jlmcgehee21, @robwebset, @DPH). At first glance I would say that it does, since it seems most of the existing text-to-speech stuff has been external scripts and this is made as a proper plugin. It might of course be very useful for @jlmcgehee21 to have a look at the code @robwebset published in #99, so if you want to continue working on the plugin and make a PR later, we can also do that.

I just wanted to check before I started reviewing.

jlmcgehee21:
After some review of the code in #99 and more testing with my pull request,
I think the best solution would be to modify #99 to function as a proper
plugin.  The use of the Snapshot class (which I didn't even realize existed
until I got some responses to my pull request) is much better than what I
tried to hack together.

My plan is to implement #99 as a plugin when I find some free time and
submit a new pull request then.  I think I'll go ahead and close this one.
Thanks for the awesome work you guys are doing!

KennethNielsen:
That sounds like a great plan.

One thing that might help you, with your next PR. We tend to stay pretty close with the python programming style defined in pep8 and by pylint, and I noticed quite a few style problems when I glanced this PR, which you would have to fix to get a PR pulled. So I would advice you to run pylint and pep8 checks on your code as you develop, to get to know the guide lines and to not have to worry about that in the review. You should note though, that we don't follow pylint **religiously**. Checks can be disabled, and there can be valid reasons to do so, but not very many ;)

Let us know if you run into problems running pylint and pep8 checks, that you can't information on with google.

jlmcgehee21:
Thanks for the tips. I have been developing with PyCharm which makes pep8
based suggestions.  I guess it's time that I stop ignoring them ;)

I do scientific coding professionally, so for the most part, any given
piece of code only needs to live long enough for me to get an answer.  It's
good for me to see/learn some proper developing techniques.

I saw that the existing plugins are coded with a heavy use of decorators.
Is that preferred styling for SoCo as well?  Just want to make any future
pull requests as painless as possible.

KennethNielsen:
I have a similar background to you (scientific and not trained as a software developer ;)), so feel free to take it with a grain of salt when regarding any advice that I give as "proper developing techniques" :D

I'm not sure which decorators you are referring to ``@staticmethod`` and ``@classmethod`` serve a particular purpose, but most likely what you have noticed are ``@property`` decorators. No, we have no preference for using decorators especially much, we use them when they serve a purpose. The ``@property`` for instance, is useful if you have a method, that essentially acts like a property (i.e. essentially just for reading out a value of something) but where the value needs to be calculated. I think probably the best example of the use of ``@property`` is the fictional ``Circle`` class, where you store the radius internally, but implement the diameter as a property. So, to put it shortly, no, we just use them when it makes sense (the cases above or any other time you need to wrap a functions in a start and end). Also, one should be careful about not overusing decorators, because they 'hide' code from plain sight and therefore can make it more difficult to read.

PR #269, opened 2014-12-31
make the value of the IP_MULTICAST_TTL option one byte long

gutworth:
The multicast TTL may have value in [0, 255], which fits in one byte. Linux allows an 32-bit integer to be passed if its value is less than 256. Some `setsockopt` implementations (like OpenBSD), though, are less tolerant and require the input to actually be one byte long.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -14,6 +14,7 @@ import re
 import itertools
 import requests
 import time
+import structt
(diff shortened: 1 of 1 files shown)


lawrenceakka:
Thanks very much for this.

PR #273, opened 2015-01-13
New year tidy

lawrenceakka:
Tidying up some odd and ends.

- remove code which has been deprecated for more than three releases

- move discover to its own module.  Note that `from soco import discover` will work just as it did before.

- slightly saner debug logging (clarified a few messages, removed pointless ones etc)

### DIFF
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -1,0 +1,99 @@
+# -*- coding: utf-8 -*-
+""" The core module contains the SoCo class that implements
+the main entry to the SoCo functionality
+"""
+
+from __future__ import unicode_literals
+
+import socket
+import select
+from textwrap import dedent
+import time
+import struct
+
+from soco import config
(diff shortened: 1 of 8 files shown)


soco/discovery.py
```diff
@@ -0,0 +1,99 @@
+# -*- coding: utf-8 -*-
+""" The core module contains the SoCo class that implements
```
KennethNielsen: This one should probably be updated

KennethNielsen:
Reviewed with a single inline comment. I have not tested it, but it seems fairly simple, so besides the small comment +1

lawrenceakka:
I will commit this shortly because there are a few other things waiting for it (including some bug fixes). Let me know if any last minute comments in addition to @KennethNielsen's which I have fixed.

stefankoegl:
Nice! +1

PR #275, opened 2015-01-14
Event timestamp

rrowe629:
I have a Play:1 that I would like to be able to control without the controller app - I don't always have my phone on me. I've built a python app that does just that, but had to make a slight change to <b>events.py</b> to accomplish it. Basically, the Play:1 behaves normally in all respects except when a "trigger" event occurs. I've defined the trigger to be a volume change in one direction, followed by a change in the opposite direction - within two seconds. Outside of that window, the volume buttons behave normally. Once the trigger event occurs, I use SoCo and Google TTS to report the local temperature and then speak the names of the favorite radio stations. The user uses the "+" button to select and the "-" button to cancel.

The change I had to make was to add a timestamp to events. This allows me to determine if the volume change events are within the two second window or not. Here's the change I made to <b>events.py</b> - it appears to be as simple as adding a timestamp to the dictionary for the event as follows:

```python
         # overridden, and will not allow direct setting of attributes
         self.__dict__['sid'] = sid
         self.__dict__['seq'] = seq
+        self.__dict__['timestamp'] = time.time()    #Local Time the Event object was created
         self.__dict__['service'] = service
         self.__dict__['variables'] = variables if variables is not None else {}
```
I've been using this for a while now and it works just fine. I'm effectively able to control the Play:1 like a desktop radio and, when I want to, join it to the other Sonos system in my house. There may be a better way to do this - but I believe it would be useful to include a timestamp on events in any case.

P.S. New to GitHub - all three commits essentially do the same thing.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -178,6 +178,7 @@ class Event(object):
         # overridden, and will not allow direct setting of attributes
         self.__dict__['sid'] = sid
         self.__dict__['seq'] = seq
+	self.__dict__['timestamp'] = time.time()
         self.__dict__['service'] = service
         self.__dict__['variables'] = variables if variables is not None else {}
 
--- a/soco/events.py
+++ b/soco/events.py
@@ -178,7 +178,7 @@ class Event(object):
         # overridden, and will not allow direct setting of attributes
         self.__dict__['sid'] = sid
         self.__dict__['seq'] = seq
-	self.__dict__['timestamp'] = time.time()
+	self.__dict__['timestamp'] = time.time()    #Local Time the Event was created
         self.__dict__['service'] = service
         self.__dict__['variables'] = variables if variables is not None else {}
 
--- a/soco/events.py
+++ b/soco/events.py
@@ -178,7 +178,7 @@ class Event(object):
         # overridden, and will not allow direct setting of attributes
         self.__dict__['sid'] = sid
         self.__dict__['seq'] = seq
-	self.__dict__['timestamp'] = time.time()    #Local Time the Event was created
+	self.__dict__['timestamp'] = time.time()    #Local Time the Event object was created
         self.__dict__['service'] = service
         self.__dict__['variables'] = variables if variables is not None else {}
 


dajobe:
FYI the test failure reports code style problems on the line you changed:

    soco/events.py:181:1: E101 indentation contains mixed spaces and tabs
    soco/events.py:181:1: W191 indentation contains tabs
    soco/events.py:181:46: E262 inline comment should start with '# '
    soco/events.py:181:80: E501 line too long (85 > 79 characters)
    soco/events.py:182:1: E101 indentation contains mixed spaces and tabs

So if you replace the tab with spaces and fix the comment (remove it?) you should be get a pass.

lawrenceakka:
Thanks for this. It's a great idea.

Rather than requiring you to fix formatting etc (and you didn't update the tests or the docs :wink:) I have added it onto the end of #273 which also deals with events. It should get committed in the next couple of days.

When do we get to see your code?  It sounds very interesting.

Lawrence

rrowe629:
Thanks, Lawrence. As you might have guessed, a bit new to how all of this works.

Happy to post the code - what is the correct place to do that?

rrowe629:
Thanks for the tips. Lawrence was kind enough to fold this into #273 for me.

From:  Dave Beckett <notifications@github.com>
Reply-To:  SoCo/SoCo
<reply+0022c34a8eab8a60c1a2a82dd611c4255bcd7c4b74f280ac92cf0000000110ce68259
2a169ce033d2f48@reply.github.com>
Date:  Wednesday, January 14, 2015 at 11:23 AM
To:  SoCo/SoCo <SoCo@noreply.github.com>
Cc:  Roger Rowe <rw.rowe@cox.net>
Subject:  Re: [SoCo] Event timestamp (#275)


FYI the test failure reports code style problems on the line you changed:
soco/events.py:181:1: E101 indentation contains mixed spaces and tabs
soco/events.py:181:1: W191 indentation contains tabs
soco/events.py:181:46: E262 inline comment should start with '# '
soco/events.py:181:80: E501 line too long (85 > 79 characters)
soco/events.py:182:1: E101 indentation contains mixed spaces and tabs
So if you replace the tab with spaces and fix the comment (remove it?) you
should be get a pass.

‹
Reply to this email directly or view it on GitHub
<https://github.com/SoCo/SoCo/pull/275#issuecomment-69954086> .

lawrenceakka:
> Happy to post the code - what is the correct place to do that?

Your github account!  Or create a gist: https://gist.github.com

lawrenceakka:
This has now been merged as part of #273 

I'll close it for now.  Thanks.

Issue #278, opened 2015-01-18
Release 0.11


stefankoegl:
It has been 2 month since the release of 0.10 already, so we should start thinking about the next release.

Is there anything specific that should go into 0.11?

stefankoegl:
So, I'll try again to get some opinions on a 0.11 release. What do you think should go into the release? When approximately should we schedule a release?

KennethNielsen:
@stefankoegl There are a few bug fixes that I think it would be nice to get out there. So if we are willing to do just that we could release now.

However, I do think the work in #296 would be worth pursuing and try to get in the next release. The only thing is that it is not done. I need feedback from people on whether they think it is a good idea or not. Provided I get the feedback, I think I should find time to finish it within the next couple of weeks and we could then schedule a release there.

On another note, I don't think we need to wait for #262, even though it would be nice. It is my feeling that it still requires more work and that @lawrenceakka is tied up with real life.

stefankoegl:
OK, then let's wait for #296. I should find time to do a review in the next days.

KennethNielsen:
@stefankoegl It's actually more a "yes" or "no" on the general idea rather than a formal review, so should not take so long

stefankoegl:
OK, it seems we have everything for the release.. If there are no objections I'd prepare the release most likely tomorrow afternoon/evening (UTC).

KennethNielsen:
@stefankoegl I think there are a few more issues marked for 0.11:

 * #225 Which I think is good to go, I'm just waiting for the go a head from @dajobe 
 * #276 Which I don't think there is any progress on, so probably just postpone it.

stefankoegl:
I have created the proposed release notes in #311, already including #225.

If there are no objections I'd publish the release after #225 is merged.

KennethNielsen:
Sounds great.

stefankoegl:
I have just released SoCo 0.11!

* [PyPI](https://pypi.python.org/pypi/soco/0.11)
* [GitHub release](https://github.com/SoCo/SoCo/releases/tag/v0.11)
* [Release notes](http://docs.python-soco.com/en/latest/releases/0.11.html)
* [Change Log](https://github.com/SoCo/SoCo/compare/v0.10...v0.11)

Thanks everyone for their contributions! @simonalpha @trollkarlen @lawrenceakka @dajobe @robwebset @gutworth @poirier @michaelotto @hildroid @KennethNielsen

KennethNielsen:
[...]

PR #283, opened 2015-01-24
Try a little harder to find a title for a stream

poirier:
When streaming from a Tune In Radio station, anyway,
there might be nothing in the streamcontent but
still a title element.


KennethNielsen:
Seems reasonable. There are a few lines that are too long, which triggers a pylint error. Can you please fix those up?

poirier:
Lines shortened. (Sorry, used to 100 char lines.)

robwebset:
I think this looks good - worth getting in before 0.11 is released https://github.com/SoCo/SoCo/issues/278

+1

PR #286, opened 2015-02-01
Add API for music library updates

poirier:


soco/core.py
```diff
@@ -1811,6 +1813,23 @@ def get_tracks_for_album(self, artist, album, full_album_art_uri=False):
...
+        :returns: True if the music library is in the process of being updated
+        :rtype: bool
+        """
+        result = self.contentDirectory.GetShareIndexInProgress()
+        return result['IsIndexing'] != '0'
+
+    def start_library_update(self):
+        """Start an update of the music library."""
+        return self.contentDirectory.RefreshShareIndex([
+            ('AlbumArtistDisplayOption', ''),
```
lawrenceakka: Perhaps expose this as a parameter, with a default value?  Does this have something to do with the "compilation Albums" settings from the official controller? Any idea what values it can take?

lawrenceakka:
Dan,

+1 from me

One small comment inline. If you feel like doing some digging into the purpose of the AlbumArtistDisplayOption parameter, please do!

soco/core.py
poirier: I experimented with GetAlbumArtistDisplayOption and changing the album artist compilation setting and have updated this pull request with how I think this parameter works.

lawrenceakka:
Thanks Dan

PR #288, opened 2015-02-04
Configurable event listener port

michaelotto:
For cases where two independent processes on the same machine both want to listen to SONOS events.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -19,6 +19,7 @@ import atexit
 
 import requests
 
+import config
 from .compat import (SimpleHTTPRequestHandler, urlopen, URLError, socketserver,
                      Queue,)
 from .xml import XML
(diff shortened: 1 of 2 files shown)


soco/events.py
```diff
@@ -576,6 +575,13 @@ def time_left(self):
             time_left = self.timeout-(time.time()-self._timestamp)
             return time_left if time_left > 0 else 0
 
+# if you want to use a different port than 1400, set
+# soco.events.event_listener_port accordingly after importing but before
+# subscribing to an event
+
+# pylint: disable=C0103
+event_listener_port = 1400
```
poirier: Would it work to make this an optional parameter on `EventServerThread.start()` instead of a module global?

lawrenceakka: If you are going to have a global, you should put something in the soco.config module, rather than here. I agree with @poirier however. It is better just to add an extra parameter to the `start` method, with a default of 1400.

michaelotto: Well, the `start` method is not called from the outside, but from the inside when a user subscribes to an event. How would you want to pass the parameter for the port from the outside then? Should I make the port a `subscribe` parameter? I don't think that's better.

I agree moving the port config to soco.config is better though.

lawrenceakka: > Should I make the port a subscribe parameter? I don't think that's better.

I agree. Don't do that!  I forgot that `start` was not called from the outside (even though I wrote it!)

> I agree moving the port config to soco.config is better though.

Yes please.

michaelotto:
Taking care of that tomorrow.

lawrenceakka:
Thanks Michael.

PR #289, opened 2015-02-05
Add a more aggressive discovery method

robwebset:
Having messed about with some auto-discovery code, I found that if you have multiple network devices, the Sonos discover code can be a bit hit and miss.  This is mostly the case if you have something like VMWare or a wired and wireless network card, and don't set up your wireless card to connect to a network.

This pull request adds a new method "discovering" that will work better in these cases.

To see it on your system, if you have an ethernet and wireless card - make sure one is not connected to a live network - then run the old discover - there is a 50/50 chance that it will not pick up the correct card - then after the discover fails - try the discovering method.

On my 2 different systems that were showing the issue of not finding the sonos devices, it started working :)

If anyone wants to hack this about a bit - then please feel free to take a copy, alter it, then do a new pull request.

Thanks

Rob

### DIFF
--- a/soco/__init__.py
+++ b/soco/__init__.py
@@ -15,13 +15,14 @@ __license__ = 'MIT License'
 
 
 from .core import SoCo
-from .discovery import discover
+from .discovery import discover, discovering
 from .exceptions import SoCoException, UnknownSoCoException
 
 # You really should not `import *` - it is poor practice
 # but if you do, here is what you get:
 __all__ = [
     'discover',
+    'discovering',
     'SoCo',
     'SoCoException',
     'UnknownSoCoException',
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -16,6 +16,58 @@ from .utils import really_utf8
 _LOG = logging.getLogger(__name__)
 
 
+def discovering(timeout=1, include_invisible=False):
+    """ Discover Sonos zones on the local network.
+
+    This is a more aggressive search than the discover method, using multiple
+    attempts to check different network addresses
+
+    Args:
+        timeout (int): block for this many seconds, at most. Default 1
+        include_invisible (bool): include invisible zones in the return set.
+            Default False
+
+    Returns:
+        (set): a set of SoCo instances, one for each zone found, or else None.
+    """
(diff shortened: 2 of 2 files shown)


poirier:
How does this compare to the interfaces that https://pypi.python.org/pypi/netifaces comes up with?

robwebset:
netifaces is a platform dependant python package - it needs things like windows binaries - this approach will not require netifaces, as it will just use the socket package.  It may not be quire as clean - but does seem to work.

lawrenceakka:
Hi Rob @robwebset,

I suppose I'm coming round to the idea of incorporating a better way of detecting the right interface if we need to (though I thought that netifaces was not the right way to go).  A few comments:

1. Are you sure that it is an interface problem, not just a question of raising of the timeout.  Actually I wonder if we might raise the default timeout from 1 sec to, say 5 secs, anyway.

2. Between the three various methods you have some up with, which works best? Can you try them all with higher timeouts to see whether one is consistently a winner on your systems (without timeouts getting in the way)? Is the third one (calling out to an external site) really necessary?

3. Your code might end up running for 3 times the max timeout specified. Better just to create several sockets and put them all in a single select loop with a single overall timeout, so that they all run in parallel?

Lawrence

robwebset:
Hi @lawrenceakka 

1) I did originally try upping the time-out - I think I got as high as 20 seconds - but still nothing, I actually use 5 seconds when I call it.

2) I have done a fair bit of hunting around, and these are the 3 different methods that seem to get mentioned - I ordered them in the "Most likely to work" order - based on what people have reported - but I believe it will all change based on the platform.

3) A valid point about the time-out, but I think we want to avoid adding them all into one call, especially the last that resorts at going to the web! (Maybe just comment in the API, that the time-out is per-request, and there can be as many as 4 requests)

Rob

pmatos:
Why has this not been merged yet?

robwebset:
I believe because @lawrenceakka re-implemented it to make it "Less Aggressive" - #292 - so there is some discussion as to which one to use, the original or the re-write.

pmatos:
@robwebset OK, thanks. I will take a look.

PR #292, opened 2015-02-08
More aggresssive discovery - thanks @robwebset. Closes #289

lawrenceakka:
Implementation of ideas in #289.  This uses the default method, and two other methods of finding the relevant local address (if they result in anything different). I have not included the suggestion for calling out to sonos.com and then checking the local socket address because (1) it takes too long, and (2) I am far from certain that it is likely to produce any address different from the other methods.  Let's wait and see if this is generally good enough.

Comments please, especially from @robwebset 

Lawrence

### DIFF
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -15,8 +15,10 @@ from .utils import really_utf8
 
 _LOG = logging.getLogger(__name__)
 
+# pylint: disable=too-many-locals, too-many-branches
 
-def discover(timeout=1, include_invisible=False, interface_addr=None):
+
+def discover(timeout=5, include_invisible=False, interface_addr=None):
     """ Discover Sonos zones on the local network.
 
     Return an set of SoCo instances for each zone found.
@@ -54,6 +56,23 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
 
     """
 
+    def create_socket(interface_addr=None):
+        """ A helper function for creating a socket for discover purposes.
+
+        Create and return a socket with appropriate options set for multicast.
+        """
+
+        _sock = socket.socket(
+            socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+        # UPnP v1.0 requires a TTL of 4
+        _sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL,
+                         struct.pack("B", 4))
+        if interface_addr is not None:
+            _sock.setsockopt(
+                socket.IPPROTO_IP, socket.IP_MULTICAST_IF,
+                socket.inet_aton(interface_addr))
+        return _sock
+
     # pylint: disable=invalid-name
     PLAYER_SEARCH = dedent("""\
         M-SEARCH * HTTP/1.1
@@ -65,11 +84,7 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
     MCAST_GRP = "239.255.255.250"
     MCAST_PORT = 1900
 
-    _sock = socket.socket(
-        socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
-    # UPnP v1.0 requires a TTL of 4
-    _sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL,
-                     struct.pack("B", 4))
+    _sockets = []
     # Use the specified interface, if any
     if interface_addr is not None:
         try:
@@ -77,14 +92,34 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
         except socket.error:
             raise ValueError("{0} is not a valid IP address string".format(
                 interface_addr))
-        _sock.setsockopt(
-            socket.IPPROTO_IP, socket.IP_MULTICAST_IF, address)
+        _sockets[0] = create_socket(interface_addr)
+    else:
+        # Find the local network address using a couple of different methods.
+        # Create a socket for each unique address found, and one for the
+        # default multicast address
+        addresses = set()
+        try:
+            addresses.add(socket.gethostbyname(socket.gethostname()))
+        except socket.error:
+            pass
+        try:
+            addresses.add(socket.gethostbyname(socket.getfqdn()))
+        except socket.error:
+            pass
+        for address in addresses:
+            _sockets.append(create_socket(address))
+        # Add a socket using the system default address
+        _sockets.append(create_socket(None))
 
-    # Send a few times. UDP is unreliable
-    _LOG.info("Sending discovery packets")
-    _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
-    _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
-    _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
+    _LOG.info(
+        "Sending discovery packets on default interface and %s",
+        list(addresses)
+    )
+    for _sock in _sockets:
+        # Send a few times. UDP is unreliable
+        _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
+        _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
+        _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
 
     t0 = time.time()
     while True:
@@ -101,7 +136,7 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
         # The timeout of the select call is set to be no greater than
         # 100ms, so as not to exceed (too much) the required timeout
         # in case the loop is executed more than once.
-        response, _, _ = select.select([_sock], [], [], min(timeout, 0.1))
+        response, _, _ = select.select(_sockets, [], [], min(timeout, 0.1))
 
         # Only Zone Players should respond, given the value of ST in the
         # PLAYER_SEARCH message. However, to prevent misbehaved devices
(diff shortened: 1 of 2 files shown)


soco/discovery.py
```diff
@@ -54,6 +56,23 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
...
+    def create_socket(interface_addr=None):
+        """ A helper function for creating a socket for discover purposes.
+
+        Create and return a socket with appropriate options set for multicast.
+        """
+
+        _sock = socket.socket(
+            socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
+        # UPnP v1.0 requires a TTL of 4
+        _sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_TTL,
```
robwebset: Would this be better as an if/else - I guess is interface_addr is set, then you are over-writing a value you have just set?

soco/discovery.py
```diff
@@ -77,14 +92,34 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
...
+        except socket.error:
+            pass
+        try:
+            addresses.add(socket.gethostbyname(socket.getfqdn()))
+        except socket.error:
+            pass
+        for address in addresses:
+            _sockets.append(create_socket(address))
+        # Add a socket using the system default address
+        _sockets.append(create_socket(None))
```
robwebset: Do you actually have to pass in None here - as it has a default value when no address is given

robwebset: With the way that this is appended - could we end up with one address being duplicated, as it is no longer adding it to a set, but an array

soco/discovery.py
```diff
@@ -123,18 +158,19 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
...
+                _LOG.debug(
+                    'Received discovery response from %s: "%s"', addr, data
+                )
+                if b"Sonos" in data:
+                    # Now we have an IP, we can build a SoCo instance and query
+                    # that player for the topology to find the other players.
+                    # It is much more efficient to rely upon the Zone
+                    # Player's ability to find the others, than to wait for
+                    # query responses from them ourselves.
+                    zone = config.SOCO_CLASS(addr[0])
```
robwebset: As we have possibly sent out multiple broadcasts on several different network devices, isn't it possible we could have received multiple responses from the same speaker (i.e. the source computer has multiple network cards connected to the same network).  In which case, shouldn't we then enusre we only add each speaker once.

soco/discovery.py
```diff
@@ -77,14 +92,34 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
...
+        for address in addresses:
+            _sockets.append(create_socket(address))
+        # Add a socket using the system default address
+        _sockets.append(create_socket(None))
+
+    _LOG.info(
+        "Sending discovery packets on default interface and %s",
+        list(addresses)
+    )
+    for _sock in _sockets:
```
robwebset: Now we can have multiple addresses, could we loop slightly differently, so we send one message to each interface in turn, i.e.
```
for 1..3:
   for _sock in _sockets:
      _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
```

soco/discovery.py
```diff
@@ -123,18 +158,19 @@ def discover(timeout=1, include_invisible=False, interface_addr=None):
...
+                )
+                if b"Sonos" in data:
+                    # Now we have an IP, we can build a SoCo instance and query
+                    # that player for the topology to find the other players.
+                    # It is much more efficient to rely upon the Zone
+                    # Player's ability to find the others, than to wait for
+                    # query responses from them ourselves.
+                    zone = config.SOCO_CLASS(addr[0])
+                    if include_invisible:
+                        return zone.all_zones
```
robwebset: With this loop, aren't we actually only matching the first network interface that returns - should we wait for all of them?
(I know my code stopped on the first find - but it also only sent them out one at a time - not as batch together)

robwebset:
Hi @lawrenceakka 

I have added some comments against each line - hope that is OK.

Can I suggest you add the method that does the call to "sonos.com" at the end if no other speakers are found.  In my experience people will only use the discover if they believe they have speakers on their network - so it would be worth doing absolutely everything we can to try and find it.

Thanks
Rob

soco/discovery.py
lawrenceakka: I don't understand, sorry.  What is being overwritten, by what?

soco/discovery.py
lawrenceakka: > Do you actually have to pass in None here - as it has a default value when no address is given

No. But I think it makes it clear

lawrenceakka: > With the way that this is appended - could we end up with one address being duplicated, as it is no longer adding it to a set, but an array

You are right. Packets may end up being sent out on the same interface twice, since we do not know (and cannot tell) what the system default address is, so it may be the same as one already in the list. There is no obvious (to me) way round this. But I don't see that it matters at all. It just means that one interface gets more packets than another.

soco/discovery.py
lawrenceakka: No, for two reasons:

1. The select call will return as soon as the very first response, on whatever interface, is received. So the rest of the code will never see more than one valid response, and 

2. Because of the way the SoCo class is written, no matter how many times you try to create one with the same IP address, you will only end up with one.

soco/discovery.py
lawrenceakka: We could ... but what difference would it make (except that you version is slightly slower ;) )

soco/discovery.py
lawrenceakka: Yes. That's the whole point. Why do we want to wait for any others? As soon as we have one valid response from a network interface, we have enough info to build the whole sonos network.

lawrenceakka:
Hi Rob,

> I have added some comments against each line - hope that is OK.

Of course.

> Can I suggest you add the method that does the call to "sonos.com" at the end if no other speakers are found. In my experience people will only use the discover if they believe they have speakers on their network - so it would be worth doing absolutely everything we can to try and find it.

The point is that I think we are doing everything we can. I just don't see why having the call to sonos.com will add anything to the other methods, except a DNS request, and a long interval (in computer terms) whilst we wait for a response. Do you have any examples where this method succeeds whilst the other three fail?

Lawrence

soco/discovery.py
robwebset: You call _sock.setsockopt twice if the interface_addr is set - an if/else would only set it once.

soco/discovery.py
robwebset: https://github.com/SoCo/SoCo/pull/292#discussion_r24344960

In which case you should remove the default value from the method - if it's there you should use it, if you don't think the default allows "calls to it to be clear" - then it shouldn't really have a default value.

soco/discovery.py
robwebset: Better coding practive - any place where you copy and paste code multiple times is bad - you should always loop a set number of times - also allows it to be configurable later.

soco/discovery.py
robwebset: If they went out to different networks then you could get different speakers on each network, so you should wait for all the replies.

soco/discovery.py
robwebset: for (1) see later - I don't think you should be returning when you get the first reply.

robwebset:
Yes, I notice you rejected every single one of the comments!

When I did my original searching - it was one of the things suggested to support all platforms, doing the call to a web page seems to set the socket up when all other things have failed - I know it takes longer, this is why I suggested doing it as a last resort - which is how it was implemented in the pull I submitted.

I see no harm in this, as at that point no speakers have been found - so we should try everything.

Rob

soco/discovery.py
lawrenceakka: But I need to call it twice! The first call sets the value of IP_MULTICAST_TTL, and the second sets IP_MULTICAST_IF.

soco/discovery.py
lawrenceakka: > in which case you should remove the default value from the method - if it's there you should use it, if you don't think the default allows "calls to it to be clear" - then it shouldn't really have a default value.

Fair enough.  I'll change it

soco/discovery.py
lawrenceakka: Sorry, I wasn't clear. I don't particularly object to the loop (though I don't really think that repeating a single line three times is bad). I was wondering why it was necessary to "send one message to each interface in turn".

soco/discovery.py
robwebset: Oops, missed that one!

soco/discovery.py
lawrenceakka: > If they went out to different networks then you could get different speakers on each network, so you should wait for all the replies.

In what circumstances? Do you mean if the user had multiple (but independent) sonos systems? We have never catered for this, and it has never been an issue. And if there are independent systems, how do we know which one we are looking for anyway?

soco/discovery.py
robwebset: No major reason (other than removing copy/pasted code) - but looping each interface, sending each in turn before repeating would generally be seen as better as it gives the last one in the list its first message a little earlier, before repeating the send to the first one. Also if there was a problem with a single interface that it "blocks for a second because of some error", then you wouldn't block for a long time before you send to the next interface.

soco/discovery.py
robwebset: I think that's the point - get all the systems and then prompt the user for which they want to use (which is what the Kodi addon does).  Otherwise where is the control for your "Preferred System".  I understand it may not be the way things work at the moment - but I see no need not to code for this sort of thing now.

lawrenceakka:
> Yes, I notice you rejected every single one of the comments!

Nothing personal. And I always agree with the good points :smile: 

> 
> When I did my original searching - it was one of the things suggested to support all platforms, doing the call to a web page seems to set the socket up when all other things have failed - I know it takes longer, this is why I suggested doing it as a last resort - which is how it was implemented in the pull I submitted.
> 
> I see no harm in this, as at that point no speakers have been found - so we should try everything.
> 

If we want to do everything we can (and I'm not sure that we should), the only sound cross-platform solution is to use `netifaces`. But as you have pointed out elsewhere, that introduces a binary dependancy, which neither of us is keen on.  I am afraid I remain unpersuaded (for the moment at least) about adding in the call to sonos.com. I just don't see why finding out which interface has a route to sonos.com will help with finding out which interface has a route to the local sonos network, especially in circumstances where the two other methods have failed. And it adds delay to a process which is supposed to be as quick as possible/messes up the timeout code.

It is always open to the user of the soco library to check for a null return from `discover()` and apply whatever additional solutions are appropriate for that particular platform to find out the relevant interface.

soco/discovery.py
KennethNielsen: I tend to agree with @lawrenceakka on this one. I can see the point of trying to cat Sonos devices on a different network device from the default, but to try and catch Sonos devices on two different networks seems like an very unlikely use case and I would surely like to hear that use case before catering to it.

KennethNielsen:
Hallo @lawrenceakka and @robwebset.

The work on this seems to have stalled a bit and there unquestionably is a need for a more aggressive discovery function.

Summarizing the discussion, there is line 112 which @lawrenceakka agreed to change, a few minor style discussions and then the 2 remaining core issues:

 1. Catering for independent Sonos systems on different networks
 2. Using a call out to Sonos.com as a last resort.

Disregarding for a moment our opinions of these, they are both in essence "additions" to the core enhancements that this PR currently contains. So might I suggest, to get at least the improvements that we agree on out into the hands of the users, that we merge this in the current form and then start up separate discussions over issues 1 and 2 if the need arises or you feel that it needs more attention?

lawrenceakka:
> Hallo @lawrenceakka and @robwebset.
> 
> The work on this seems to have stalled a bit and there unquestionably is a need for a more aggressive discovery function.
> 
> Summarizing the discussion, there is line 112 which @lawrenceakka agreed to change, a few minor style discussions and then the 2 remaining core issues:
> 
> Catering for independent Sonos systems on different networks
> Using a call out to Sonos.com as a last resort.
> Disregarding for a moment our opinions of these, they are both in essence "additions" to the core enhancements that this PR currently contains. So might I suggest, to get at least the improvements that we agree on out into the hands of the users, that we merge this in the current form and then start up separate discussions over issues 1 and 2 if the need arises or you feel that it needs more attention?

Hi @KennethNielsen 

Sorry I've been out of action for a while. I'll look at this in the next couple of days, retest, and commit something. I'll also try to catch up on a few of the other issues.

Lawrence

Issue #138, opened 2014-05-21
Executing commands on a slave zone


DPH:
Many commands only apply to a master (coordinator) zone and you would not expect a user to execute them on a slave. However people do! 
Below is a list of commands tested against a slave zone, and at the bottom one against a zone that has volume set to fixed (typically a ZP90 driving an external amp). It shows the command and the behaviour / error that occurs.

Maybe this behaviour is what we should expect, and users should trap it?
Or should we do nothing and just return (as play and stop do)?
Or should we raise a trappable exception with a clear message that it should not be executed on a slave?
Or......

Welcome thoughts
Cheers David

```
1) commands executed on a slave:

    play               does nothing - ok?
    stop               does nothing - ok?

    play_uri           un_joins zone to be a master and plays -ok?
    play_from_queue    un_joins zone to be a master and plays -ok?
    switch_to_line_in  un_joins zone to be a master and sets to line in -ok?

    pause              soco.exceptions.SoCoUPnPException: UPnP Error 701 received: Transition not available from 192.168.1.69
    seek               soco.exceptions.SoCoUPnPException: UPnP Error 701 received: Transition not available from 192.168.1.69
    next               soco.exceptions.SoCoUPnPException: UPnP Error 701 received: Transition not available from 192.168.1.69
    previous           soco.exceptions.SoCoUPnPException: UPnP Error 701 received: Transition not available from 192.168.1.69
    partymode          soco.exceptions.SoCoUPnPException: UPnP Error 501 received: Action Failed from 192.168.1.72
    add_to_queue       soco.exceptions.SoCoUPnPException: UPnP Error 800 received:  from 192.168.1.69
    remove_from_queue  soco.exceptions.SoCoUPnPException: UPnP Error 800 received:  from 192.168.1.69
    clear_queue        soco.exceptions.SoCoUPnPException: UPnP Error 800 received:  from 192.168.1.69
    
2) command executed on zones with fixed volume set: 
    volume             soco.exceptions.SoCoUPnPException: UPnP Error 501 received: Action Failed from 192.168.1.71
```

KennethNielsen:
please check out #296 :)

KennethNielsen:
This should be fixed with #296 as found the latest release 0.11. Please re-open if that is not the case.

AdrienPensart:
This simply does not work : 

```
#!/usr/bin/env python
from soco import SoCo

if __name__ == '__main__':
    sonos = SoCo('10.65.0.167')                        
[...]

PR #296, opened 2015-03-05
Only on master

KennethNielsen:
This is my attempt to solve the problem of calling master-only methods on a slave. It fixes this along the lines of option 2 as discussed here https://github.com/SoCo/SoCo/issues/138#issuecomment-44221244 by @lawrenceakka (Raises SoCoSlaveException when calling a master command on a slave). It is implemented as a decorator should be used to decorate all the master only methods in SoCo (of course it might it might turn out to be better to reverse the logic if there are many more master-only than master-and-slave method).

Right now it is only implemented on the play and pause method, because I wanted to get some input about whether you like this approach or not, before I start testing all the methods in SoCo for whether they can be called on a slave.

Please also have a look at the test modifications. This is my first run-in with mock (I spent about twice the time on the test modifications as I did on the implementation), so I'm very curious if there is a better way to adapt the tests for this change.

EDIT: I improved the test modifications, so that they actually are still unit tests, so it should be better now, but still could use a review.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -17,6 +17,7 @@ from .services import RenderingControl, AVTransport, ZoneGroupTopology
 from .services import AlarmClock
 from .groups import ZoneGroup
 from .exceptions import SoCoUPnPException, SoCoSlaveException
+    SoCoSlaveException
 from .data_structures import DidlPlaylistContainer,\
     SearchResult, Queue, DidlObject, DidlMusicAlbum,\
     from_didl_string, to_didl_string, DidlResource
@@ -89,8 +90,8 @@ def only_on_master(function):
     def inner_function(self, *args, **kwargs):
         """Master checking inner function"""
         if not self.is_coordinator:
-            message = 'The method "{}" can only be called on the '\
-              'coordinator in a group'.format(function.__name__)
+            message = 'The method or property "{0}" can only be called/used '\
+                'on the coordinator in a group'.format(function.__name__)
             raise SoCoSlaveException(message)
         return function(self, *args, **kwargs)
     return inner_function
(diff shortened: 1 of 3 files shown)


KennethNielsen:
I improved the test changes, so that they actually are still unit tests.

I'd like some feedback on this, if anyone has a little time (I know it can be tight).

stefankoegl:
I like the approach! From my perspective we can continue like this.

KennethNielsen:
Ok, I have now applied the decorator to all the methods that I think it should be used on. Please test it and let me know what you think.

A few specifics of the application of the decorator below:

For cross_fade the getter works and returns the value for the slave, but the setter fails. Since it would be required to apply the decorator to the setter I decided to also apply it to the getter to not break the symmetry.

play_uri will add the uri to the slave and then break the group and ultimately fail because we cannot call play, therefore mark the entire method as only_on_master

stop, stops playback on the slave, but without breaking the group and the master still plays?!?, also mark that as only_on_master

I think the above cases of somewhat inconsistent or unexpected behavior seems to me to support our idea of trying to catch this incorrect use before it is sent to the player.

When methods calls other methods for the actual interaction, I have chosen to mark both (e.g. add_uri_to_queue and add_to_queue) because the implementation is hidden from normal use and the error message will then mention another method, which will be confusing (and if the implementation is changed later, the information from my testing will not be lost).

add_item_to_queueable_item seems to have been forgotten in the data structures restructuring, at least it tries to form metadata in the old way. Since it is broken I cannot test it. I will open a separate issue for that.

stefankoegl:
>I think the above cases of somewhat inconsistent or unexpected behavior seems to me
>to support our idea of trying to catch this incorrect use before it is sent to the player.

Agreed

Good catch regarding ``add_item_to_queueable_item`` (#308)!

+1 to the whole PR -- nice work :)

KennethNielsen:
I seem to have messed up the rebase, please hold

stefankoegl:
Seems to be fixed now, right?

KennethNielsen:
Yes. I'll merge.

PR #302, opened 2015-04-26
Timeout request

xxdede:
Added a timeout option to ‘requests’ while getting speaker info.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1018,6 +1018,9 @@ class SoCo(_SocoSingletonBase):
         timeout -- How long to wait for the server to send
                    data before giving up, as a float, or a
                    (`connect timeout, read timeout`_) tuple.
+                   If not specified a default (3, 5) timeout
+                   will be used, to disable it at all set
+                   explicitly to None.
 
         Returns:
         Information about the Sonos speaker, such as the UID, MAC Address, and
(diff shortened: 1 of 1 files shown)


stefankoegl:
Looks good! The comment describing the parameter should however describe how to disable the timeout. I assume passing ``None`` would work, but this should be stated explicitly.

xxdede:
You're right.

I used the timeout parameter of http://docs.python-requests.org/en/latest/api/#main-interface and it's optional so I think `None` should work.

Probably we can simply remove the default value and let you choose if you want to set a timeout

KennethNielsen:
After quite a bit of searching through requests source code I was able to confirm that it does indeed default to None. @xxdede if you include this information in the docstring like @stefankoegl requested then we should be able to merge and get your change into the next release.

xxdede:
Added doc as suggested!
=)

stefankoegl:
One more thing: ``flake8`` (which is part of the test suite) fails with the following error:

````
soco/core.py:1013:57: E231 missing whitespace after ','
````

When you add the space the build should succeed, and (from my point of view) your change should be ready to be merged.

KennethNielsen:
Merged. Thank you @xxdede for your contribution.

Can you add a comment to the release notes issue #332 that describes your addition?

PR #305, opened 2015-05-29
Added ability to add metadata to items being added to the queue

Utsav2:
Please let me know if there's anything I should change.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1568,31 +1568,29 @@ class SoCo(_SocoSingletonBase):
         return response, metadata
 
     @only_on_master
-    def add_uri_to_queue(self, uri, meta=''):
+    def add_uri_to_queue(self, uri, meta=None):
         """Adds the URI to the queue
 
         :param uri: The URI to be added to the queue
         :type uri: str
         :param meta: The track metadata to show in the player, DIDL format.
-        :type uri: str
+        :type meta: str
         """
         # FIXME: The res.protocol_info should probably represent the mime type
         # etc of the uri. But this seems OK.
         res = [DidlResource(uri=uri, protocol_info="x-rincon-playlist:*:*:*")]
-        if meta == '':
+        if meta is None:
             item = DidlObject(resources=res,
                               title='', parent_id='', item_id='')
         else:
             didl_objects = from_didl_string(meta)
             if didl_objects is None or len(didl_objects) == 0:
-                _LOG.debug('Failed to construct metadata object'
-                           'from string')
-                return False
-            if len(didl_objects) > 0:
-                if len(didl_objects) > 1:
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Hi @Utsav2 

This looks really interesting. Thanks a lot for bringing it here. I don't think we ever did manage to figure out how to add metadata to generic uri's when we add them to the queue.

A few comments. SoCo would like to be a "highlevel" API, so hopefully the user shouldn't need to know anything about the underlying technologies. Therefore, I think it would be better if it was also possible to use this, without constructing the metadata by hand. So if you have generic metadata template that works for any uri, then I would suggest adding an extra optional argument for the title and if this is given use that to construct the metadata inside the method.

On the more technical side:

* It does not look as if the default empty string for the metadata is actually used as an empty string. Therefore I think it is cleaner to make the default argument None and check for that. I know it is done in this way other places in the library, but no reason to continue ;)

There are few more technical details I'll add as inline comments.

But once again, this looks really great, can't wait to get it in.

soco/core.py
```diff
@@ -1501,16 +1501,32 @@ def _music_lib_search(self, search, start, max_items):
...
-        item = DidlObject(resources=res, title='', parent_id='', item_id='')
+        if meta == '':
+            item = DidlObject(resources=res,
+                              title='', parent_id='', item_id='')
+        else:
+            didl_objects = from_didl_string(meta)
+            if didl_objects is None or len(didl_objects) == 0:
+                _LOG.debug('Failed to construct metadata object'
+                           'from string')
+                return False
```
KennethNielsen: We are trying to get away from using the return value to indicate success. I would suggest to raise a ValueError here instead.

soco/core.py
```diff
@@ -1501,16 +1501,32 @@ def _music_lib_search(self, search, start, max_items):
...
+            if len(didl_objects) > 0:
```
KennethNielsen: You already checked in 1520 that the length is not 0, so this seems redundant.

Utsav2:
Rebased it to 0.11 and made the recommended changes, please let me know if I should change anything else.

KennethNielsen:
Hallo @Utsav2.

Sorry about taking this long to get back to you on this. In order to use this ability, it is still required to create the metadata string by hand before passing it to the method, which is sort of a low lever approach to this. As I mentioned in my original review comment, I think it would be better if the metadata passed could be either in the for of a metadata Didl string or e.g. a dict with the attributes to be formatted into such a string.

Let me know if you have any questions.

\Kenneth

ghcs27:
I'm closing this PR since it is very old and I think we don't need it anymore. Please reopen if you feel otherwise.

PR #307, opened 2015-05-31
Added play_local_files example

KennethNielsen:
One of the more frequently asked questions is how to play local files. While, as far as I know, it is not possible in the sense that we have found no interface to stream data to the units, we can cheat and just setup a temporary http server with python.

This example shows how, and plays a random file from the current folder on a named Sonos unit.

### DIFF
--- a/examples/play_local_files/play_local_files.py
+++ b/examples/play_local_files/play_local_files.py
@@ -1,0 +1,106 @@
+"""A play local music files example
+
+To use the script:
+
+ * Make sure soco is installed
+ * Drop this script into a folder that, besides python files, contains
+nothing but music files
+ * Adjust the settings on the first three lines of the main function
+ * Run the script
+
+"""
+
+
+from __future__ import print_function, unicode_literals
+
+import os
+import time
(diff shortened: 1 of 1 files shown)


DPH:
Great idea for playing local files - well done. 

To date I have used the network path to a file, which works if you know the network path and the path is available for Sonos to read in terms of permissions. Then I can use play_uri with a file name:
for example: 'x-file-cifs://MyNAS/sounds/Tune.mp3'. 

Your approach is more generic
Cheers David

stefankoegl:
Great idea :)

KennethNielsen:
@DPH Yes. I also think it would be a good idea to the question "How do I play local files" to tell people that they should setup a network share and refer them to a few good resources, bacause that solution is much more better in terms of reliability (e.g. the proposed example requires that you keep the python script running to serve the files).

The idea with this, was just to have a entry point with a lower barrier for entrance.

KennethNielsen:
I think since this is just an example and there are a few positive evaluations it should be safe to merge. I will go ahead.

PR #309, opened 2015-06-02
Test clean up

KennethNielsen:
This is a simple pull request that removes some old un-used stuff.

First and foremost it removes all the tests for the old data structures. It seems pretty clear that the new data structures are much better and they already have full test coverage, so there really is no reason to keep them around, and I was getting annoyed with all those XFAIL in the test output.

The second removal is the old hand written test execution script that I wrote before learning about pytest and likes. This has not been used in ages and should be thrown out.

Let me know what you think.
Regards Kenneth


stefankoegl:
+1

KennethNielsen:
I'll assume that a single +1 in 3 days is enough for this simple cleanup and merge. If anyone disagrees just shout out.

Issue #308, opened 2015-06-02
add_item_to_sonos_playlist fails after data structure refactoring


KennethNielsen:
Test code:
```python
import soco
zone = soco.discover().pop()
album = zone.get_albums()[0]
sonos_playlist = zone.get_sonos_playlists()[0]
zone.add_item_to_sonos_playlise(album, sonos_playlist)
```
fails with a traceback like:

```
Traceback (most recent call last):
  File "../../test.py", line 30, in <module>
    zone.add_item_to_sonos_playlist(album, pl)
  File "/home/kenneth/venv/soco/SoCo/soco/core.py", line 1741, in add_item_to_sonos_playlist
    raise AttributeError(message)
AttributeError: queueable_item has no attribute didl_metadata
```

I'm fairly certain the only problem is that we need to replicate the behavior from other places for how to form the metadata. This should block 0.11 since it is a regression.

PR #310, opened 2015-06-05
Iss308 fix add item to sonos playlist after data structure refactoring

KennethNielsen:
This is the PR that should fix issue #308. I merely change the implementation of the method to reflect the new data structures by essentially copying the implementation of add_to_queue.

This PR also removes a do-nothing-line from add_to_queue that was left after the fix in #306. At first I didn't understand why it even worked. In #306 I made sure that to_didle_string always returns a unicode object, the way its docstring suggests it should. But then I actually thought that the service call always wanted utf-8, because that is what is needed for the XML, so I was very surprised that the encoded version of the metadata was not assigned to anything and therefore not used. Then I checked up on and as it turns out the service calls does expect unicode (the way we originally decided, decode early, unicode inside, encode late), so that was the reason that it worked even with this line doing nothing. In any case, it is now removed.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1516,7 +1516,6 @@ class SoCo(_SocoSingletonBase):
     def add_to_queue(self, queueable_item):
         """ Adds a queueable item to the queue """
         metadata = to_didl_string(queueable_item)
-        metadata.encode('utf-8')
         response = self.avTransport.AddURIToQueue([
             ('InstanceID', 0),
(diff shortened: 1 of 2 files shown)


KennethNielsen:
Hmm. You'd think I would have learned to remember to lint and test before submitting by now. Oh well, now it is ready.

stefankoegl:
+1, can go in anytime!

KennethNielsen:
Awesome, I'll merge.

PR #328, opened 2015-08-13
Radio redux

KennethNielsen:
I have had this stuff laying around for a while. It is the last part of finally using data structures everywhere. The last part that was missing was the radio shows and favorites.

This PR adds a generic `play_stream` method, which plays a playable DidlObject as a stream i.e. in Sonos single play mode which doesn't affect the queue.

Besides this I have also changed the `get_favorite_radio_shows` and `get_favorite_radio_stations` methods to return DidlObjects (and cleaned them up a bit). This means that you can now do:

``` python
>>> import soco
>>> zone = soco.discover().pop()
>>> 
>>> stations = zone.get_favorite_radio_stations()
>>> print(stations)
SearchResult(items=[<DidlAudioBroadcast 'b'DR P3 93.9 (Euro Hits)'' at 0x7f199661cd68>, <DidlAudioBroadcast 'b'Radio ?stsj?lland 104.7 (World Music)'' at 0x7f199661cb00>], search_type='browse_favorite_radio_stations')
>>> 
>>> station = stations[0]
>>> zone.play_stream(station)
```

I seems to work good, `but I need help testing the favorite_radio_shows part`, as I have been unable to find a show I could mark as a favorite. I need someone that has a radio show favorite to either test that it works getting the favorite shows and that play_stream can play them or simple send me the name of the radio station and show, then I can test it myself.

Let me know what you think

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -487,7 +487,30 @@ class SoCo(_SocoSingletonBase):
             return self.play()
         return False
 
-    @only_on_master
+    def play_stream(self, stream_object, start=True):
+        """Play a DIDL object as a stream
+
+        To play as a stream means that it will this single item without
+        affecting the queue
+
+        Args:
+            stream_object (DidlObject): A Didl Object that represents a stream
+            start (bool): Whether to start play back after adding the stream
+        """
+        # E.g. radio favorites are returned without an desc element. This hack
+        # may turn out to be insufficient if we want to try and play radio
+        # stations from music services in the future
+        if stream_object.desc is None:
+            stream_object.desc = 'SA_RINCON65031_'
+        meta = to_didl_string(stream_object).encode('utf-8')
+        self.avTransport.SetAVTransportURI([
+            ('InstanceID', 0),
+            ('CurrentURI', stream_object.resources[0].uri),
+            ('CurrentURIMetaData', meta)
+        ])
+        if start:
+            self.play()
+
     def pause(self):
         """Pause the currently playing track.
 
--- a/soco/core.py
+++ b/soco/core.py
@@ -492,22 +492,24 @@ class SoCo(_SocoSingletonBase):
             return self.play()
         return False
 
+    @only_on_master
     def play_stream(self, stream_object, start=True):
         """Play a DIDL object as a stream
 
-        To play as a stream means that it will this single item without
+        To play as a stream means that this single will be played without
         affecting the queue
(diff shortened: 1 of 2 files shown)


lawrenceakka:
Works well for me (using BBC Radio 3 as a favorite)

KennethNielsen:
@lawrenceakka but is BBC Radio 3 not a station? What I haven't been able to test yet is the favorite *show*.

lawrenceakka:
Ah. Sorry!

I have found a show (in the desktop client) by searching for "bbc world commute" in the "podcasts and shows tab". Click on "Recent Episodes" and select "Info and Options". You should then be able to add it your favorite shows.

Unfortunately, your code then returns `DIDLMetadataError: Unknown UPnP class: object.container.radioShow` so it looks as though you will need to add that class to data_structures

KennethNielsen:
@lawrenceakka Ah yes, I finally found one and I am now running into the same problem. Will need to look a little more into this because: 1) It is not part of the spec 2) As is, I can't really do anything with the structure (browse nor play). Grumble grumble, I will have to pull out wireshark for this I think. Thanks for helping me test.

KennethNielsen:
Ok. So I have now dug a little more into the situation with the favorite radio shows. It turns out, because radio station support is this mish mash of built in Sonos support and TuneIn, that when you need to browse the radio favorite shows (which is not playable)[1] to get to the actual shows, then you need to use the TuneIn plugin. And when doing that, you probably need to extract the proper TuneIn ID from the URI. So I think a more proper implementation of this has to be considered. Possibly when/if moving radio stuff into its own module, this back and forth between Sonos built in stuff and plugins can be implemented there or maybe the plugin can do everything and we can avoid the shift. For now however, I'm satisfied with the state it is in. I have replaced the return types with data structures, which I will want to do anyway. The returned object for the favorite shows cannot be used for anything except information, but that is (as far as I can tell) not different from the way it was before, so no harm done. So for now, instead of fixing the deficiencies, I have merely documented them in the docstring.

[1] If you try the Sonos app, you will notice that when (in the music tree) you enter the favorite shows. The items listed there are not playable, when browsing those, you get into a single item with "Current shows" and only when browsing that do you get to a list of actual broadcasts, which can be played.

This is ready for review in the current state.

ghcs27:
This is probably redundant now that #478 is merged.

PR #330, opened 2015-08-16
added .to_dict() to DidlResource, wrote tests.  Closes #318

### LINKED ISSUE #318: data_structures.py : DidlResource - add to_dict method
When using get_queue.to_dict, it returns a dict that has a 'resources' element that is a DidlResource object -- which is not serializable in json.  Have a to_dict method on the DidlResource would make life easier, and then get_queue's to_dict could call to_dict on the DidlResource.

I'll submit a patch but I have a question first.  When a DidlResource property is None, should the to_dict method return {'property_foo': None, ...}  or just not include the key in the returned dict?  I would think that the first option is preferable as it requires simpler logic to deal with key=>None than deal with a missing key.    Let me know and I'll start on the patch.   (I'll open a second issue on the get_queue.to_dict calling to_dict on the DidlResource, if/when this patch is accepted, because that might have other ramifications.)

dundeemt:
Created the to_dict() method for DidlResource.  When called just like the .to_dict() of the objects, it will create a python dictionary with all elements.  There is an optional parameter, optimized that can be set to True and the returned dict will have all elements with a value of None stripped from the return.  This is helpful since there are only 2 or 3 elements that have been seen in the wild and when using to_dict() to format a json return object, those unused elements are just unhelpful baggage.  "uri" is the most important of the lot.

I did not make the DidlObject recursively call .to_dict() on resources because that could end up breaking code.
```python
dct = didlobject.to_dict()
dct['resources'] = [resource.to_dict() for resource in dct['resources']]
```
Will get the desired effect, without breaking existing code.


KennethNielsen:
Hallo @dundeemt 

I think this looks really good. A few things crossed my mind while reading though it:

1) About the docstring. We have enabled napoleon for sphinx for our autogenerated API docs, which should pick up the arguments (with type) etc, provided that the doc string is formatted according to the google style doc strings. I can understand that you didn't notice, since we haven't really started the conversion of the docstrings yet, but that will happen. So if you don't mind, it would be nice if you could format new docstrings in this way. It should be something along the lines of:

```python
def to_dict(self, optimized=False):
    """Return a dictionary representation of the DidlResource

    Args:
        optimized (bool): Optionally filter dictionary to only with a value other than None.
    """
```

2) I'm unsure if optimize is the best (most descriptive name), maybe ``filter_out_nones`` or ``reduced``, I'll leave that entirely up to you
3) I completely agree with your evaluation about it being safe to filter out the Nones. For my part, the important part is that the dict can be used to re-instantiate the object by ``DidlResource(**resource_dict)``. In that respect, it should always be safe to filter out the default value, so I think that is a good idea.
4) Speaking of re-instantiating, would you mind also writing the symmetrical classmethod ``from_dict`` (just as the ``DidlObject`` has), as we usually prefer to have both there, so the user is not required to have any inside knowledge to be able to use these methods?
5) With regards to ``DidlObject.to_dict`` calling ``to_dict`` on ``DidlResource``. I would actually prefer that it did it and I think it is ok to break interface for. My reasoning for this is first and foremost that my main expectation of these methods should be that they are symmetrical, so provided that you modify ``from_dict`` on ``DidlObject`` too, that would still be the case. Second, I would also usually expect of a ``to_dict`` method, that it contains only simple types, so I would actually consider it more of a bug in the first place that a break of API. @lawrenceakka @stefankoegl do you guys have an opinion on the matter? If we do decide to change it, the interface change should be documented in the doc string.

I guess you are familiar with the github workflow, so if you do decide to make any change, you can just commit them to you the branch that you made the PR from (looks like your master branch) and they will show up here.

I apologize in advance for the verbose feedback, I don't really do short ;)

Thanks for you work on this.

dundeemt:
Agreed.  I'll start work on it this week.
I appreciate your feedback.  I'll keep an eye out for any feedback from the
people you mentioned.

Best,
Jeff

On Mon, Aug 17, 2015 at 1:37 PM, Kenneth Nielsen <notifications@github.com>
wrote:

> Hallo @dundeemt <https://github.com/dundeemt>
>
> I think this looks really good. A few things crossed my mind while reading
> though it:
>
> 1) About the docstring. We have enabled napoleon for sphinx for our
> autogenerated API docs, which should pick up the arguments (with type) etc,
> provided that the doc string is formatted according to the google style doc
> strings. I can understand that you didn't notice, since we haven't really
> started the conversion of the docstrings yet, but that will happen. So if
> you don't mind, it would be nice if you could format new docstrings in this
> way. It should be something along the lines of:
>
> def to_dict(self, optimized=False):
>     """Return a dictionary representation of the DidlResource    Args:        optimized (bool): Optionally filter dictionary to only with a value other than None.    """
>
> 2) I'm unsure if optimize is the best (most descriptive name), maybe
> filter_out_nones or reduced, I'll leave that entirely up to you
> 3) I completely agree with your evaluation about it being safe to filter
> out the Nones. For my part, the important part is that the dict can be used
> to re-instantiate the object by DidlResource(**resource_dict). In that
> respect, it should always be safe to filter out the default value, so I
> think that is a good idea.
> 4) Speaking of re-instantiating, would you mind also writing the
> symmetrical classmethod from_dict (just as the DidlObject has), as we
> usually prefer to have both there, so the user is not required to have any
> inside knowledge to be able to use these methods?
> 5) With regards to DidlObject.to_dict calling to_dict on DidlResource. I
> would actually prefer that it did it and I think it is ok to break
> interface for. My reasoning for this is first and foremost that my main
> expectation of these methods should be that they are symmetrical, so
> provided that you modify from_dict on DidlObject too, that would still be
> the case. Second, I would also usually expect of a to_dict method, that
> it contains only simple types, so I would actually consider it more of a
> bug in the first place that a break of API. @lawrenceakka
> <https://github.com/lawrenceakka> @stefankoegl
> <https://github.com/stefankoegl> do you guys have an opinion on the
> matter? If we do decide to change it, the interface change should be
> documented in the doc string.
>
> I guess you are familiar with the github workflow, so if you do decide to
> make any change, you can just commit them to you the branch that you made
> the PR from (looks like your master branch) and they will show up here.
>
> I apologize in advance for the verbose feedback, I don't really do short ;)
>
> Thanks for you work on this.
>
> —
> <https://github.com/SoCo/SoCo/pull/330#issuecomment-131920176>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

dundeemt:
PR includes all  except item #4.  I wanted to wait for any input without holding up this issue/pr.  There are some questions I have about #4 and since it will break things and require updated documentation, it might be better to make an issue and pr for just that item.

soco/data_structures.py
```diff
@@ -231,6 +231,58 @@ def to_element(self):
...
+        return content
+
+    @classmethod
+    def from_dict(cls, content):
+        """Create an instance from a dict.
+
+        An alternative constructor. Equivalent to DidlResource(**content).
+
+        Arg:
+            content (dict): Dict containing metadata information.Required and
```
KennethNielsen: Just missing a space after the '.' there

soco/data_structures.py
```diff
@@ -598,7 +650,7 @@ class DidlAudioItem(DidlItem):
 
     """An audio item."""
 
-    item_class = 'object.item.audioItem'
+    item_class = 'object.item.audioitem'
```
KennethNielsen: I think something went wrong with some rebasing as this reverts you misspelling bug fix

KennethNielsen:
I had just a few final inline comments. Other than those I am very happy :D

dundeemt:
I repull from upstream, to take care of the item_class and fix the
spacing.  Will be later today.
Best.

On Fri, Aug 21, 2015 at 10:53 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> I had just a few final inline comments. Other than those I am very happy :D
>
> —
> <https://github.com/SoCo/SoCo/pull/330#issuecomment-133470442>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

dundeemt:
@KennethNielsen if you have any other concerns, comments with this PR please let me know.

lawrenceakka:
+1 from me.  Happy to commit tomorrow if no other comments.

dundeemt:
That would be outstanding.  I ignorantly have the PR going against my main
branch, so I don't believe I can start another PR until after that is
committed.  (I'll make sure and work in a branch other than main. ;)

On Thu, Aug 27, 2015 at 11:46 AM, Lawrence Akka <notifications@github.com>
wrote:

> +1 from me. Happy to commit tomorrow if no other comments.
>
> —
> <https://github.com/SoCo/SoCo/pull/330#issuecomment-135489537>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

lawrenceakka:
Thanks very much for this. Please make sure to add a comment on #332 which we will use to create release notes

KennethNielsen:
@dundeemt Sorry, was away for a few days. No, no further comments and nice work.

Thanks for the contribution. I'll look forward to the next one.

Regards Kenneth

PR #331, opened 2015-08-18
Fix capitalization of object.item.audioItem

mikix:
I was seeing exceptions like "soco.exceptions.DIDLMetadataError: Unknown UPnP class: object.item.audioItem" when subscribing to events.

Fixing the capitalization of DidlAudioItem's item_class member fixed it for me.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -598,7 +598,7 @@ class DidlAudioItem(DidlItem):
 
     """An audio item."""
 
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@mikix I just checked with the spec and this does indeed seem to be a bug. Thanks for fix. I'll pull it right away. Can you add a short comment here to the release notes about the fix #332 ?

Issue #320, opened 2015-07-12
How to Determine Sonos Model?


dundeemt:
Looking to create a method to return the model type (PLAY:1, PLAY:3, PLAY:5, PLAYBAR, SUB, CONNECT, CONNECT:AMP, ??)

I have 3 PLAY:1 that all show a Hardware Version of 1.8.3.7-1  and a PLAYBAR with a hardware version of 1.9.1.10-1 -- if others could pull up their info and list as a comment:
  http://zone_ip:1400/support/review  and click on "Zone Player Info" and report back their Hardware version info along with the model, I believe we can create a method to return the model name based on it.  i.e.
  PLAYBAR - 1.19.10-1
  PLAY:1  - 1.8.3.7-1

I've google'd and couldn't find a resource but if I am missing something, please let me know (resource already available, etc)

Current Feedback as of 2015-07-24, 15:46 CST

|Model      |Hardware Versions
|---------     |-----------------------------
|PLAY:1    | **1.8.3.7-1**, 1.8.3.7-2
|PLAY:3    | 1.8.1.2-2
|PLAY:5    | **1.16.4.1-2**
|PLAYBAR| 1.9.1.10-1
|SUB          |
|CONNECT| **1.16.5.5-2**
|CONNECT:AMP|

If a hardware version has more than one person reporting the same version, it is in bold.  (One person reporting multiple matching versions is currently being counted as 1.  subject to feedback.)

poirier:
Hardware version 1.8.3.7-1 - Play 1
1.16.5.5-1 - Connect

DPH:
1.16.4.1-2 Play 5
1.16.5.5-2 Connect

Cheers David

KennethNielsen:
```python
>>> for zone in soco.discover():
...     print(zone.player_name, zone.get_speaker_info()['hardware_version'])
... 
('Stue', '1.16.5.5-2')  # Connect
(u'K\xf8kken', '1.8.1.2-2')  # Play 3
(u'Sovev\xe6relse', '1.8.3.7-2')  # Play 1
(u'Badev\xe6relse', '1.8.3.7-2')  # Play 1
('Kontor', '1.16.4.1-2')  # Play 5
```

KennethNielsen:
@dundeemt Maybe you cound create a table in the first post, so that we can easier see what we are missing.

dundeemt:
@KennethNielsen an excellent suggestion.  done.

the01:
[...]

Issue #332, opened 2015-08-18
Release notes for 0.12


KennethNielsen:
For every change that goes into the 0.12 release. Please make a small description here, link to the PR (any issues) and mention whether it is a bug fix or not.

@stefankoegl you usually open these, but I thought I would just get it started.

KennethNielsen:
Add information to the docs about how to put SoCo in the Python path, for test execution. PR #327 issue #319.

KennethNielsen:
Clear zone group cache and reparse zone group information after join and unjoin to prevent giving wrong topology information. (BUGFIX). PR #323 issue #321.

mikix:
Fix typo preventing SoCo from parsing the audio metadata object used when a TV is playing. PR #331.

dundeemt:
added to_dict() / from_dict() to DidlResource PR #330  Issue #318

lawrenceakka:
New MusicService class for access to all the music services known to Sonos.  Note that some of this code is still unstable, and in particular the data structures returned by methods such as `get_metadata` may change in future. (#262)

lawrenceakka:
All tests have been moved from the unittests directory to the tests directory ( #336 )

lawrenceakka:
For developers, more make targets, and a better sdist build ( #346 )

lawrenceakka:
Added `discovery.any_soco()`, for when you need a SoCo instance, but you don't care which. This is slightly better than the traditional `device = soco.discover().pop()` since it will return an existing instance if one is available, without sending discovery packets ( #262 )

dundeemt:
Modified `DidlObject.to_dict()` so that any associated `resources` list will be also returned as a list of dictionaries instead of returning a list of DidlResource objects. (#340)

lawrenceakka:
Added a `sonosdump` tool in `dev_tools`, which can print out the various UPnP methods which Sonos uses

lawrenceakka:
Methods relating to the music library (`get_artists`, `get_album_artists`, `get_albums` and others) have been moved to the music_library module. Instead of `device.get_album_artists()`, please now use `device.music_library.get_album_artists()` etc. Old code will continue to work for the moment, but will raise deprecation warnings (#350)

dundeemt:
Added remove_sonos_playlist method.  Closes issue #341 (PR #345)

dundeemt:
[...]

PR #335, opened 2015-08-24
Preliminary speaker info with more information

the01:
Updated the ```get_speaker_info()``` function to parse ```xml/device_description.xml``` as described in #320

- changed zone_icon to player_icon
- added model_number
- added model_name
- added display_version

No unit tests yet

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1095,21 +1095,38 @@ class SoCo(_SocoSingletonBase):
             return self.speaker_info
         else:
             response = requests.get('http://' + self.ip_address +
-                                    ':1400/status/zp')
-            dom = XML.fromstring(response.content)
+                                    ':1400/xml/device_description.xml')
+            dom = XML.fromstring(response.content.encode('UTF-8'))
 
-        if dom.findtext('.//ZoneName') is not None:
-            self.speaker_info['zone_name'] = \
-                dom.findtext('.//ZoneName')
-            self.speaker_info['zone_icon'] = dom.findtext('.//ZoneIcon')
+        device = dom.find('{urn:schemas-upnp-org:device-1-0}device')
+        if device is not None:
+            self.speaker_info['zone_name'] = device.findtext(
+                '{urn:schemas-upnp-org:device-1-0}roomName')
+
+            # no zone icon in device_description.xml -> player icon
+            self.speaker_info['player_icon'] = device.findtext(
+                '{urn:schemas-upnp-org:device-1-0}iconList/'
+                '{urn:schemas-upnp-org:device-1-0}icon/'
+                '{urn:schemas-upnp-org:device-1-0}url'
+            )
+
             self.speaker_info['uid'] = self.uid
-            self.speaker_info['serial_number'] = \
-                dom.findtext('.//SerialNumber')
-            self.speaker_info['software_version'] = \
-                dom.findtext('.//SoftwareVersion')
-            self.speaker_info['hardware_version'] = \
-                dom.findtext('.//HardwareVersion')
-            self.speaker_info['mac_address'] = dom.findtext('.//MACAddress')
+            self.speaker_info['serial_number'] = device.findtext(
+                '{urn:schemas-upnp-org:device-1-0}serialNum')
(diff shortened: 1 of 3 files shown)


the01:
My guess is that ```dom.findtext('.//{urn:schemas-upnp-org:device-1-0}device')```returns ```None``` and that is the reason the test fails. Any ideas why this only happens for python 2.6?

dundeemt:
It would be nice to also have, memory, flash, ampOnTime.  Also an option to get the full url to the icon i.e. full_urls=True.

Failing on 2.6 - soco/core.py line 1099  -- I am doing something similar in my helper library but I encode the response from requests using .encode('UTF-8').  Translating to your code... 
```python
dom = XML.fromstring(response.content.encode('UTF-8'))
```
This was cannibalized from the commented out code @ soco/core.py def uid

my helper lib is https://bitbucket.org/dundeemt/chilledsoco/src/66be490323f863eabd6ac3e4fdbd8419b4a12372/chilled.py?at=default  class DeviceProperties

the01:
The question is whether memory, flash,..etc. is really all that important or if we are starting to pollute the speaker_info-dict. For now you can always build the full url with the ```.ip_address```property.
Sadly your proposed fix did not work. I can't get my python2.6 installation up and running properly, so maybe someone with a working env could investigate?

lawrenceakka:
There is a bug in path support in python 2.6. We had some previous discussion about it here: https://github.com/SoCo/SoCo/pull/154 (You may need to view comments on outdated diffs to see it):

> This seems to be a bug in Elementtree < 1.3 (i.e. the version in Python 2.6). Your fix is too widely targeted though, since it testing for xml.etree.ElementTree' in sys.modules will be true for Python 3.

> You are right however that the .// is not in fact needed at all here, since the item element is always an immediate child of the top level element, in this particular case.

> So in fact, the best solution is for you to remove the xml_find_prefix code, and remove the ".//" in all cases where you have otherwise used it. i.e. in your patch, replace

> for item in metadata.findall(self.xml_find_prefix +
    '{urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/}item'): 
with

>  for item in metadata.findall('{urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/}item'): 


One way to handle it might be to find the `device` element, and then conduct a search of its immediate children (i.e. avoid ".//" as much as possible).  I haven't tested this though.

I don't mind including memory, flash etc (though I can't image why anyone would need them), but I'm not keen on the full_urls idea. I agree with @the01 that this can easily be done by the user by combining with the ip address if necessary.



Lawrence

dundeemt:
full_urls @lawrenceakka @the01 is a common idiom in soco.  The argument against it here is the same that could be made else where browse/ search/ and variants have a parameter to specify the full url is returned.  I bring it up for consistency.

As for, those other things in device_description.xml, I can see the point about not polluting the get_speaker_info return structure, but offering access to all of that information, somewhere would be a good thing.  Perhaps, that c/sh ould be a different feature request and pr.
-Jeff

the01:
Maybe I will have another go at the whole 2.6 tests later, though my only way to test this is to do a commit and let travis to its thing. Which is not really optimal for debuging purposes (and makes for an interesting git history :smile: ). So it would be greatly appreciated if someone with a working 2.6 installation could have a look.
Though is it actually necessary to support a platform (2.6) that hasn't seen an update since, I think, 2013? Are there any usage statistics on pypi on which version a package gets installed for?

Maybe a ```get_speaker_info_all()``` whitch does the actual request and returns a dict with all possible entries and ```get_speaker_info()``` uses the all-method instead of requests/lxml and just filters the dict for the values we want?

the01:
Alrigth, the tests finally surrendered, we are good to go :wink:

lawrenceakka:
+1 from me. Please squash a few of the commits, and I'll commit.

Separate PRs for absolute URIs, and for additional info (which I think should be in the same dict), are welcome

the01:
Down to only a single commit

lawrenceakka:
Thanks very much for this.  Please make sure to add a comment on #332 which we will use to create release notes

PR #336, opened 2015-08-27
Move unittests to tests

lawrenceakka:
Currrently, tests are in a directory called `unittest`. This shadows the builtin module of the same name, and has caused by numerous problems with flake8 and pytest (which, in a virtualenv, can try to import it instead of the builtin).  Please can we move the tests to a `tests` directory instead?

Lawrence


stefankoegl:
+1

KennethNielsen:
+1, ideally, in the long run, I would like to see all the ones that actually are unittests be moved into a subdirectory of tests called unittest (which should be ok from a path import point of view), so as to make room for other kinds of tests which could have their own folder integration, functional, smoke etc. But I think we can worry about that later.

dundeemt:
+1

PR #344, opened 2015-08-30
Adding sonosdump tool

lawrenceakka:
I use this tool to find out what services are available. It is probably sufficiently useful to be included

### DIFF
--- a/dev_tools/sonosdump.py
+++ b/dev_tools/sonosdump.py
@@ -1,0 +1,72 @@
+#! /usr/bin/env python
+# -*- coding: utf-8 -*-
+
+
+""" Dump all known UPnP commands """
+
+from __future__ import unicode_literals, print_function
+import argparse
+import re
+
+import soco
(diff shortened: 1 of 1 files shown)


stefankoegl:
+1

KennethNielsen:
@lawrenceakka Awesome, this will be really useful.

I just tested it from upstream master and it seems there is a small bug. In line 40, you are assuming that any_soco() is located in the SoCo class and not in fact in discovery. This was a change I believe we discussed, but it was never implemented, so the tool fails without a device.

Changing:
```python
        device = soco.SoCo.any_soco()
```
to
```python
        device = soco.discovery.any_soco()
```
fixes it.

lawrenceakka:
Oops.  In fact, any_soco used to be (briefly) in SoCo, but you and I thought it should be in discovery, so I moved it. I forgot to update this tool.  Now fixed (along with a few whitespace issues)

Issue #341, opened 2015-08-29
add: remove_sonos_playlist


dundeemt:
There are currently a couple of ways to create a sonos playlist but no way to delete them.  I am getting ready to do a PR to add this functionality.  I do have a couple questions and figured it best to get them sorted here before I create the PR.

1) parameter name - the underlying contentDirectory.DestroyObject call requires an 'object_id' to be set. I have the function signature looking like this: `remove_sonos_playlist(self, object_id)`  The content of object_id should be the be the item_id (i.e. SQ:10) of the playlist to delete.  object_id and item_id are used in the code and I am not sure if the signature should be object_id or item_id.  I guessed object_id since that is what the underlying call is using.

2) Proper position in the SoCo class doc string for the new function.  Methods and Properties are split out but within those subgroups it is not strictly alpha order.  It appears to be sub-grouped by functionality and then alpha within that.  I have it listed after add_item_to_sonos_playlist in the docstring and within the code.  Ok? OR Should it come after the two create_... methods?

3) Testing - against the mock soco object only or against the mock and an actual device (--ip)? I am unsure if the project is trying to go completely mock or if the mock usage is just to augment testing when you don't have a reachable device.

4) doc strings and docs - I see a variety of doc string/documenting methods in play.  Which is the preferred version?  The method I used for to_dict @KennethNielsen behaved rather badly when my parameter doc string needed to wrap and I had to butcher it to make it fit on one line within the length limit of pyflake/pep8. I am pretty sure it was my formatting or some such, but I was unsure of where to read up on the requirements to make the output work as intended. 

Best,
Jeff

lawrenceakka:
Hi Jeff,

> 1) parameter name - the underlying contentDirectory.DestroyObject call requires an 'object_id' to be set. I have the function signature looking like this: remove_sonos_playlist(self, object_id) The content of object_id should be the be the item_id (i.e. SQ:10) of the playlist to delete. object_id and item_id are used in the code and I am not sure if the signature should be object_id or item_id. I guessed object_id since that is what the underlying call is using.
> 

item_id seems to be what Sonos uses to refer to the id of a metadata or music item (e.g. a track or album). It uses object_id for UPnP calls.  So I guess object_id is correct here.


[...]

PR #345, opened 2015-08-31
Added remove_sonos_playlist function. 

dundeemt:
Tests written both in test_core using moco and in test_integrations against a device. Closes issue #341.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1825,7 +1825,7 @@ class SoCo(_SocoSingletonBase):
         """ Remove a Sonos playlist.
 
         Args:
-            object_id (str): Id of the Sonos playlist to remove. i.e. SQ:10
+            object_id (DidlPlaylistContainer): Sonos playlist to remove.
 
         Returns:
             bool: True if succesful, False otherwise
@@ -1834,6 +1834,7 @@ class SoCo(_SocoSingletonBase):
             SoCoUPnPException: If object_id does not point to a valid object.
 
         """
+        object_id = getattr(object_id, 'item_id', object_id)
         return self.contentDirectory.DestroyObject([('ObjectID', object_id)])
 
     def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
--- a/soco/core.py
+++ b/soco/core.py
@@ -148,6 +148,7 @@ class SoCo(_SocoSingletonBase):
         create_sonos_playlist -- Create a new empty Sonos playlist
         create_sonos_playlist_from_queue -- Create a new Sonos playlist
                                             from the current queue.
+        remove_sonos_playlist -- Remove a Sonos playlist.
         add_item_to_sonos_playlist -- Adds a queueable item to a Sonos'
                                        playlist
         get_item_album_art_uri -- Get an item's Album Art absolute URI.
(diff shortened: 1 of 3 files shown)


lawrenceakka:
Looks good to me.  +1

dundeemt:
I am working on reordering Sonos Playlists for my next PR.  If @KennethNielsen, or other interested parties would like to weigh on this PR so it can be merged too, it would be a great help in writing the tests for reorder_sonos_playlist.

lawrenceakka:
I'll merge this tomorrow if there are no other comments

KennethNielsen:
Hi @dundeemt 

Sorry about the delay. I like that we get some better coverage on Sonos playlists, but I'm unhappy about the input to the function being only the ID. I would prefer it, if most of the time, these internal IDs were invisible to the user. I would therefore suggest that you make the input to the function be either the playlist object (which conatains the ID) or the ID, but with the play list object being the primary input mode.

Another thing that caught my eye is that the underlying call you are using is to ``self.contentDirectory.DestroyObject``, which seems fairly general. Can this be used to delete e.g. tracks from the music collection index? And if so, should we check and deny that or make the method description and name more general?

dundeemt:
Agreed with the object_id vs. Object.   I've written the
reorder_sonos_playlist and friends to accept either a DidlPlaylistContainer,
or any object that has an item_id property (like browse) or the item_id as
a straight string.
I can/will happily update remove_sonos_playlist to have the same
capabilities, but only document / primarily document the object_id
parameter as accepting a DidlPlaylistContainer object.

As far as being more generic, it might well be.  I've been focused on Sonos
Playlists recently, but you are correct in that it "smells" like it might
be able to modify the Queue.  I haven't done much with the queue outside of
use the already existing soco method.  But can/will take a look at it.  Do
we need to make it more discriminating even if it is capable of deleting an
item from the queue? I've not tried to see if it can delete an item from
the library.  (Wouldn't that be transient at best, with the item being
added back during the next library scan?)  If it can, then we can always
refactor and rename it _destroy_object() then make a helper function
remove_sonos_playlist that calls _destroy_object.

Point being, we can refactor later if it is discovered that destroy_object
has more potential than just deleting Sonos Playlists.    Now that I think
about it, you can't delete the queue, it is persistent.

So, I'll make the change to the param to take a DidlPlaylistContainer /
item_id and we can go from there?



I'll update my PR later today

Best,
Jeff

On Wed, Sep 9, 2015 at 2:21 PM, Kenneth Nielsen <notifications@github.com>
wrote:

> Hi @dundeemt <https://github.com/dundeemt>
>
> Sorry about the delay. I like that we get some better coverage on Sonos
> playlists, but I'm unhappy about the input to the function being only the
> ID. I would prefer it, if most of the time, these internal IDs were
> invisible to the user. I would therefore suggest that you make the input to
> the function be either the playlist object (which conatains the ID) or the
> ID, but with the play list object being the primary input mode.
>
> Another thing that caught my eye is that the underlying call you are using
> is to self.contentDirectory.DestroyObject, which seems fairly general.
> Can this be used to delete e.g. tracks from the music collection index? And
> if so, should we check and deny that or make the method description and
> name more general?
>
> —
> <https://github.com/SoCo/SoCo/pull/345#issuecomment-139020205>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

lawrenceakka:
@KennethNielsen Are you happy with this now.  Looks great to me.

L

KennethNielsen:
> I've written the reorder_sonos_playlist and friends to accept either a DidlPlaylistContainer, or any object that has an item_id property (like browse) or the item_id as a straight string.

Great. I like the ```get_attr(playlist, 'item_id', None) or playlist``` I think you did somewhere ;). I will have a look at reorder_sonos_playlist and friends in that PR.

> I can/will happily update remove_sonos_playlist to have the same capabilities, but only document / primarily document the object_id parameter as accepting a DidlPlaylistContainer object.

Yes, please do. I think that would be great. I think however maybe that I expressed myself un-clearly. Besides from making the playlist object the primary input type in the docstring, I think it would be best to also change the name of the input parameter to reflect it. People read documentation differently, some just glance the call spec, so I think it too should reflect the playlist container being the primary input type. So replacing the name of ``object_id`` with the something like ``sonos_playlist``.

Great work with this. With those few minor changes, this is ready to go in.

dundeemt:
PR in line with other PR #352  Have updated doc strings, variable naming and preference for DidlPlaylistContainers.

KennethNielsen:
+1 I'll merge it. Remember to leave a line in the release notes.

iloveagent57:
@dundeemt Thank you for this function!  I had a couple hundred playlists I wanted to delete, and was able to build from master and delete them without issue.

PR #346, opened 2015-08-31
More make targets, and better sdists

lawrenceakka:
Added a few useful Makefile targets, and improved the Manifest.in file to ensure that tests, examples etc are included in the source dist

### DIFF
--- a/makefile
+++ b/makefile
@@ -3,4 +3,20 @@ lint: soco
 	flake8 --ignore=E402 soco
 	pylint soco
 
-.PHONY: lint
+test:
+	py.test
+
+docs:
+	$(MAKE) -C doc html
+	@echo "\033[95m\n\nBuild successful! View the docs at docs/_build/html/index.html.\n\033[0m"
+
+clean:
+	find . -name '*.py[co]' -delete
+
+	find . -name '*~' -delete
+	find . -name '__pycache__' -delete
+	rm -rf soco.egg-info
+	rm -rf dist
+	$(MAKE) -C doc clean
+
+.PHONY: lint docs test clean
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -3,4 +3,20 @@ lint: soco
...
+docs:
+	$(MAKE) -C doc html
+	@echo "\033[95m\n\nBuild successful! View the docs at docs/_build/html/index.html.\n\033[0m"
+
+clean:
+	find . -name '*.pyc' -delete
+	find . -name '*.pyo' -delete
+	find . -name '*~' -delete
+	find . -name '__pycache__' -delete
+	rm -rf soco.egg-info
```
KennethNielsen: Recently I was made aware, that if working with python projects in a virtual environment, one way to get the PYTHONPATH set up (instead of manually modifying the activate script) is to simply install SoCo from source in **editable** mode with pip.
```
# Inside the SoCo dir
pip install -e .
```
This however also leaves an soco.egg-info, which I am guessing* is actually needed for it to work. This would get in the way of that. I don't know if we should do anything about that at all. If something is to be done, there could be two clean targets (I believe they can call each other) or this line could be removed.

* I'm still not sure how this editable mode works and I haven't had time to investigate it yet.

KennethNielsen:
Good idea. A single inline comment.

+1

lawrenceakka:
@KennethNielsen `pip install -e .` is basically the same as `python setup.py develop` (see http://pythonhosted.org//setuptools/setuptools.html#development-mode ). the `egg-info` directory just contains some metadata, which is not needed for it to work - it finds everything in needs in setup.py.

KennethNielsen:
@lawrenceakka cool, thanks for the info

Issue #127, opened 2014-05-10
core.py and services.py have too many lines


stefankoegl:
core.py and services.py have too many lines and should probably be refactored.

ghcs27:
In the core.py, the methods for getting favorites and managing playlists aren't really functionalities of the player, but rather of the whole sonos system.
I would suggest refactoring them to their own module.

PR #350, opened 2015-09-04
Moving music library functions to own module

lawrenceakka:
`core.py` is getting very large indeed (see #127). I find it very hard to navigate, and increasingly confusing. It is unclear for new users that some functions, e.g. `get_artists` work only on the music library, and not on music services such as Spotify. I propose that we move all the music library related features to their own module.  This takes about 500 lines out of core.  

Instead of `device.get_album_artists()`, you will have to use `device.music_library.get_album_artists()` etc.

To me, this makes it much clearer which methods relate to the music library.  This PR achieves this but keeps the old methods in place (marked as deprecated).  I have not updated any docs (yet).

What do people think?

We might also consider moving radio related functions to a radio module, and perhaps queue related functions to a player_queue module (to avoid conflicting with the stdlib queue module).

An alternative would be to have a top level `MusicLibrary` object, e.g. 
```python
>>> from soco import MusicLibrary
>>> library = MusicLibrary()
>>> library.get_album_artists()
SearchResult(items=[<DidlContainer 'b' (Compilations)'' at 0x108ea6160>, <DidlMusicArtist 'b'ABBA'' at 0x108ea61d0>,....
```

This is perhaps more consistent with the new music services code. It would be easy to change the PR to this if preferred.  I haven't decided which I like yet. [Update: We can have both quite easily]

Lawrence

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
-# pylint: disable=C0302, fixme, protected-access
-
-"""The core module contains the SoCo class that implements the main entry to
-the SoCo functionality."""
+# pylint: disable=fixme, protected-access
+""" The core module contains the SoCo class that implements
+the main entry to the SoCo functionality
+"""
 
 from __future__ import unicode_literals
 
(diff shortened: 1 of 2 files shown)


KennethNielsen:
Hi @lawrenceakka 

I think this looks like a good idea. If it was only a matter of the size of core, another (really terrible) idea could be to use a mix-in class, but they are difficult to read and as you say it is also becoming difficult for users to figure out which methods act on what. So I think this approach looks really good.

I was about to comment (before your update) that I think it would be really cool to have both, i.e. have e.g. the music library available as a property of core and as a standalone class, so the way you have implemented looks good to me.

Regarding your comment about the music services, I would actually also in the long run like to see us have an entry point for those in core, which would one again make it consistent with the structure you propose here.

With regards to potentially moving radio and queue, I think the key reason for whether to move or not, should be what users expect and therefore something along the lines of "where does this concept naturally belong". In that respect I agree that probably radio could be moved, but probably that queue shouldn't.

I think since most of this PR is code being moved, there is no reason for en incredibly thorough review, so I'll say +1.

Do you want to move the tests in a seperate PR?

Issue #349, opened 2015-09-04
Code formatting todo


lawrenceakka:
There are a few code formatting issues I would like to correct:
- [x] Docstring formatting to be PEP257 compliant (using [docformatter --no-blank](https://pypi.python.org/pypi/docformatter))
- [ ] Ensure docstring params are in accordance with the Google style guide
- [x] Import reordering to be pep8 compliant, and easier to follow (i.e. in the order stdlib, third party, local) (using [isort](https://pypi.python.org/pypi/docformatter)). I prefer the 'vertical hanging indent' option (option 3 on the homepage.
- [x] Ensuring presence of appropriate `from __future__` imports, and general consistency of file headers
- [x] Consistent use of express relative imports (i.e. `from . import`)

Does anyone have comments? If, not, I will produce a PR

KennethNielsen:
Another issue, that also touches the docstrings, us to napoleonize them. This will however probably be a bigger change, so maybe it is something we should do on a module by module basis.

With regards to docformatter I must admit that I have always had a hard time with that extra blanc line at the end of multiline doc strings, especially because if I remember it correctly, it was basically made a recommendation to make it easier to use auto linebrraking features of cettains editor (or maybe just 1, emacs), which although I use it, I think is a bit silly. In any case, it is in pep257 so ok by me.

lawrenceakka:
@KennethNielsen Docformatter offers:
```
--no-blank            do not add blank line after description
```

I'm happy with that.

KennethNielsen:
That would be cool. BTW sorry about barely legible English in the previous post. That was from my phone.

pwt:
Superseded by other changes.

PR #351, opened 2015-09-06
Tidy up

lawrenceakka:
Most of #349 implemented.  Still to do is the napoleonisation of docstrings.  Everyone happy to commit this at the moment?

Lawrence

### DIFF
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -5,9 +5,10 @@
 from __future__ import unicode_literals
 
 import logging
-from datetime import datetime
 import re
 import weakref
+from datetime import datetime
+
 from .core import PLAY_MODES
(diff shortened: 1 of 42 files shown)


soco/alarms.py
```diff
@@ -16,7 +17,8 @@
 
 
 def is_valid_recurrence(text):
-    """Check that text is a valid recurrence string.
+    """
```
KennethNielsen: Ah, I had not seen from the proposal that this was part of it. Skipping the first line, don't think it is pretty, but not a complaint worth discussing further. I'm just happy to see the doc strings cleaned up.

KennethNielsen:
I just looked this over at a glance. Looks great +1

Issue #348, opened 2015-09-03
Preferred return structure


dundeemt:
working on reorder_sonos_playlist PR, I have 3 values to return from the call. I've been looking at existing code trying to get a feel on whether to return a tuple of values or a dict.  I've found two examples. One returns a tuple of two values (_music_lib_search) tuple(dict, int) and the other returns a dictionary(__get_radio_favorites).

Given that I am new to the project, I am unsure what the current project preference is in this situation.  I am leaning towards returning a dict. It looks cleaner for the caller. The values have assigned names (keys).  Looking for guidance.

On a different topic: update_id.  Most methods just hard code a 0, while the add_item_to_sonos_playlist fetches the current update_id and use that.   How sensitive is update_id?  I am currently mimicing add_item if update_id is set to 0, otherwise I use what is given and then returning the NewUpdateID from reorder_sonos_playlist so if multiple reorders are called in a row, we can save the extra call to fetch the current.  Good / Bad / meaningless?

KennethNielsen:
Hallo @dundeemt 

The radio stuff is a matter of legacy. I think for all cases where (some of) the returned information fits into on of the data structures for a media, then that is the preferred form. I have a PR (#328) to finally make this happen for the radio stuff.

I don't think we have every really decided on a general practice in the project, so this is just my personal view. In general I think it depends on the nature if the returned information:

 * I think, that if the information fits **naturally** in a data structure, then use or make one. See e.g. all the search and browse methods. If not then:
 * If these are all "values" for a single entity of something, then I think a dict would be a good choice.
 * The there are only a few (2-3) and 1 or 2 or these are more support information then actual information (say the number of results, success status etc.) then I would use multiple returns i.e. tuple. As a side note, I really fond of named tuples for these cases (well after I learned about them anyway).

If you can give information about what it is you are returning, then I can give you my personal view on this specific case.

dundeemt:
The method's return signature:
            (int): Change in size between original playlist and the
resulting playlist.
            (int): Length of resulting playlist.
            (int): update_id for next operation.
The  reorder call can not only move tracks, but can be used to delete
tracks from a Sonos playlist. All of the values are status values, except
for update_id. Out of the 3 return results, the softest is the first,
because it can be calculated, in the calling code, by tracking the length
of the original playlist. I included it because it is part of the
underlying call's output.  If am -0 on keeping it but wanted group input
before I just dropped it.


[...]

PR #352, opened 2015-09-07
Major PR for Sonos Playlist Management.

dundeemt:
Add base method reorder_sonos_playlist and a number of helpers, documentation and tests.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -11,9 +11,11 @@ import re
 import socket
 from functools import wraps
 
+
 import requests
 
 from . import config
+from .compat import UnicodeType
 from .data_structures import (
     DidlObject, DidlPlaylistContainer, DidlResource,
     Queue, from_didl_string, to_didl_string
@@ -1240,7 +1242,8 @@ class SoCo(_SocoSingletonBase):
 
         """
         args = tuple(['sonos_playlists'] + list(args))
-        return self.get_music_library_information(*args, **kwargs)
+        return self.music_library.get_music_library_information(*args,
+                                                                **kwargs)
 
     @only_on_master
     def add_uri_to_queue(self, uri):
@@ -1597,6 +1600,248 @@ class SoCo(_SocoSingletonBase):
 
     def _music_lib_search(self, search, start, max_items):
         return self.music_library._music_lib_search(search, start, max_items)
+
+    @only_on_master
+    def reorder_sonos_playlist(self, sonos_playlist, tracks, new_pos,
+                               update_id=0):
+        """Reorder and/or Remove tracks in a Sonos playlist.
+
+        The underlying call is quite complex as it can both move a track
+        within the list or delete a track from the playlist.  All of this
+        depends on what tracks and new_pos specify.
+
+        If a list is specified for tracks, then a list must be used for
+        new_pos. Each list element is a discrete modification and the next
+        list operation must anticipate the new state of the playlist.
+
+        If a comma formatted string to tracks is specified, then use
+        a similiar string to specify new_pos. Those operations should be
+        ordered from the end of the list to the beginning
+
+        See the helper methods
+        :py:meth:`clear_sonos_playlist`, :py:meth:`move_in_sonos_playlist`,
+        :py:meth:`remove_from_sonos_playlist` for simplified usage.
+
+        update_id - If you have a series of operations, tracking the update_id
+        and setting it, will save a lookup operation.
+
+        Examples:
+          To reorder the first two tracks::
+
+            # sonos_playlist specified by the DidlPlaylistContainer object
+            sonos_playlist = device.get_sonos_playlists()[0]
+            device.reorder_sonos_playlist(sonos_playlist,
+                                          tracks=[0, ], new_pos=[1, ])
+            # OR specified by the item_id
+            device.reorder_sonos_playlist('SQ:0', tracks=[0, ], new_pos=[1, ])
+
+          To delete the second track::
+
+            # tracks/new_pos are a list of int
+            device.reorder_sonos_playlist(sonos_playlist,
+                                          tracks=[1, ], new_pos=[None, ])
+            # OR tracks/new_pos are a list of int-like
+            device.reorder_sonos_playlist(sonos_playlist,
+                                          tracks=['1', ], new_pos=['', ])
+            # OR tracks/new_pos are strings - no transform is done
+            device.reorder_sonos_playlist(sonos_playlist, tracks='1',
+                                          new_pos='')
+
+          To reverse the order of a playlist with 4 items::
+
+            device.reorder_sonos_playlist(sonos_playlist, tracks='3,2,1,0',
+                                          new_pos='0,1,2,3')
+
+        Args:
+            sonos_playlist
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`): The
+                Sonos playlist object or the item_id (str) of the Sonos
+                playlist.
+            tracks: (list): list of track indices(int) to reorder. May also be
+                a list of int like things. i.e. ``['0', '1',]`` OR it may be a
+                str of comma separated int like things. ``"0,1"``.  Tracks are
+                **0**-based. Meaning the first track is track 0, just like
+                indexing into a Python list.
+            new_pos (list): list of new positions (int|None)
+                corresponding to track_list. MUST be the same type as
+                ``tracks``. **0**-based, see tracks above. ``None`` is the
+                indicator to remove the track. If using a list of strings,
+                then a remove is indicated by an empty string.
+            update_id (int): operation id (default: 0) If set to 0, a lookup
+                is done to find the correct value.
+
+        Returns:
+            dict: Which contains 3 elements: change, length and update_id.
+                Change in size between original playlist and the resulting
+                playlist, the length of resulting playlist, and the new
+                update_id.
+
+        Raises:
+            SoCoUPnPException: If playlist does not exist or if your tracks
+                and/or new_pos arguments are invalid.
+        """
+        # allow either a string 'SQ:10' or an object with item_id attribute.
+        object_id = getattr(sonos_playlist, 'item_id', sonos_playlist)
+
+        if isinstance(tracks, UnicodeType):
+            track_list = [tracks, ]
+            position_list = [new_pos, ]
+        elif isinstance(tracks, int):
+            track_list = [tracks, ]
+            if new_pos is None:
+                new_pos = ''
+            position_list = [new_pos, ]
+        else:
+            track_list = [str(x) for x in tracks]
+            position_list = [str(x) if x is not None else '' for x in new_pos]
+        # track_list = ','.join(track_list)
+        # position_list = ','.join(position_list)
+        if update_id == 0:  # retrieve the update id for the object
+            response, _ = self._music_lib_search(object_id, 0, 1)
+            update_id = response['UpdateID']
+        change = 0
+
+        for track, position in zip(track_list, position_list):
+            if track == position:   # there is no move, a no-op
+                continue
+            response = self.avTransport.ReorderTracksInSavedQueue([
+                ("InstanceID", 0),
+                ("ObjectID", object_id),
+                ("UpdateID", update_id),
+                ("TrackList", track),
+                ("NewPositionList", position),
+            ])
+            change += int(response['QueueLengthChange'])
+            update_id = int(response['NewUpdateID'])
+        length = int(response['NewQueueLength'])
+        response = {'change': change,
+                    'update_id': update_id,
+                    'length': length}
+        return response
+
+    @only_on_master
+    def clear_sonos_playlist(self, sonos_playlist, update_id=0):
+        """Clear all tracks from a Sonos playlist.
+        This is a convenience method for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            device.clear_sonos_playlist(sonos_playlist)
+
+        Args:
+            sonos_playlist
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`):
+                Sonos playlist object or the item_id (str) of the Sonos
+                playlist.
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            dict: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            ValueError: If sonos_playlist specified by string and is not found.
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        if not isinstance(sonos_playlist, DidlPlaylistContainer):
+            sonos_playlist = self.get_sonos_playlist_by_attr('item_id',
+                                                             sonos_playlist)
+        count = self.music_library.browse(ml_item=sonos_playlist).total_matches
+        tracks = ','.join([str(x) for x in range(count)])
+        if tracks:
+            return self.reorder_sonos_playlist(sonos_playlist, tracks=tracks,
+                                               new_pos='', update_id=update_id)
+        else:
+            return {'change': 0, 'update_id': update_id, 'length': count}
+
+    @only_on_master
+    def move_in_sonos_playlist(self, sonos_playlist, track, new_pos,
+                               update_id=0):
+        """Move a track to a new position within a Sonos Playlist.
+        This is a convenience method for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            device.move_in_sonos_playlist(sonos_playlist, track=0, new_pos=1)
+
+        Args:
+            sonos_playlist
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`):
+                Sonos playlist object or the item_id (str) of the Sonos
+                playlist.
+            track (int): **0**-based position of the track to move. The first
+                track is track 0, just like indexing into a Python list.
+            new_pos (int): **0**-based location to move the track.
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            dict: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        return self.reorder_sonos_playlist(sonos_playlist, int(track),
+                                           int(new_pos), update_id)
+
+    @only_on_master
+    def remove_from_sonos_playlist(self, sonos_playlist, track, update_id=0):
+        """Remove a track from a Sonos Playlist.
+        This is a convenience method for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            device.remove_from_sonos_playlist(sonos_playlist, track=0)
+
+        Args:
+            sonos_playlist
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`):
+                Sonos playlist object or the item_id (str) of the Sonos
+                playlist.
+            track (int): *0**-based position of the track to move. The first
+                track is track 0, just like indexing into a Python list.
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            dict: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        return self.reorder_sonos_playlist(sonos_playlist, int(track), None,
+                                           update_id)
+
+    @only_on_master
+    def get_sonos_playlist_by_attr(self, attr_name, match):
+        """Return the first Sonos Playlist DidlPlaylistContainer that
+        matches the attribute specified.
+
+        Args:
+            attr_name (str): DidlPlaylistContainer attribute to compare. The
+                most useful being: 'title' and 'item_id'.
+            match (str): Value to match.
+
+        Returns:
+            (:class:`~.soco.data_structures.DidlPlaylistContainer`): The
+                first matching playlist object.
+
+        Raises:
+            (AttributeError): If indicated attribute name does not exist.
+            (ValueError): If a match can not be found.
+
+        Example::
+
+            device.get_sonos_playlist_by_attr('title', 'Foo')
+            device.get_sonos_playlist_by_attr('item_id', 'SQ:3')
+
+        """
+        for sonos_playlist in self.get_sonos_playlists():
+            if getattr(sonos_playlist, attr_name) == match:
+                return sonos_playlist
+        raise ValueError('No match on "{0}" for value "{1}"'.format(attr_name,
+                                                                    match))
+
 
 # definition section
 
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -391,7 +391,8 @@ class TestSonosPlaylist(object):
         """A fixture which cleans up after each sonos playlist test."""
         if self.existing_playlists is None:
             self.existing_playlists = soco.get_sonos_playlists()
-            if self.playlist_name in [x.title for x in self.existing_playlists]:
+            if self.playlist_name in [x.title
+                                      for x in self.existing_playlists]:
                 msg = '%s is an existing playlist.' % self.playlist_name
                 pytest.fail(msg)
 
@@ -413,10 +414,10 @@ class TestSonosPlaylist(object):
 
     def test_create_from_queue(self, soco):
         """Test creating a Sonos playlist from the current queue."""
-        new_playlist = soco.create_sonos_playlist_from_queue(self.playlist_name)
-        assert type(new_playlist) is DidlPlaylistContainer
+        playlist = soco.create_sonos_playlist_from_queue(self.playlist_name)
+        assert type(playlist) is DidlPlaylistContainer
 
-        prslt = soco.browse(ml_item=new_playlist)
+        prslt = soco.browse(ml_item=playlist)
         qrslt = soco.get_queue()
         assert len(prslt) == len(qrslt)
         assert prslt.total_matches == qrslt.total_matches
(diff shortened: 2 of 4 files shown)


dundeemt:
blocking on issue #353, I have a commit to fix the flake8 errors but I can't run my tests since .get_sonos_playlists() is now failing.

Other: my local flake8 is passing without issue, it fails on the CI.  I've got corrections but am waiting to commit until I can get my tests to start running again.

lawrenceakka:
I have fixed #353 - sorry about that. Sonos can't make its mind up about whether a playlistContainer is a `<container>` or an `<item>`. I re-ran the Travis job, but you are still failing lint checking :disappointed:

dundeemt:
@lawrenceakka no worries.  Working with Sonos can be interesting.  I need to go through the CI script and figure out what the difference is between flake8 on Travis and me running it locally.  The mismatched closed was not showing locally when I ran pep8 and flake8 against my local source.  

Yeah, soco/compat.py -- solves my `unicode` problem.

Is the pylint check new or changed?  I haven't made any changes to data_structures.py in the PR.

dundeemt:
Looked at data_structures.DidlObject.from_element, the source of the pylint R0914 too-many-locals error and I am not sure what can be reasonably done outside of telling pylint to ignore it for that method.  The code looks tight and readable.

lawrenceakka:
@dundeemt I have no idea why pylint has suddenly started complaining. I have just run up against this too.  Stick a `#pylint: disable=too-many-locals` at an appropriate place.

Your patch looks like it will be very useful.  I will provide some inline comments when I get a moment. One thing I am not so sure about is the syntax for specifying multiple tracks. I would much prefer to use something more like Python's slice syntax, instead of inventing something new. You will see that I have just put in a PR (#354) to change this for ListofMusicItems (and hence the Queue class), so that they can support multiple slice indices (e.g. `playlist[1:3, 6, 20:12:-1]` will return a list of the second and third, fifth, and 20th to 13th elements of the playlist. 

Lawrence

soco/core.py
```diff
@@ -134,6 +135,7 @@ class SoCo(_SocoSingletonBase):
         get_composers -- Get composers from the music library
         get_tracks -- Get tracks from the music library
         get_playlists -- Get playlists from the music library
+        get_sonos_playlists -- Get Sonos playlists from the music library.
```
lawrenceakka: It is not getting playlists from the music library (which, strictly, is the locally stored music, i.e. not Spotify etc).

soco/core.py
```diff
@@ -1255,6 +1266,34 @@ def get_sonos_playlists(self, *args, **kwargs):
         args = tuple(['sonos_playlists'] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
+    def get_sonos_playlist_by_attr(self, attr_name, match):
+        """ Return the first Sonos Playlist DidlPlaylistContainer that
+        matches the attribute specified.
+
+        Args:
+            attr_name (str): DidlPlaylistContainer attribute to compare.
```
lawrenceakka: "e.g. title, or ..." would be useful here

soco/core.py
```diff
@@ -1255,6 +1266,34 @@ def get_sonos_playlists(self, *args, **kwargs):
...
+            match (str): Value to match.
+
+        Returns:
+            (:py:class:`~.soco.data_structures.DidlPlaylistContainer`): The
```
lawrenceakka: I don't think you need the `:py:` here since python is the default in Sphinx. Do you need the "~"? I haven't checked

dundeemt:
This PR has a number of areas where I expect to get feedback from the group.  However, it covers an area currently under-served by soco, Sonos Playlist management.  

The underlying service is complex and not well documented and i did my best to tame it but still retain the ability to use it optimally.  I've written simplified helper functions for the 3 most common operations, remove a track, move a track and clear the Sonos playlist.  

Areas of note:
 * Indexing - I've used 0-based indexing.  The underlying services refer to the first item in a Sonos Playlist as 0, Python lists use 0 based indexing.  This is in line with the other 'queue' management methods.

* update_id - I am still unsure just how important it is, but the existing add_item_to_sonos_playlist methods deemed it important.  I allow the calling code to set it so a lookup can be avoided, the input parameter is optional and defaults to 0.  I also return it as part of every Sonos playlist mgmt method so the caller can pass it on to the next call if a multi operation is being done.

* object_id - allows for a simple string with the item_id but also accepts an object that has an item_id attribute. I wanted to make it simple for the caller to use what they had available.

* tracks/new_pos 
  * reorder_sonos_playlists - takes a list of ints, a list of int-like or a string containing a comma separated list of numbers.  The tracks/new_pos parameters are expected to be of the same logical type.
  * helper functions accept simple int or int-likes.  (i.e. 1 vs '1')

soco/core.py
```diff
@@ -1255,6 +1266,34 @@ def get_sonos_playlists(self, *args, **kwargs):
...
+                first matching playlist object.
+
+        Raises:
+            (AttributeError): If indicated attribute name does not exist.
+            (ValueError): If a match can not be found.
+
+        Example::
+
+            obj.get_sonos_playlist_by_attr('item_id', 'SQ:3')
```
lawrenceakka: Please use `device` instead of `obj`.  We will eventually standardise this, but you have to start somewhere!

soco/core.py
```diff
@@ -1833,6 +1872,181 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
             ('AddAtIndex', 4294967295)
         ])
 
+    def reorder_sonos_playlist(self, object_id, tracks, new_pos, update_id=0):
+        """ Reorder and/or Remove tracks in a Sonos playlist. The underlying
+        call is quite complex. See the helper methods
+        :py:meth:`clear_sonos_playlist`, :py:meth:`move_in_sonos_playlist`,
+        :py:meth:`remove_from_sonos_playlist` for simplified usage.
+
+        update_id - If you have a series of operations, tracking the update_id
```
lawrenceakka: Please can you explain more - if not in the doctstring then in a normal comment

soco/core.py
dundeemt: Just using (DidlPlaylistContainer) resulted in no hyperlink.  So I just cut an pasted from another docstring. However, only newer methods are using Napoleon so it most likely came from a different style docstring.  I'll try and remove and see what it does.

soco/core.py
```diff
@@ -1833,6 +1872,181 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or the Sonos
+                playlist object
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`).
+            tracks: (list): list of track indices(int) to reorder. May also be
+                a list of int like things. i.e. ``['0', '1',]`` OR it may be a
+                str of comma separated int like things. ``"0,1"``.  Tracks are
+                **0**-based. Meaning the first track is track 0, just like
+                indexing into a Python list.
```
lawrenceakka: See my comments about multiple slice lists.  Can we do something more pythonic, e.g. `[1, 4:8, 9]` - see also #354

soco/core.py
dundeemt: It wasn't originally listed, so I added it to the main docstring.  I've
modified it now say:
  get_sonos_playlists -- Get Sonos playlists.

Too little or just right?

On Mon, Sep 7, 2015 at 10:44 AM, Lawrence Akka <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r38870472>:
>
> > @@ -134,6 +135,7 @@ class SoCo(_SocoSingletonBase):
> >          get_composers -- Get composers from the music library
> >          get_tracks -- Get tracks from the music library
> >          get_playlists -- Get playlists from the music library
> > +        get_sonos_playlists -- Get Sonos playlists from the music library.
>
> It is not getting playlists from the music library (which, strictly, is
> the locally stored music, i.e. not Spotify etc).
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r38870472>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: I've modified the string to:
            attr_name (str): DidlPlaylistContainer attribute to compare.
The
                most useful being: 'title' and 'item_id'.


On Mon, Sep 7, 2015 at 10:46 AM, Lawrence Akka <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r38870587>:
>
> > @@ -1255,6 +1266,34 @@ def get_sonos_playlists(self, *args, **kwargs):
> >          args = tuple(['sonos_playlists'] + list(args))
> >          return self.get_music_library_information(*args, **kwargs)
> >
> > +    def get_sonos_playlist_by_attr(self, attr_name, match):
> > +        """ Return the first Sonos Playlist DidlPlaylistContainer that
> > +        matches the attribute specified.
> > +
> > +        Args:
> > +            attr_name (str): DidlPlaylistContainer attribute to compare.
>
> "e.g. title, or ..." would be useful here
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r38870587>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: I've expanded the doc string some what.  But I'll explain in more detail
here and based on feedback, I'll remodify the docstring.

The call can reorder and delete tracks.  In general the underlying call
expects two string arguments, tracks which indicate the 'current' track
position you want to modify and the new_pos that dictate what should happen
to it.  You can, in a single request, specify moving and deleting, but they
happen grouped into two calls.  First doing all the delete requests in an
op, and then doing the moves (however that is with respect to the new track
positions after the deletes.  If you delete a track, all following tracks
are immediately renumbered to fill in the gap. So any move requests need to
take this into account.  It is simpler to think about if you make the
changes individually.

So the first option of list of ints.   tracks=[1,3], new_pos=[0, None]
(since it is using ints, I chose to indicate a delete with None) The
results of this operation would move the second track into the first
position, what was in the 1 position moves to the second postion and the
4th track is deleted from the playlist.

trackA       pos:0
trackB       pos:1
trackC       pos:2
trackD       pos:3

is transformed into:
trackB      pos:0
trackA      pos:1
trackC      pos:2

When the inputs are lists, each pair zip(tracks, new_pos) creates an
underlying call to avTransport.ReorderTracksInSavedQueue.

The same call using a list of int-likes  would look like:  tracks=['1',
'3'], new_pos['0', ''], which would also create two underlying calls.
(note: when using int-likes a delete is specified by an empty string.)

Now, if you really wanted to have fun, you can call reorder with strings,
tracks='1,3,5', new_pos='0,2,4' which would move the 2nd, 4th and 6th track
into the first, third and 5th position in a single underlying call. Using a
string, says: "I know what I'm doing, get out of my way."  It seems to
happen "atomically"  in that the moves happen before the track order
shifts.  However, getting this right can be frustrating when you have
deletes going on at the same time.   Using the helper functions clear, move
and remove make it simpler to think about because the list is being
transformed one step at a time and you only need to supply simple parameters

maybe we should prepend an underscore to reorder_sonos_playlist ->
_reorder_sonos_playlist. To encourage people to use just the helpers. ??

On Mon, Sep 7, 2015 at 10:48 AM, Lawrence Akka <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r38870720>:
>
> > @@ -1833,6 +1872,181 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
> >              ('AddAtIndex', 4294967295)
> >          ])
> >
> > +    def reorder_sonos_playlist(self, object_id, tracks, new_pos, update_id=0):
> > +        """ Reorder and/or Remove tracks in a Sonos playlist. The underlying
> > +        call is quite complex. See the helper methods
> > +        :py:meth:`clear_sonos_playlist`, :py:meth:`move_in_sonos_playlist`,
> > +        :py:meth:`remove_from_sonos_playlist` for simplified usage.
> > +
> > +        update_id - If you have a series of operations, tracking the update_id
>
> Please can you explain more - if not in the doctstring then in a normal
> comment
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r38870720>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

dundeemt:
@lawrenceakka  re #354 A read-only structure of course does not make sense for a sonos playlist, it has to be modifiable, after all that what this PR is all about.  I do get your point though about more Pythonic. In general it would be an object that maintained it's state with the backend.  In that regards, it is closer to an ORM than just a subclass of a standard python object (list, queue, ..).  For this PR I was looking to bring parity to Sonos playlists with the current queue methods  ....

soco/core.py
KennethNielsen: I don't think this changes with Napoleop. Napoleon is mostly about formatting the list of Args, Returns, and Raises and their type.

KennethNielsen:
Hallo @dundeemt 

This looks really good and definitely something we have been missing. I have a few comments about style and preferred arguments and such. I will comment in-line.

soco/core.py
KennethNielsen: I don't know if I wouldn't prefer an title example instead. Most of the time, users should not have to meddle with these IDs themselves.

soco/core.py
```diff
@@ -1255,6 +1266,35 @@ def get_sonos_playlists(self, *args, **kwargs):
...
+            (AttributeError): If indicated attribute name does not exist.
+            (ValueError): If a match can not be found.
+
+        Example::
+
+            obj.get_sonos_playlist_by_attr('item_id', 'SQ:3')
+            obj.get_sonos_playlist_by_attr('title', 'Foo')
+
+        """
+        for spl in self.get_sonos_playlists():
```
KennethNielsen: In general, please write variable names out, especially if we are not short on space. attr for attribute is fine, because it is an established shorthand, but ``spl`` really should be ``sonos_playlist``. "Readability counts" ;)

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
             ('AddAtIndex', 4294967295)
         ])
 
+    def reorder_sonos_playlist(self, object_id, tracks, new_pos, update_id=0):
```
KennethNielsen: Not only here, but in all methods. I like that you have made the argument that specifies the Sonos playlist flexible, so that it takes both an ID and the playlist object, but please turn the logic around, so that the playlist object is the first class citizen and the ID the alternative, both in the argument name and in the phrasing in the docstring. As I mentioned in one of the other PRs, I think one of the goals of the SoCo API should be to elegantly mask some of all the gory details of Sonos by clear Pythonic types and concepts and I really feel that this falls in that category.

I will make smaller comments with the keyword playlist_object_first in the places were I think there should be a modification.

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        """ Reorder and/or Remove tracks in a Sonos playlist. The underlying
+        call is quite complex as it both can move a track within the list or
+        delete a track from the playlist.  All of this depends on what tracks
+        and new_pos specify.
+
+        See the helper methods
+        :py:meth:`clear_sonos_playlist`, :py:meth:`move_in_sonos_playlist`,
+        :py:meth:`remove_from_sonos_playlist` for simplified usage.
```
KennethNielsen: Nice. I like this. Make the advanced methods available and with clear description along with the helpers :)

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        update_id - If you have a series of operations, tracking the update_id
+        and setting it, will save a lookup operation.
+
+        Examples:
+          To reorder the first two tracks::
+
+            # object_id specified by the item_id
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            obj.reorder_sonos_playlist('SQ:0', tracks=[0, ], new_pos=[1, ])
+            # OR object_id specified by the DidlPlaylistContainer object
+            spl = obj.get_sonos_playlists()[0]
```
KennethNielsen: please wirte out spl

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            obj.reorder_sonos_playlist(spl, tracks=[0, ], new_pos=[1, ])
+
+          To delete the second track::
+
+            # tracks/new_pos are a list of int
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            obj.reorder_sonos_playlist('SQ:0', tracks=[1, ], new_pos=[None, ])
+            # OR tracks/new_pos are a list of int-like
+            obj.reorder_sonos_playlist('SQ:0', tracks=['1', ], new_pos=['', ])
+            # OR tracks/new_pos are strings - no transform is done
+            obj.reorder_sonos_playlist('SQ:0', tracks='1', new_pos='')
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or the Sonos
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            obj.reorder_sonos_playlist(spl, tracks=[0, ], new_pos=[1, ])
+
+          To delete the second track::
+
+            # tracks/new_pos are a list of int
+            obj.reorder_sonos_playlist('SQ:0', tracks=[1, ], new_pos=[None, ])
+            # OR tracks/new_pos are a list of int-like
+            obj.reorder_sonos_playlist('SQ:0', tracks=['1', ], new_pos=['', ])
+            # OR tracks/new_pos are strings - no transform is done
+            obj.reorder_sonos_playlist('SQ:0', tracks='1', new_pos='')
```
KennethNielsen: I'm really not sure whether exposing this is more confusing than it is worth. O.k. so this is for performing a single operation. This is already doable with the other methods, so what it saves really is only typing the lists around the single arguments. The downside is that with the ``no transform is done`` comment, you are implying and expectation of knowledge about the underlying protocol. I think, if we are to keep it, scratch the no transform comment.

On another note, I would suggest to make all these multiple track examples, some like to code by example, and it is always easier to infer the singular than the plural

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or the Sonos
+                playlist object
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`).
+            tracks: (list): list of track indices(int) to reorder. May also be
+                a list of int like things. i.e. ``['0', '1',]`` OR it may be a
+                str of comma separated int like things. ``"0,1"``.  Tracks are
```
KennethNielsen: as above, suggest to remove the string input form

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+                is done to find the correct value.
+
+        Returns:
+            tuple: A  tuple(int, int, int) containing 3 values:
+                Change in size between original playlist and the resulting
+                playlist, the length of resulting playlist, and the new
+                update_id.
+
+        Raises:
+            SoCoUPnPException: If object_id does not exist.
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        """
+        # allow either a string 'SQ:10' or an object with item_id attribute.
+        object_id = getattr(object_id, 'item_id', object_id)
+
+        if isinstance(tracks, UnicodeType):
+            track_list = [tracks, ]
+            position_list = [new_pos, ]
+        elif type(tracks) is int:
+            track_list = [tracks, ]
+            position_list = [new_pos or '', ]
```
KennethNielsen: Is this correct, remember 0 also evaluates to false:
```python
>>> 0 or ''
''
```
that would mean that you cannot use the non-iterable argument form to move a track to the first position.

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        else:
+            track_list = [str(x) for x in tracks]
+            position_list = [str(x) if x is not None else '' for x in new_pos]
+
+        if update_id == 0:  # retrieve the update id for the object
+            response, _ = self._music_lib_search(object_id, 0, 1)
+            update_id = response['UpdateID']
+        change = 0
+
+        for track, position in zip(track_list, position_list):
```
KennethNielsen: Here we need to careful. When I wrote above, that I was unsure if you handled the non-iterable case correctly, it was because I was unsure if the under lying call takes a single position or a list of position. I now see that it takes a list, which of course means that the non-iterable input case is correct.

However, if it does take a list, I would highly suggest to turn all iterable input args into a list and execute only one request. The reason for this is two-fold:

 1. Performance, we should of course never prematurely optimize, but I think this just makes sense. For apps doing network IO, that will always be the "slow thing" and since it is made so easy to do all at once we might as well do it.
 2. Deletions. The significantly more important issue, is that if we do only do one request, we can hope that everything will also be executed as one query on the other end and this is important for deletions. We went through this somewhere else, where the call only took one argument, maybe it was for queue manipulation. Consider this, if I ask this method to delete tracks 0 and 3. After having deleted track 0, the second track that I wanted to delete is now in position 2 **not** 3. This can be fixed by sorting deletions descending, but here, at least of it works, it would be soooo much easier to just do one request. It may also be, that they actually iterate on the other end, in which case, we have to handle this problem here.

Deletions stuff, and the example above, should be tested.

The one thing that I do not know is, if the list form accepts mixing operations, move and delete, or if it might separate them out on the other end. This will make everything more complicated.

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+                ("TrackList", track),
+                ("NewPositionList", position),
+            ])
+            change += int(response['QueueLengthChange'])
+            update_id = int(response['NewUpdateID'])
+        length = int(response['NewQueueLength'])
+
+        return change, length, update_id
+
+    def clear_sonos_playlist(self, object_id, update_id=0):
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        """ Clear all tracks from a Sonos playlist.
+        This is a convenience function for :py:meth:`reorder_sonos_playlist`.
+
+        Example:
+            ``obj.clear_sonos_playlist('SQ:0')``
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            ])
+            change += int(response['QueueLengthChange'])
+            update_id = int(response['NewUpdateID'])
+        length = int(response['NewQueueLength'])
+
+        return change, length, update_id
+
+    def clear_sonos_playlist(self, object_id, update_id=0):
+        """ Clear all tracks from a Sonos playlist.
+        This is a convenience function for :py:meth:`reorder_sonos_playlist`.
```
KennethNielsen: strictly speaking "convenience method"

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        Example:
+            ``obj.clear_sonos_playlist('SQ:0')``
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            tuple: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        sonos_pl = None
```
KennethNielsen: ``sons_pl`` and ``spl`` can be written out more, plenty of space

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        else:
+            for spl in self.get_sonos_playlists():
+                if spl.item_id == object_id:
+                    sonos_pl = spl
+        if sonos_pl is None:
+            raise ValueError('"{0}" is not a valid Sonos Playlist'
+                             ' item_id'.format(object_id))
+        num_tracks = self.browse(ml_item=sonos_pl).total_matches
+        tracks = ','.join([str(x) for x in range(num_tracks)])
+        return self.reorder_sonos_playlist(sonos_pl, tracks=tracks,
```
KennethNielsen: O.k. I see here that you use the non-iterable input form, but it would only be one extra line not to, if it were to be removed. If this works however, this does seem to indicate that one request is handled as one *query* on the other end, otherwise, this would not work, it would run into non-existing positions along th way.

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+                                           new_pos='', update_id=update_id)
+
+    def move_in_sonos_playlist(self, object_id, track, new_pos, update_id=0):
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        """ Move a track to a new position within a Sonos Playlist.
+        This is a convenience function for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            obj.move_in_sonos_playlist('SQ:0', track=0, new_pos=1)
```
KennethNielsen: as above for Lwarence comment about ``device``

also playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            tuple: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        return self.reorder_sonos_playlist(object_id, int(track),
```
KennethNielsen: Here you actually require ints and arguments, but you also type convert.

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+                                           int(new_pos), update_id)
+
+    def remove_from_sonos_playlist(self, object_id, track, update_id=0):
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+        """ Remove a track from a Sonos Playlist.
+        This is a convenience function for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            obj.remove_from_sonos_playlist('SQ:0', track=0)
```
KennethNielsen: device and playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+
+        Args:
+            object_id (str): The item_id of the Sonos playlist or
```
KennethNielsen: playlist_object_first

soco/core.py
```diff
@@ -1833,6 +1873,185 @@ def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
...
+            update_id (int): Optional update counter for the object. If left
+                at the default of 0, it will be looked up.
+
+        Returns:
+            tuple: See :py:meth:`reorder_sonos_playlist`
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        return self.reorder_sonos_playlist(object_id, int(track), None,
```
KennethNielsen: as above about requiring ints and type converting

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
             'No difference between '
             'queues before and after removing the last item')
         assert len(new_queue) == len(old_queue) - 1
+
+
+class TestReorderSonosPlaylist(object):
+    """ Integration tests for Sonos Playlist Management. """
+    existing_playlists = None
+    pl_name = 'zSocoTestPlayList42'
```
KennethNielsen: also for test code about writing variable names out, I will not comment any further on this

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+                return spl
+
+    @pytest.yield_fixture(autouse=True, scope="class")
+    def restore_sonos_playlists(self, soco):
+        """ A fixture which cleans up after each sonos playlist test."""
+        if self.existing_playlists is None:
+            self.existing_playlists = soco.get_sonos_playlists()
+        if self.pl_name in [x.title for x in self.existing_playlists]:
+            msg = '%s is an existing playlist.' % self.pl_name
+            pytest.fail(msg)
```
KennethNielsen: If this is meant for the unlikely case that someone might have a playlist by this name, then you could as python for a UUID. If it is just meant to check for missing cleanup after a failed test, then maybe the error message should say so.

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+        soco.contentDirectory.DestroyObject([('ObjectID',
+                                              self.test_pl.item_id)])
+
+    def _reset_spl_contents(self, soco):
+        """ Ensure test playlist matches queue for each test. """
+        update_id = 0
+        que = soco.get_queue()
+        test_pl = self._get_test_pl(soco)
+        spl = soco.browse(ml_item=test_pl)
+        tracks = list(reversed(range(len(spl))))
```
KennethNielsen: Ahhh, I see that you were aware of the deletion problem

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+        return test_pl
+
+    def test_reverse_track_order(self, soco):
+        """ Test reversing the tracks in the Sonos playlist. """
+        test_pl = self._reset_spl_contents(soco)
+        num_tracks = soco.browse(ml_item=test_pl).total_matches
+        tracks = range(num_tracks)
+        new_pos = list(reversed(tracks))
+
+        args = {'object_id': test_pl.item_id,
```
KennethNielsen: or maybe just test_pl ;)

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+                'tracks': tracks,
+                'new_pos': new_pos}
+        change, count, update_id = soco.reorder_sonos_playlist(**args)
+        assert change == 0
+        assert count == num_tracks
+        assert update_id != 0
+        spl = soco.browse(ml_item=test_pl)
+        for s_item, q_item in zip(spl, reversed(soco.get_queue())):
```
KennethNielsen: Testing up against queue here, means that you have to know that the playlist is reset to something that comes from the queue, what about just testing up against test_pl?

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+            assert s_item.resources[0].uri != q_item.resources[0].uri
+
+    def test_swap_first_two_items(self, soco):
+        """ Test a use case in doc string. Swapping the positions of the first
```
KennethNielsen: Test a usecase in docstring ???

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+        spl = soco.browse(ml_item=test_pl)
+        que = soco.get_queue()[1:]
+        for s_item, q_item in zip(spl, que):
+            assert s_item.resources[0].uri == q_item.resources[0].uri
+
+    def test_remove_first_track_full(self, soco):
+        """ Test removing first track from Sonos Playlist. """
+        test_pl = self._reset_spl_contents(soco)
+        num_tracks = soco.browse(ml_item=test_pl).total_matches
+        tracks = [0] + range(num_tracks - 1)    # [0, 0, 1, ..., n-1]
```
KennethNielsen: Does this fail if you go all the way up to num_tracks? Just curious as this might shed some light on how the Sonos devices executes this.

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+        del que[ndx]
+        for s_item, q_item in zip(spl, que):
+            assert s_item.resources[0].uri == q_item.resources[0].uri
+
+    def test_remove_some_tracks(self, soco):
+        """ Test removing some tracks from Sonos Playlist. """
+        test_pl = self._reset_spl_contents(soco)
+        num_tracks = soco.browse(ml_item=test_pl).total_matches
+        # get rid of the even numbered tracks
+        tracks = sorted([x for x in range(num_tracks) if not x & 1],
```
KennethNielsen: I'd prefer the form without bitwise operators for something like this: if x % 2 == 1

tests/test_integration.py
```diff
@@ -378,3 +378,362 @@ def test(self, soco):
...
+                                            update_id=update_id)
+            if count:
+                msg = 'Failed to empty playlist. Unexpected result.'
+                pytest.fail(msg)
+        for track in que:
+            soco.add_item_to_sonos_playlist(track, test_pl)
+        return test_pl
+
+    def test_reverse_track_order(self, soco):
+        """ Test reversing the tracks in the Sonos playlist. """
```
KennethNielsen: I think @lawrenceakka is in the process of correcting all the docstrings in another PR. I think the correct form is ```"""Test reversing the tracks in the Sonos playlist."""```. Followed by full stop, but no space before or after.

KennethNielsen:
Hallo @dundeemt 

First of all, really nice work. I especially enjoy the generous amount of integration tests \o/ (which really are functional tests, I will have to rename this soon).

I left some inline comments. Forgive the number of them, I tend to be a bit nit-picky. If you disagree with any of them, just let me know so we can discuss.

A few general thoughts came up while reviewing.

As you may have noticed, I was rambling on and on about whet the Sonos devices actually do, when you hand them a list of arguments. If it forms a single query to its internal database or executes them one by one. The more I read on, it seems as if you have actually concluded, that it executes them one by one. If I am correct in this assumption, then:

 1. I think it is all the more prudent to collect interable inputs into a single argument and always make just the one request, as that is what will be done on the other side anyway and it will have a (potentially large) performance impact.
 2. I think this one-by-one execution mode **and** its implications for deletions **and** the fact that it happens whether the user suppliers lists or strs should be clearly highlighted in the docstring, maybe with a Note or Warning box.

With regards to @lawrenceakka comments about using indexing. While I agree that it maybe could create a nicer interface, I still think we should also have the normal methods as implemented here. So that in the future, one would be able to to either or.

soco/core.py
dundeemt: Using this method is the best for removing a bunch of tracks.  It is used by clear_sonos_playlist to remove all tracks in a single operation.  If you had N tracks, without this capability, you would make N network calls.  Using this no-transform method there is 1 call.   A very substantial speed up. This call would have `tracks='1,2,...,n'` and `new_pos=''`

soco/core.py
dundeemt: That might be possible but I would like to get capabilities that put sonos_playlists on par with queues first.

soco/core.py
```diff
@@ -1255,6 +1266,35 @@ def get_sonos_playlists(self, *args, **kwargs):
         args = tuple(['sonos_playlists'] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
+    def get_sonos_playlist_by_attr(self, attr_name, match):
+        """ Return the first Sonos Playlist DidlPlaylistContainer that
```
lawrenceakka: Please remove the space at the start!

soco/core.py
lawrenceakka: I would emphasise this even more. I think we need a Sphinx "Warning" that using this method, it is possible to mess up your playlist, if you are not very clear about the counting. You should only use this if you know what you are doing, and you probably want to use the helper methods instead!

soco/core.py
lawrenceakka: Agreed.  I think it is too ambitious to go for slicing at the moment.  I amy work on it though, if I get bored .... :wink:

soco/core.py
lawrenceakka: No space before docstr. please.

tests/test_integration.py
lawrenceakka: Yup.

lawrenceakka:
@dundeemt Sorry you have run into a barrage of comments. I think this is an important addition, and once we can get the few wrinkles ironed out, we will commit.  Please let us know if you have any questions, and feel free to argue back!

soco/core.py
dundeemt: are you going to differentiate between any sonos device and a controller/master?  There is a bit of a taxonomy problem. Because a device can be, a stand alone speaker, a pair, a member of a playbar group, etc.  These methods are only valid against a "master".  I need to add `@only_on_master` on these like I did for the `remove_sonos_playlist` PR that is waiting.

soco/core.py
dundeemt: bug

soco/core.py
dundeemt: Yup, it's a helper.  It meant to hand hold.  The definition also specifies an int.

soco/core.py
dundeemt: also a helper, with well defined track. (int)

tests/test_integration.py
dundeemt: Ok, just following what the prior code had done.

dundeemt:
@KennethNielsen @lawrenceakka  No need to argue back.  I understand your requests and will try and demonstrate the underlying calls unique rule set.

The following underlying calls return Work, for some value of work,  - No Error is returned
```python
[(u'InstanceID', 0), (u'ObjectID', u'SQ:143'), (u'UpdateID', '5'), (u'TrackList', u'3,2,1,0'), (u'NewPositionList', u'0,1,2,3')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:147'), (u'UpdateID', '5'), (u'TrackList', u'0'), (u'NewPositionList', u'3')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:149'), (u'UpdateID', '5'), (u'TrackList', u'2'), (u'NewPositionList', u'1')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:148'), (u'UpdateID', '5'), (u'TrackList', u'1'), (u'NewPositionList', u'2')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:150'), (u'UpdateID', '5'), (u'TrackList', u'3'), (u'NewPositionList', u'0')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:152'), (u'UpdateID', '5'), (u'TrackList', u'3,2'), (u'NewPositionList', u'0,1')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:153'), (u'UpdateID', '5'), (u'TrackList', u'3,1'), (u'NewPositionList', u'0,2')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:154'), (u'UpdateID', '5'), (u'TrackList', u'3,0'), (u'NewPositionList', u'0,3')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:156'), (u'UpdateID', '5'), (u'TrackList', u'3,1,0'), (u'NewPositionList', u'0,2')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:158'), (u'UpdateID', '5'), (u'TrackList', u'0,3,1'), (u'NewPositionList', u',0,2')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:159'), (u'UpdateID', '5'), (u'TrackList', u'3,0,1'), (u'NewPositionList', u'0,,2')]
```

However, the following calls return a UPNP 402 - Invalid Arguments
```python
[(u'InstanceID', 0), (u'ObjectID', u'SQ:144'), (u'UpdateID', '5'), (u'TrackList', u'0,1,2,3'), (u'NewPositionList', u'3,2,1,0')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:145'), (u'UpdateID', '5'), (u'TrackList', u'0,1,2'), (u'NewPositionList', u'3,2,1')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:146'), (u'UpdateID', '5'), (u'TrackList', u'0,1'), (u'NewPositionList', u'3,2')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:151'), (u'UpdateID', '5'), (u'TrackList', u'2,3'), (u'NewPositionList', u'1,0')]

[(u'InstanceID', 0), (u'ObjectID', u'SQ:155'), (u'UpdateID', '5'), (u'TrackList', u'3,0,1'), (u'NewPositionList', u'0,3,2')]
```
It appears, and I encourage you to test it/experiment yourself, that moves must be listed in reverse order while deletes can occur interspersed among moves.  The resulting changes appear as a single unit, but their implementation makes it apparent that this is an illusion.

To take any old list from a caller and attempt to do it in a single call to the underlying method would require that we "sort" their input, before relaying it to the underlying call.  Which is a recipe for trouble and bug reports.  I strongly urge that the helpers, in their simple form (a single int-like parameter for a move, remove) be kept in their simple form and that reorder_sonos_playlist be renamed _reorder_sonos_playlist to reinforce the notion, that you should know what your doing if you are going to use it.

I do not have windows at home so I have not used the desktop package, but the android version only allows you to do one reorder/remove operation at a time.  As for attempting to tie advance slicing operations to this, I have no interest in doing because I would never use it.

I am going to work on the docstrings, object over id, variable names and fixing the two bugs you have found.  I'll get that in place while everyone else has a chance to go over my findings and see if some one has an insight into solving the rules for multiple operations in one call.  Right now, clear_sonos_playlist is the only helper that takes advantage of it.

I am all good on changing the docstring formatting, but it would be good for them to be formalized in dev notes.  It is extremely hard to hit a moving target.  Look for an updated pull request later this weekend.    @KennethNielsen and you think you write long posts ; )

soco/core.py
KennethNielsen: @dundeemt I think as @lawrenceakka said ``device`` will be the term we will prefer. The fact that you can only do all of these actions on the master, should eventually turn out to be implicit.

soco/core.py
KennethNielsen: Yes, but as I mentioned somewhere later, the reorder_sonos_playlist could (and maybe should) simply turn the iterable inputs into strings form, and always perform just one request.

tests/test_integration.py
KennethNielsen: @dundeemt Yeeess :$ not particularly consistent up to this, sorry that this bites you.

KennethNielsen:
> It appears, and I encourage you to test it/experiment yourself, that moves must be listed in reverse order while deletes can occur interspersed among moves. The resulting changes appear as a single unit, but their implementation makes it apparent that this is an illusion.

Oh my!

> To take any old list from a caller and attempt to do it in a single call to the underlying method would require that we "sort" their input, before relaying it to the underlying call. Which is a recipe for trouble and bug reports.

I agree completely!

>  I strongly urge that the helpers, in their simple form (a single int-like parameter for a move, remove) be kept in their simple form

Once again. I agree completely.

> and that reorder_sonos_playlist be renamed _reorder_sonos_playlist to reinforce the notion, that you should know what your doing if you are going to use it.

I don't think that is necessary. I like it that there is a more advanced method that gives you more power. If you simply add a warning to the docstring, along the lines of what you wrote above and your best guess at the rules of execution and saying that advanced users should themselves make sure that their calls does what they want them to, then I see no reason to make it pseudo private.

> I am going to work on the docstrings, object over id, variable names and fixing the two bugs you have found. I'll get that in place while everyone else has a chance to go over my findings and see if some one has an insight into solving the rules for multiple operations in one call. Right now, clear_sonos_playlist is the only helper that takes advantage of it.

I don't think I will have time to experiment myself, but as mentioned above, I'm am happy with the state in which clear_sonos_playlist is the only method that uses it, all other helpers and one by one and the advanced method has a verbose warning as outlined above.

> I am all good on changing the docstring formatting, but it would be good for them to be formalized in dev notes. It is extremely hard to hit a moving target.

Yes, definitely. The docs have for a long time been a weak spot and dev notes is a good idea. More than that, it is especially annoying when a convention is applied while you have an active PR. We apologize for that.

I look forward to the updated pull request. Please drop a comment when you are done.

dundeemt:
Ok, I've udpated doc strings, formatting, put object over id and I think the rest of the suggestions.  However this was a lengthy process, so please do not assume that if a change was requested and is not there that I disagreed, I most likely just missed it. So please bring it back up again.

dundeemt:
I messed up the merge some how and now I am in a hole.  I don't use git that much so its going to take a while to figure out.  Taking a break now.

dundeemt:
changed return type of reorder_sonos_playlist to a dict.  See Issue #348  Updated tests, refactored to remove deprecation warnings due to changes in project structure.

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
     def _music_lib_search(self, search, start, max_items):
         return self.music_library._music_lib_search(search, start, max_items)
 
+    @only_on_master
+    def reorder_sonos_playlist(self, sonos_playlist, tracks, new_pos,
+                               update_id=0):
+        """Reorder and/or Remove tracks in a Sonos playlist.
+
+        The underlying call is quite complex as it both can move a track within
```
KennethNielsen: Shouldn't it be "as it can both"

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+        the list or delete a track from the playlist.  All of this depends on
+        what tracks and new_pos specify.
+
+        If a list is specified for tracks, then a list must be used for
+        new_pos. Each list element is a discrete modification and the next
+        list operation must anticipate the new state of the playlist.
```
KennethNielsen: Very nice.

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+
+        If a comma formatted string to tracks is specified, then use
+        a similiar string to specify new_pos. Those operations should be
```
KennethNielsen: I'm a little unsure what you mean here by "Those operations should be ordered from the end of the list to the beginning"

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+        update_id - If you have a series of operations, tracking the update_id
+        and setting it, will save a lookup operation.
+
+        Examples:
+          To reorder the first two tracks::
+
+            # sonos_playlist specified by the DidlPlaylistContainer object
+            sonos_playlist = device.get_sonos_playlists()[0]
+            device.reorder_sonos_playlist(sonos_playlist,
+                                          tracks=[0, ], new_pos=[1, ])
```
KennethNielsen: On the really nit-picky end of things, these commas in ``[0, ]`` are not necessary when dealing with lists the way they are for tuples. It is considered good style to add these trailing commas for **multi-line** lists (because it will make modifications easier to read in the version control diffs), but I believe not for single line lists.

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+
+          To reverse the order of a playlist with 4 items::
+
+            device.reorder_sonos_playlist(sonos_playlist, tracks='3,2,1,0',
+                                          new_pos='0,1,2,3')
+
+        Args:
+            sonos_playlist
+                (:py:class:`~.soco.data_structures.DidlPlaylistContainer`): The
+                Sonos playlist object or the item_id (str)of the Sonos
```
KennethNielsen: "(str)of" missing space

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+            new_pos (list): list of new positions (int|None)
+                corresponding to track_list. MUST be the same type as
+                ``tracks``. **0**-based, see tracks above. ``None`` is the
+                indicator to remove the track. If using a list of strings,
+                then a remove is indicated by an empty string.
+            update_id (int): operation id (default: 0) If set to 0, a lookup
+                is done to find the correct value.
+
+        Returns:
+            dict: Which contains 3 elements. change, length and update_id.
```
KennethNielsen: "elements" colon, semi-colon?

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+                                           int(new_pos), update_id)
+
+    @only_on_master
+    def remove_from_sonos_playlist(self, sonos_playlist, track, update_id=0):
+        """Remove a track from a Sonos Playlist.
+        This is a convenience method for :py:meth:`reorder_sonos_playlist`.
+
+        Example::
+
+            device.remove_from_sonos_playlist('SQ:0', track=0)
```
KennethNielsen: maybe ``sonos_playlist``here as well

soco/core.py
```diff
@@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
...
+
+        Raises:
+            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
+        """
+        return self.reorder_sonos_playlist(sonos_playlist, int(track), None,
+                                           update_id)
+
+    @only_on_master
+    def get_sonos_playlist_by_attr(self, attr_name, match):
+        """ Return the first Sonos Playlist DidlPlaylistContainer that
```
KennethNielsen: No space before "Return"

KennethNielsen:
@dundeemt Very nice work with the doc strings etc. I'm really happy with the result. I left just a hand full of inline comments on minor issues (typos etc.). Other than those, I consider this ready to merge. Just let us know when.

+1

soco/core.py
dundeemt: Commas make lists unambiguous.  While not required for a single integer,
['this',] != ['this'].  Any iterable will be split into elements within the
list if you don't mark them with commas.  If you would like me to remove
them, I will.  Personally, I don't like to vary based on data types, so I
choose to be explicit.   If you have a pep I'd be happy to read and perhaps
alter my opinion, but I've not read any thing authoritative against it.
 "Explicit is better than implicit." -
https://www.python.org/dev/peps/pep-0020/

On Mon, Sep 14, 2015 at 7:33 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r39387778>:
>
> > +        ordered from the end of the list to the beginning
> > +
> > +        See the helper methods
> > +        :py:meth:`clear_sonos_playlist`, :py:meth:`move_in_sonos_playlist`,
> > +        :py:meth:`remove_from_sonos_playlist` for simplified usage.
> > +
> > +        update_id - If you have a series of operations, tracking the update_id
> > +        and setting it, will save a lookup operation.
> > +
> > +        Examples:
> > +          To reorder the first two tracks::
> > +
> > +            # sonos_playlist specified by the DidlPlaylistContainer object
> > +            sonos_playlist = device.get_sonos_playlists()[0]
> > +            device.reorder_sonos_playlist(sonos_playlist,
> > +                                          tracks=[0, ], new_pos=[1, ])
>
> On the really nit-picky end of things, these commas in [0, ] are not
> necessary when dealing with lists the way they are for tuples. It is
> considered good style to add these trailing commas for *multi-line* lists
> (because it will make modifications easier to read in the version control
> diffs), but I believe not for single line lists.
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r39387778>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: Are you asking that I include an example?  i.e.
To reverse the order of the first 3 tracks, you should use  "2,0", "0,2"
 as compared to "0,2", "2,0".


On Mon, Sep 14, 2015 at 7:27 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r39387395>:
>
> > @@ -1578,6 +1580,246 @@ def _build_album_art_full_uri(self, url):
> >      def _music_lib_search(self, search, start, max_items):
> >          return self.music_library._music_lib_search(search, start, max_items)
> >
> > +    @only_on_master
> > +    def reorder_sonos_playlist(self, sonos_playlist, tracks, new_pos,
> > +                               update_id=0):
> > +        """Reorder and/or Remove tracks in a Sonos playlist.
> > +
> > +        The underlying call is quite complex as it both can move a track within
> > +        the list or delete a track from the playlist.  All of this depends on
> > +        what tracks and new_pos specify.
> > +
> > +        If a list is specified for tracks, then a list must be used for
> > +        new_pos. Each list element is a discrete modification and the next
> > +        list operation must anticipate the new state of the playlist.
> > +
> > +        If a comma formatted string to tracks is specified, then use
> > +        a similiar string to specify new_pos. Those operations should be
>
> I'm a little unsure what you mean here by "Those operations should be
> ordered from the end of the list to the beginning"
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r39387395>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: done

On Mon, Sep 14, 2015 at 7:36 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r39387993>:
>
> > +                playlist.
> > +            tracks: (list): list of track indices(int) to reorder. May also be
> > +                a list of int like things. i.e. ``['0', '1',]`` OR it may be a
> > +                str of comma separated int like things. ``"0,1"``.  Tracks are
> > +                **0**-based. Meaning the first track is track 0, just like
> > +                indexing into a Python list.
> > +            new_pos (list): list of new positions (int|None)
> > +                corresponding to track_list. MUST be the same type as
> > +                ``tracks``. **0**-based, see tracks above. ``None`` is the
> > +                indicator to remove the track. If using a list of strings,
> > +                then a remove is indicated by an empty string.
> > +            update_id (int): operation id (default: 0) If set to 0, a lookup
> > +                is done to find the correct value.
> > +
> > +        Returns:
> > +            dict: Which contains 3 elements. change, length and update_id.
>
> "elements" colon, semi-colon?
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r39387993>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: done

On Mon, Sep 14, 2015 at 7:41 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r39388401>:
>
> > +            dict: See :py:meth:`reorder_sonos_playlist`
> > +
> > +        Raises:
> > +            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
> > +        """
> > +        return self.reorder_sonos_playlist(sonos_playlist, int(track),
> > +                                           int(new_pos), update_id)
> > +
> > +    @only_on_master
> > +    def remove_from_sonos_playlist(self, sonos_playlist, track, update_id=0):
> > +        """Remove a track from a Sonos Playlist.
> > +        This is a convenience method for :py:meth:`reorder_sonos_playlist`.
> > +
> > +        Example::
> > +
> > +            device.remove_from_sonos_playlist('SQ:0', track=0)
>
> maybe sonos_playlisthere as well
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r39388401>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

soco/core.py
dundeemt: done

On Mon, Sep 14, 2015 at 7:43 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> In soco/core.py
> <https://github.com/SoCo/SoCo/pull/352#discussion_r39388508>:
>
> > +                track is track 0, just like indexing into a Python list.
> > +            update_id (int): Optional update counter for the object. If left
> > +                at the default of 0, it will be looked up.
> > +
> > +        Returns:
> > +            dict: See :py:meth:`reorder_sonos_playlist`
> > +
> > +        Raises:
> > +            SoCoUPnPException: See :py:meth:`reorder_sonos_playlist`
> > +        """
> > +        return self.reorder_sonos_playlist(sonos_playlist, int(track), None,
> > +                                           update_id)
> > +
> > +    @only_on_master
> > +    def get_sonos_playlist_by_attr(self, attr_name, match):
> > +        """ Return the first Sonos Playlist DidlPlaylistContainer that
>
> No space before "Return"
>
> —
> <https://github.com/SoCo/SoCo/pull/352/files#r39388508>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

dundeemt:
I'll merge and update.

dundeemt:
Once everyone is +1 with it, I can rebase and squash if you desire.

lawrenceakka:
@dundeemt please rebase, squash as appropriate, and your self to AUTHORS so I can commit

dundeemt:
working on the rebase now

On Sun, Sep 20, 2015 at 6:10 PM, Lawrence Akka <notifications@github.com>
wrote:

> @dundeemt <https://github.com/dundeemt> please rebase, squash as
> appropriate, and your self to AUTHORS so I can commit
>
> —
> <https://github.com/SoCo/SoCo/pull/352#issuecomment-141843855>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

dundeemt:
working on squashing

lawrenceakka:
Thanks Jeff.  All done.  As a final task, please can you add a note to #332

dundeemt:
@KennethNielsen @lawrenceakka  Thank you for your help in getting this PR accepted and merged.  I appreciated your feedback, insights and efforts not only on the PR but the project in general.

KennethNielsen:
@dundeemt thank you for sticking with it. Sorry about not getting back to you sooner about some of the feedback, but I was pulled away by real life. The way I remember it, there was none of my comments in the last round that were really important, so I was happy to see that it had been pulled. Thanks for the effort. Just in the interest of discussion about style (which I must admit I lov), I'll comment on a single one of your comments from earlier.

KennethNielsen:
> Commas make lists unambiguous. While not required for a single integer, ['this',] != ['this']. Any iterable will be split into elements within the list if you don't mark them with commas. If you would like me to remove them, I will. Personally, I don't like to vary based on data types, so I choose to be explicit. If you have a pep I'd be happy to read and perhaps alter my opinion, but I've not read any thing authoritative against it. "Explicit is better than implicit." - https://www.python.org/dev/peps/pep-0020/

No. List already are unambiguous. If you try and type in your example, you will see that ``['this'] == ['this',]`` returns True. Now of course if you use the type function on a something that is iterable ``list('this')`` something else entirely happens. The way that I understand it, the trailing comma is without any semantic meaning whatsoever (except for tuples), but are **allowed**, and the main reason for that, as I understand it, is to make diffs look cleaner if you have multiline definitions. If you have a counter examples here, I'd be thrilled to see it. As far as tuples go, it is purely out of necessity, that a single element tuple has to be defined with a trailing comma, because the parenthesis is also used to function calls. So while, as you say, maybe it would be better to not have to vary depending on type, I really prefer to look at as if it is consistent without trailing comma for almost all container types (list, set, dict) and the tuple is just an ugly exception that we have to live with, but which is also easy to remember because we all know we use parenthesis for functions.

As far as ``"Explicit is better than implicit."``, I'm very aware of the Zen of Python and I think it, and the general focus on readability, is one of the absolutely brilliant things about the language. I'm just not sure I think it applies here, or that it overrules ``"Although practicality beats purity."``and most especially ``"Readability counts."`` ;)

PR #354, opened 2015-09-07
Better list slicing, removing deprecated code

lawrenceakka:
1) deprecated code in data_structures removed, and attribute fetching is streamlined.

2) ListOfMusicInfoItems is no longer a subclass of `list`, but a subclass of `collections.Sequence`. This makes it read only (which is what we want, for the moment), but also allows us do fancy indexing, which will be useful for playlists.

``` python
>>> alist = ListOfMusicInfoItems(
...     [1, 2, 3, 4, 5, 6, 7],
...     number_returned=7,
...     total_matches=7,
...     update_id=0
... )
>>> alist
ListOfMusicInfoItems(items=[1, 2, 3, 4, 5, 6, 7])

# Standard slicing and negative indices

>>> alist[2]
3
>>> alist[2:5]
[3, 4, 5]
>>> alist[-2]
6

# Multiple slices

>>> alist[1:3, 5:6]
[2, 3, 6]

#Repetition of indices

>>> alist[3, 3, 3]
[4, 4, 4]

#Steps are supported

>>> alist[3:7:2]
[4, 6]
```

3) tests - these lists were previously without them

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1867,7 +1867,7 @@ class SoCo(_SocoSingletonBase):
             full_album_art_uri=full_album_art_uri,
             subcategories=subcategories, search_term=track,
             complete_result=True)
-        result._metadata['search_type'] = 'search_track'
+        result.search_type = 'search_track'
         return result
 
     def get_albums_for_artist(self, artist, full_album_art_uri=False):
@@ -1892,12 +1892,10 @@ class SoCo(_SocoSingletonBase):
         # It is necessary to update the list of items in two places, due to
         # a bug in SearchResult
         result[:] = reduced
-        result._metadata.update({
-            'item_list': reduced,
-            'search_type': 'albums_for_artist',
-            'number_returned': len(reduced),
-            'total_matches': len(reduced)
-        })
+        result.item_list = reduced
+        result.search_type = 'albums_for_artist'
+        result.number_returned = len(reduced)
+        result.total_matches = len(reduced)
         return result
 
     def get_tracks_for_album(self, artist, album, full_album_art_uri=False):
@@ -1919,7 +1917,7 @@ class SoCo(_SocoSingletonBase):
             full_album_art_uri=full_album_art_uri,
             subcategories=subcategories,
             complete_result=True)
-        result._metadata['search_type'] = 'tracks_for_album'
+        result.search_type = 'tracks_for_album'
         return result
 
     @property
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -19,15 +19,13 @@ items such as tracks, playlists, composers, albums etc.
 
 from __future__ import unicode_literals
 
+import collections
+import operator
 import sys
-import warnings
-warnings.simplefilter('always', DeprecationWarning)
-import textwrap
-
-from .xml import XML, ns_tag
 
 from .exceptions import DIDLMetadataError
 from .utils import really_unicode
+from .xml import XML, ns_tag
 
 
 ###############################################################################
@@ -394,6 +392,7 @@ class DidlObject(DidlMetaClass(str('DidlMetaClass'), (object,), {})):
             # way.
             setattr(self, key, value)
 
+    # pylint: disable=too-many-locals
     @classmethod
     def from_element(cls, element):
         """Create an instance of this class from an ElementTree xml Element.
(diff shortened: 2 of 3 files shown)


dundeemt:
Can you give a hypothetical code snippet of using these features?  I would think that sorting would be a more common use case than slicing with respect to a playlist.  For example and completely hypothetical a sonos playlist as an ORM style implementation:
```python
my_spl = SonosPlaylist(obj.get_sonos_playlist_by_attr('title', 'Foo'))
my_spl.sort(['+album', '+original_track_number'])
my_spl.save()
#
# or even more wild
my_spl = SonosPlaylist.objects.get(title='Foo')
my_spl = my_spl.filter(creator='Del Fuegos')
my_spl = my_spl.order_by('album', 'original_track_number')
my_spl.save()
```
There are a number of ORM projects that could be used as a basis to create these classes. The allow you to create your own and attach to non traditional RDBM backends.

The sonos queue is like a Sonos Playlist, except there is only 1 queue (I could be wrong, the form of the item_id for a queue would suggest multiple queues had been considered. Q:0 compared to a Sonos Playlist SQ:0).   

Please don't misunderstand.  I am just trying to understand your idea.

KennethNielsen:
@lawrenceakka I think this looks fine, could go in as is +1

> 3) tests - these lists were previously without them

I'm afraid that was me. That was one of those "I will do it later" that I'm always telling others no to do :$

@dundeemt I can certainly see you point. I think in one of the other PRs, @lawrenceakka was talking about using indexing for re-organisation. So essentially, (If I understood correct), you would implement inplace operations to change to the lists. So e.g:
```python
playlist = zone.get_sonos_playlist()[0]
playlist[0:3] = playlist[2::-1]
# would reverse the order of the first 3 elements

# or
playlist[0, 2, 4] = [None, None, None]
# would delete the first 3 even elements
```
(and send the changes back). I can however see your point and I think the kind of interface you are talking about would be useful for sonos playlists and for the queue (and maybe even be a place, where would be able to do something cool I don't think the original controller does). Ideally however, I think would could have both. No-matter how much of this manipulate-by-methods stuff you attach, we could still do the fancy indexing, so I see no reason not to merge this as is (also since the work has been done) and then we can sort it out later if a PR turns up with the kind of functionality you are talking about.

(The only thing that might block this, is that for the kind of stuff you are talking about, you would probably need to be able to change the objects, but I'm sure that can be fixed as well.)

dundeemt:
I am in no way against the merge or idea.  Only pointing out my
thoughts/ideas.   Most of my attention is focused on getting sonos playlist
management ready for 0.12

On Sat, Sep 12, 2015 at 3:49 AM, Kenneth Nielsen <notifications@github.com>
wrote:

> @lawrenceakka <https://github.com/lawrenceakka> I think this looks fine,
> could go in as is +1
>
> 3) tests - these lists were previously without them
>
> I'm afraid that was me. That was one of those "I will do it later" that
> I'm always telling others no to do :$
>
> @dundeemt <https://github.com/dundeemt> I can certainly see you point. I
> think in one of the other PRs, @lawrenceakka
> <https://github.com/lawrenceakka> was talking about using indexing for
> re-organisation. So essentially, (If I understood correct), you would
> implement inplace operations to change to the lists. So e.g:
>
> playlist = zone.get_sonos_playlist()[0]
> playlist[0:3] = playlist[2::-1]# would reverse the order of the first 3 elements
> # or
> playlist[0, 2, 4] = [None, None, None]# would delete the first 3 even elements
>
> (and send the changes back). I can however see your point and I think the
> kind of interface you are talking about would be useful for sonos playlists
> and for the queue (and maybe even be a place, where would be able to do
> something cool I don't think the original controller does). Ideally
> however, I think would could have both. No-matter how much of this
> manipulate-by-methods stuff you attach, we could still do the fancy
> indexing, so I see no reason not to merge this as is (also since the work
> has been done) and then we can sort it out later if a PR turns up with the
> kind of functionality you are talking about.
>
> (The only thing that might block this, is that for the kind of stuff you
> are talking about, you would probably need to be able to change the
> objects, but I'm sure that can be fixed as well.)
>
> —
> <https://github.com/SoCo/SoCo/pull/354#issuecomment-139740951>.
>



-- 
Best,

Jeff Hinrichs
dundeemt@gmail.com
402.218.1473

KennethNielsen:
@lawrenceakka this seemed largely ready to merge, except now there are conflicts. Can you maybe rebase and merge at some point?

ghcs27:
It's a shame this hasn't been merged yet, since it would be a great improvement for playlist management.
Maybe I should try to rebase this myself?

KennethNielsen:
@ghcs27 agreed, this PR is a good idea. Feel free to give a rebase a try.

stefankoegl:
Closing this in favour of #504.

PR #358, opened 2015-09-12
A few more useful methods in MusicService

lawrenceakka:
Helpful methods for constructing playable items from music service track ids.  See:

https://gist.github.com/lawrenceakka/2d21dca590b4fa7e3af2

### DIFF
--- a/soco/music_services/music_service.py
+++ b/soco/music_services/music_service.py
@@ -15,7 +15,7 @@ import requests
 from xmltodict import parse
 
 from .. import discovery
-from ..compat import parse_qs, urlparse
+from ..compat import parse_qs, quote_url, urlparse
 from ..exceptions import MusicServiceException
 from ..music_services.accounts import Account
 from ..soap import SoapFault, SoapMessage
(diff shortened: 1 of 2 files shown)


soco/music_services/music_service.py
```diff
@@ -551,8 +551,51 @@ def available_search_categories(self):
...
+        # Add the account info to the end as query params
+        account = self.account
+        result = "soco://{0}?sid={1}&sn={2}".format(
+            item_id, account.service_type, account.serial_number
+        )
+        return result
+
+    @property
+    def desc(self):
+        """str: The Sonos descriptor to use for this service.
```
KennethNielsen: Hmm, what does that ``str:`` do?

lawrenceakka: > Hmm, what does that str: do?

It appears to be the way to document a property, according to Napoleon.  See the example property here: https://sphinxcontribnapoleon.readthedocs.org/en/latest/example_google.html#example-google

KennethNielsen:
There's something in the tests, but I suspect you've git your eye on it.

This will turn out to be very useful, a single inline comment and then +1.

On a side note. I was wondering what your thoughts are now about data structures. My fingers are really itching to try and see do something useful there. If you remember I did the original ones for the didl stuff and then you showed that it could all be done sooo much simpler, because I hadn't really understood the underlying spec. So I would like to give it a try sometime, to see if I can do something better now. But if you prefer to do it yourself, or you would still prefer that we wait and play a bit more with it, then I won't start it and will probably try to look at the docs in stead.

lawrenceakka:
> There's something in the tests, but I suspect you've git your eye on it.

I just fixed it!  Problems with unicode on PY2.6.

> But if you prefer to do it yourself, or you would still prefer that we wait and play a bit more with it, then I won't start it and will probably try to look at the docs in stead.

You're more than welcome to have a go, but I think it might need exercising a bit more first so that we have a clearer idea what sort of API we want. Perhaps play with https://gist.github.com/lawrenceakka/2d21dca590b4fa7e3af2 and https://gist.github.com/lawrenceakka/c8597d2533b6cc3e0d18 (which is a bit out of date).  Maybe we should start a separate issue, and exchange a few ideas about what the API should end up looking like, before figuring out how to implement it!

soco/music_services/music_service.py
KennethNielsen: Do'h! (face palm) it is of course the return type. Oh my, more coffee.

lawrenceakka:
:coffee: :coffee:

KennethNielsen:
> Maybe we should start a separate issue, and exchange a few ideas about what the API should end up looking like, before figuring out how to implement it!

I'm all for that. It is just sometimes for public FOSS projects, not everyone is happy to discuss things in the abstract, but prefer an example implementation. But I think it could often save us some time. If you don't mind, I will have a look at your gists, read some more of the documentation you linked to in the original ms PR and start the discussion PR.

soco/music_services/music_service.py
```diff
@@ -551,8 +551,55 @@ def available_search_categories(self):
...
+        # x-sonos-http:tr%3a92352286.mp3?sid=2&flags=8224&sn=4 The
+        # extension (.mp3) presumably comes from the mime-type returned in a
+        # MusicService.get_metadata() result (though for Spotify the mime-type
+        # is audio/x-spotify, and there is no extension. See
+        # http://musicpartners.sonos.com/node/464 for supported mime-types and
+        # related extensions). The scheme ( x-sonos-http) presumably
+        # indicates how the player is to obtain the stream for playing. It
+        # is not clear what the flags param is used for (perhaps bitrate,
+        # or certain metadata such as canSkip?). Fortunately, none of these
+        # seems to be necessary. We can leave them out, (or in teh case of
```
fliiiix: Just a small thing: **teh**
Oh and on line 573 is a space to much `The scheme ( x-sonos-http)`

:+1: for the clear and understandable documentation

lawrenceakka: Thanks. Fixed it.

PR #371, opened 2015-10-02
Handle dependencies properly, in setup.py

lawrenceakka:
We seem to have got into a bit of a tangle with various setup methods, duplication in requirements, and installation of unnecessary dependencies for certain versions of python.  In general, _all_ dependencies should be handled in just one place: `setup.py`.  `requirements` and `requirements-dev.txt` are unnecessary for us (see https://caremad.io/2013/07/setup-vs-requirement/) and inflexible (it is not straightforward to handle conditional dependencies there, for example).

This PR tidies up setup.py a bit, removes code which I think could never have worked (!) and hopefully makes things clearer.  The two requirements files are removed.  Installation with pip still works, so:

``` sh
$ pip install soco
```

will install SoCo and its dependencies (no need for any additional manual installation of requests etc).

For the dev version, use

``` sh
$ pip install -e git+https://github.com/SoCo/SoCo.git#egg=soco[dev]
```

or, if you already have a copy of the source files, change into the
directory containing `setup.py` and run::

``` sh
$ pip install -e .[dev]
```

This will install SoCo in editable mode, together with all development dependencies.


stefankoegl:
I am not totally sure about this... I generally like it to have a list of dependencies in a separate file, separate from code.

What do the others think about this change?

dsully:
I like this change, as it allows per-version dependencies, unlike requirements.txt

setup.py is configuration, not code in my opinion.

KennethNielsen:
I also like that we can have per version dependencies. My only gripe with this, is that I'll have to un-teach my fingers to install dependencies from requirements.txt when installing from source, but I do think it is the better way.

stefankoegl:
This has been sitting quite a while... I noticed that others agree with the change, so I've brought it up to date and prepared an updated PR at #609.

PR #383, opened 2015-12-19
Added support for getting Sonos Favorites.  Fixed some bugs with getting Radio Shows

tlc:
Added `get_sonos_favorites(self, start=0, max_items=100)`.
Fixed `get_favorite_radio_shows(self, start=0, max_items=100)`.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1358,8 +1358,9 @@ class SoCo(_SocoSingletonBase):
             favorite_type = SONOS_FAVORITES
 
         response = self.contentDirectory.Browse([
-            ('ObjectID', 'FV:2' if favorite_type is SONOS_FAVORITES
-                         else 'R:0/{0}'.format(favorite_type)),
+            ('ObjectID',
+             'FV:2' if favorite_type is SONOS_FAVORITES
+             else 'R:0/{0}'.format(favorite_type)),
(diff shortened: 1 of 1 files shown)


tlc:
The remaining travis complaints don't appear to be associated with code I've touched.

robwebset:
Seems like a nice addition - any reason this can't be merged?

KennethNielsen:
Hallo @tlc.

Sorry for the long delay in review. This looks like a really nice addition.

Can you comment on what the bug fixes were to the original methods?

Also, I would prefer if the results came out as SoCo data structures. I will think on it a little and try to elaborate shortly.

\Kenneth

tlc:
At this point I don't really remember.  But glancing over the diff, this looks problematic:

    if favorite_type != RADIO_SHOWS or RADIO_STATIONS:

KennethNielsen:
@tlc regarding the comments about changing the return types to use SoCo data structures I thought that since this had been sitting around here for so long I might as well merge it. So I added warning that it will probably change in the future and then if you want to work on that, we can do it in a separate PR.

Many thanks for the contributions.

I apologize for the long wait to have your changes reviewed, but it seems that most of the maintainers of this project find time to work on it in little sprints.

Can I get you to add a comment to the #332 about the features you added?

Issue #386, opened 2016-03-06
get_current_track_info: ParseError: not well-formed (invalid token)


ArjanM83:
When requesting current track info, this results in an error when an invalid token is returned by the sonos zone:

```
  File "C:\Python27\lib\site-packages\soco\core.py", line 1019, in get_current_t
rack_info
    ('Channel', 'Master')
  File "C:\Python27\lib\site-packages\soco\services.py", line 156, in _dispatche
r
    return self.send_command(action, *args, **kwargs)
  File "C:\Python27\lib\site-packages\soco\services.py", line 351, in send_comma
nd
    result = self.unwrap_arguments(response.text) or True
  File "C:\Python27\lib\site-packages\soco\services.py", line 244, in unwrap_arg
uments
    tree = XML.fromstring(xml_response)
  File "<string>", line 124, in XML
ParseError: not well-formed (invalid token): line 1, column 850
```

This is the xml_response, I believe the invalid content is within the stream content:

```
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle=
"http://schemas.xmlsoap.org/soap/encoding/"><s:Body><u:GetTransportInfoResponse
xmlns:u="urn:schemas-upnp-org:service:AVTransport:1"><CurrentTransportState>PLAY
ING</CurrentTransportState><CurrentTransportStatus>OK</CurrentTransportStatus><C
urrentSpeed>1</CurrentSpeed></u:GetTransportInfoResponse></s:Body></s:Envelope>
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle=
"http://schemas.xmlsoap.org/soap/encoding/"><s:Body><u:GetPositionInfoResponse x
mlns:u="urn:schemas-upnp-org:service:AVTransport:1"><Track>1</Track><TrackDurati
on>0:00:00</TrackDuration><TrackMetaData>&lt;DIDL-Lite xmlns:dc=&quot;http://pur
l.org/dc/elements/1.1/&quot; xmlns:upnp=&quot;urn:schemas-upnp-org:metadata-1-0/
upnp/&quot; xmlns:r=&quot;urn:schemas-rinconnetworks-com:metadata-1-0/&quot; xml
ns=&quot;urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/&quot;&gt;&lt;item id=&quot
[...]

PR #392, opened 2016-04-19
Filter invalid chars in xml fixes

KennethNielsen:
Apparently, certain radio stations sometimes sends invalid XML in the form of XML with disallowed characters in it (#386). This PR attempts for fix that. If the XML parsing in services.unwrap_arguments fail, we now try to filter out disallowed chars and retry the parsing.

I consider this a bug fix, so I will merge within a couple of days if there are no objections.

### DIFF
--- a/soco/services.py
+++ b/soco/services.py
@@ -45,7 +45,7 @@ from .exceptions import (
     SoCoUPnPException, UnknownSoCoException
 )
 from .utils import prettify
-from .xml import XML, illegal_xml_re
+from .xml import XML, illegal_xml_re, PARSEERROR
 
 # UNICODE NOTE
 # UPnP requires all XML to be transmitted/received with utf-8 encoding. All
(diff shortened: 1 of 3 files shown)


KennethNielsen:
Grrr. After fighting Python 2.6 for 45 minutes I am still left with the issue on Travis CI above. Does anyone have any ideas?

KennethNielsen:
Finally. This is just amazing. I used just as much time fighting Python 2.6 on this one as I did on the actual fix. Not least because the only place I have access to Python 2.6 is on Travis CI. As it turns out, Python 2.6 raised SyntaxError !!! on invalid chars. I must admit I have never seen that used in that way.

Oh well, I finally got it working and squished all the extra commits so it should now be ready for merge.

KennethNielsen:
Since there seems to be no objections and this is a bug fix I will merge it.

PR #393, opened 2016-04-23
Fix any_soco call in get_alarms

simonalpha:
Required import of soco module not included, and would be masked regardless due to soco variable in function scope. Submodule import included instead.

### DIFF
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -9,6 +9,7 @@ import re
 import weakref
 from datetime import datetime
 
+from . import discovery
 from .core import PLAY_MODES
 from .xml import XML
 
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@simonalpha Looks good (reviewed and tested). Since it's a bug fix I can merge immediately, but I was wondering if I could persuade you to fix that use of the name soco in the get_alarms function? Not that it's really a problem anymore, but it is inconsistent with how we usually name soco instances. Maybe rename to `zone` the way it is in Alarm.

simonalpha:
Done! Thought it was a bit strange.

lawrenceakka:
Great - thanks

PR #407, opened 2016-06-12
Updated links to point to SoCo/Soco repository

alasdairnicol:
The remaining links to rahims/Soco appear to be ok. One is to the
original repository in the README, the rest are to historic issues.


stefankoegl:
Good catch, thanks!

PR #412, opened 2016-07-17
Data structures for music services (request for comments)

KennethNielsen:
This PR is **NOT** for pulling, only for discussion.

Ok, so I finally have some work to show on the data structures for music
services. In-stead of laying a ton of work in polishing it in just the way I
think it should be, I thought it would be better to get a round of feedback
early, which is what this this PR is for. Thus there are no tests or docs
yet and only minimal doc strings.

Quick summary:

My initial purpose was:
1. To have data structures similar to the ones we have for music library items
   i.e. that can be browsed and added as is to the queue.
2. To expose the metadata
3. (Optionally) To have the metadata in the correct types

Initially I was planning on doing only the minimal to get them queueable
(1. and 2. and just one class with the type as a attribute), but it turned out
to be only very little extra work to get a fully spec[1] compliant
implementation with classes for all the types.

The work is based on the spec[1] and two gists from @lawrenceakka [2][3].

Note: The types here seem to be completely independent from the DidlItem
hierarchy which is why I choose a new implementation. Trying to shoehorn it
into the Didl hierarchy and types is in my view not a good idea.

The status now is:
- I have an implementation of the class hierarchy and metadata parsing and
  type conversion, that I am very satisfied with. It does what it is supposed
  to and is straight forward and easy to read (3 levels on inheritance and no
  fancy object oriented tricks).
- The changes for integration with the rest of the code base is minimal (23
  lines of changed code in music_service.py and soco/data_structure.py, though
  that is of course not a goal in itself, see below:

Where I really could use some feedback are on the following points:

<style type="text/css">ol { list-style-type: upper-alpha; }</style>

 A. What do you think about the overall design?

 B. The to_element implementation is actually instantiating an DidlItem and
    callings its to_element. A way to make this a little cleaner, would be to
    factor the to_element mechanics out into a pure function in
    soco/data_structures and have both classes methods calling that. What do
    you think?

 C. Naming: I have placed the code in soco/music_services/data_structures.py
    whereas the music llibrary data structures is in
    soco/data_structures.py. Code and functionallity wise I don't think this
    should be a problem, but it may be confusing. Thoughts?

 D. Upgrading music service elements in the queue: The problem is this. When
    we ask for the queue, should we upgrade those elements to music services
    elements. I tend to think so. A few questions then pop up. The item_id is
    special when the elements come from a queue, but in order to be able to
    re-add the element to the queue, we need the original item_id, so when to
    do the conversion? Should we keep the item_id (to keep an indcation of
    that the item came from the queue) and only convert when asking to_element
    (in the assumption that it will now be added to the queue) or just convert
    immediately.

 E. Continued on the queue is when and how to check whether the element is a
    music service element. Currently I check by looking on the form of the uri
    and I do it by first creating an original DidlItem and then asking the ms
    data strctures to upgrade it if necessary. What do you think about that.

 E. ~~Circular imports. soco/data_structures.py:from_didl_string is the central
    entrypoint for having a reply turned into data structures. This means that
    this module needs to import stuff from music/services/data_structures.py,
    but that module needs to import DidlResource, DidlItem, SearchResult from
    soco/data_structures.py and thus we have a circular dependence. Similarly,
    music_services.py and the ms data_strctures also need to import from each
    other. For now, this has been solved by changing "from module import
    stuff" to just "import module" and then using "module.stuff" (a trick that
    I have not yet researched why works). The "from .." imports make it
    fail. Generally circular imports are frowned upon, so should I re-factor
    to avoid them? I don't mind doing the work (including probably a lot of
    thinking about anice logical structure). I just wanted to hear you
    opinions first (also because it might break users imports). A big part of
    the problem can be solved by moving from_didl_string out into its own
    module (say data_structures_common), which I actually have an
    implementation laying around for.~~

 F. There is a minor issue about the names in _valid_fields and _types
    now being CamelCase (as in the spec and therefore easy to compare)
    but the fields in the actual metadata are converted to lowercase
    underscore. I think I will convert the names in the class
    attributes, it should still be easy to compare with the spec.

Examples of usage:

```
zone = any_soco()
ym = MusicService('YouSee Musik')

# Search for an album
items = ym.search('albums', 'the last resort', 0, 1)
for item in items:
    print(item)
album = items[0]
zone.add_to_queue(album)

tracks = ym.get_metadata(album)
last_track = tracks[-1]
zone.add_to_queue(last_track)
# And enjoy some electronic music from a Danish DJ

# Visualize the content tree (note the music service is hardcoded)
def tree(browse_item='root', indent=0, max_depth=10):
    if indent > max_depth:
        return
    if isinstance(browse_item, track_class):
        return
    for item in ym.get_metadata(browse_item):
        print("*" * indent + ' ', item.title, sep='')
        tree(item, indent=indent+1, max_depth=max_depth)
tree(max_depth=2)
```

Other types like genre and artist can of course also be browsed, but not all can be added to the queue, but that is a restriction from the music service, not in the implementation. Give it a go and let me know what you think.

TODO (except from the things that spring out of the discussion of the points above):
- Test and play around with playing streams
- Think about having the music_service module have a cache of instances and have the queue items update themselves with the full metadata from the music service instances in the cache, if they are there.
- Others.

[1] http://musicpartners.sonos.com/node/83
[2] https://gist.github.com/lawrenceakka/2d21dca590b4fa7e3af2
[3] https://gist.github.com/lawrenceakka/c8597d2533b6cc3e0d18

EDIT: Question E has been taken care of by 82384a7, see comment below.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -21,8 +21,9 @@ from . import config
 from .compat import UnicodeType
 from .data_structures import (
     DidlObject, DidlPlaylistContainer, DidlResource,
-    Queue, from_didl_string, to_didl_string
+    Queue, to_didl_string
 )
+from .data_structures_entry import from_didl_string
 from .exceptions import SoCoSlaveException
 from .groups import ZoneGroup
 from .music_library import MusicLibrary
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -38,6 +38,7 @@ from .utils import really_unicode
 from .xml import (
     XML, ns_tag
 )
+from .music_services import data_structures as ms_data_structures
 
 
 ###############################################################################
@@ -96,7 +97,9 @@ def from_didl_string(string):
                 cls = _DIDL_CLASS_TO_CLASS[item_class]
             except KeyError:
                 raise DIDLMetadataError("Unknown UPnP class: %s" % item_class)
-            items.append(cls.from_element(elt))
+            item = cls.from_element(elt)
+            item = ms_data_structures.attempt_datastructure_upgrade(item)
+            items.append(item)
         else:
             # <desc> elements are allowed as an immediate child of <DIDL-Lite>
             # according to the spec, but I have not seen one there in Sonos, so
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -70,12 +70,6 @@ def to_didl_string(*args):
         return XML.tostring(didl, encoding='unicode')
 
 
-
-            # In case this class has an # specified unofficial
-            # subclass, ignore it by stripping it from item_class
-            if '.#' in item_class:
-                item_class = item_class[:item_class.find('.#')]
-
 ###############################################################################
 # DIDL RESOURCE                                                               #
 ###############################################################################
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -38,7 +38,6 @@ from .utils import really_unicode
 from .xml import (
     XML, ns_tag
 )
-from .music_services import data_structures as ms_data_structures
 
 
 ###############################################################################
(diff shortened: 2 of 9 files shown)


KennethNielsen:
Question E has been taken care of by 82384a7. It introduces soco/data_structures_entry.py that has the from_didl_string and attempt_datastructure_upgrade implementations. Thus this modules contains parsing functions that needs a leg into both music library and music services data structures.

KennethNielsen:
@lawrenceakka I would love your input on this.

It does not need to be detailed with solutions for all the little quirks, just high level, what do you think works, what doesn't and what needs more work.

KennethNielsen:
I finished the implementation and tests without feedback and will submit a proper PR shortly.

PR #413, opened 2016-07-17
Add sleep timer

americanwookie:
This pull request adds two new methods: set_sleep_timer, and get_sleep_timer. One can pass a free form string, like '02:00:00' to set_sleep_timer to set the sleep timer on a speaker, and then use get_sleep_timer to get information about the timer. The speaker must be a master.

I used #341 as a model for this, adding unit and integration tests. I look forward to feedback!


robwebset:
Great work to add sleep timer interface support.

I do have one comment - I think I'd prefer passing (and returning) something like "number of secconds" into the interface and have the interface itself sort out the "expected format" that Sonos expects inside the method (i.e. seconds to "00:00:00" format)

Thanks
Rob

americanwookie:
This has been added. I ensured ```get_sleep_timer()``` returns a ```dict``` that mirrors mostly the actual return from the device, while retaining consistency with the seconds.

Also, I chose to disable pylint's redefining error in this case to avoid long lines or single use variable names. An alternative here is to send H:MM:SS (when H < 10), that'd remove lines 1517-1519, but I don't know if we want to count on that.

I'm open to suggestions as to the consistency or the method I used to get the HH:MM:SS back. Thanks!

robwebset:
Hi americanwookie,

Thanks for taking the time. My personal feeling is that the get_sleep_timer() would be better just to return the number of seconds (to match the set method) rather than a structure. I think this keeps it much simpler and easier to follow/understand.

But it would be good to get others comments on this.

Thanks, Rob

robwebset:
Hi 

I just started trying this out, and I don't think the "get" interface is returning the details document in your get method, I'm getting the following returned:

{'CurrentSleepTimerGeneration': '8', 'RemainingSleepTimerDuration': 7182}

Which doesn't seem quite right?
Thanks
Rob

americanwookie:
I'm not quite sure I follow. The docstring in the `get_sleep_timer` diff promises a dict with the two elements, and that seems to be what you're getting back, right? Please bear with me, while I'm strong in Perl, I'm still getting my bearings in Python, so I may be using the wrong terminology.

robwebset:
Sorry, Wasn't very clear, the description stated:

       Returns:
            dict: Which contains 2 elements: RemainingSleepTimerDuration
                (`int`) as seconds left and CurrentSleepTimerGeneration (`str`)

But the CurrentSleepTimerGeneration doesn't seem to equate to anything? I had thought it would be the time remaining as a string based on the header description - but I think I miss-understood.

I'm not sure what the advantage of returning this value is really now I give it more thought.

Thanks
Rob
BTW: Got it reading and setting the sleep timer now - great work! - Thanks

americanwookie:
I agree, `get_sleep_timer` return seems overly complicated. `CurrentSleepTimerGeneration`  looks like some sort of serial number that increments every time the timer is modified. I couldn't understand it, but I kept it to retain fidelity with the underlying API, and I wasn't sure what the project values in situations like this.

I've added a commit to this pull request simplifying `get_sleep_timer`'s return to simply return the number of seconds as an int. If you'd prefer to keep fidelity, we can just pop that commit off.

robwebset:
@americanwookie yes I like the simplicity of it now, set it in seconds, get the time remaining in seconds.

@lawrenceakka @KennethNielsen are we OK to merge this one in - I've run some tests on my system and it works very well.

Thanks, Rob

KennethNielsen:
Hi @americanwookie 

I think this looks great and we should merge it soon. I do however have a few nit picks that I wanted to ask you to take care of before commit. I have left those as inline comments.

Regarding your discussion about what the better interface is '1:00:00' or 3600 I can see both ends of that argument. I can follow @robwebset arguments that seconds is somewhat cleaner, on ther other hand, if implemented in a UI, it may very well be, that the have a integer control for each one of hours, minutes and seconds, which would make the string the better choice. Let's stick with int for how and see what happens.

There is one thing that I will comment on here. We, some time ago, abandoned the idea of return code (i.e. returning True or False), though not all the methods have been updated to reflect the change. Point is, that the policy that we are aiming for now is, that if the method executes, it is a success, otherwise an exception will be raised. I think this also opens up for some small refactoring to make the code a little cleaner.

Regarding your statement earlier about disabling a pylint warning. Just as a general observation. For Python we put readability at the top of the list of priorities, so variable names should be descriptive and we very rarely compromise on that. Also, and to follow up on readability, single use variables are not frowned upon, if they help readability (and besides, method or function local variables are very fast to create).

Comments inline to follow.

soco/core.py
```diff
@@ -1490,6 +1493,59 @@ def get_item_album_art_uri(self, item):
...
 
+    @only_on_master
+    def set_sleep_timer(self, sleep_time_seconds):
+        """Sets the sleep timer.
+
+        Args:
+            sleep_time_seconds (`int`): How long to wait before turning off
+                speaker in seconds. Maximum value of 86399
+
+        Returns:
```
KennethNielsen: As mentioned in the conversation. We would rather move to not doing this. Failure should be indicated by an exception and it is a success otherwise.

soco/core.py
```diff
@@ -1490,6 +1493,59 @@ def get_item_album_art_uri(self, item):
...
+            bool: True if succesful, False otherwise
+
+        Raises SoCoException (or a subclass) upon errors, ValueError for
+            incorrect syntax.
```
KennethNielsen: I think the proper way to document this with Napoleon (which we are moving to use) and sphinx is one indented line per possible exceptions i.e:

Raises:
    SoCoException: ... reason for that
    ValueError: ... reason for that

KennethNielsen:
More inline comment to come, but I didn't have time to finish them now.

soco/core.py
```diff
@@ -1490,6 +1493,59 @@ def get_item_album_art_uri(self, item):
...
+
+        """
+        if sleep_time_seconds and \
```
KennethNielsen: I'm wondering whether the checks and the exception throwing could be refactored slightly, to become a little cleaner.

How about:
 1. Doing only the second check here, the "can accurately be interpreted as integer" check"
 2. Then convert to int and then to '1:00:00' string
 3. Then simply make the self.avTransport.ConfigureSleepTimer call inside an try except. Assuming you get a different exception with a bad value than with e.g a communications error, the you could check for the former and raise a ValueError and reraise the latter. Let me know if I'm not making any sense and I can exemplify it.

Note, that this is just a suggestion.

soco/core.py
```diff
@@ -1490,6 +1493,59 @@ def get_item_album_art_uri(self, item):
...
+           sleep_time_seconds != int(sleep_time_seconds):
+            raise ValueError('invalid sleep_time_seconds, must be integer \
+                value between 0 and 86399 inclusive')
+        if sleep_time_seconds:
```
KennethNielsen: One question arose while reading this. How does one disable the sleep timer? When reading this line I might be led to believe that a value of 0 will disable it, but it is not documented in the docstring. Also what does the get method return if there is no sleep timer?

soco/core.py
```diff
@@ -1490,6 +1493,59 @@ def get_item_album_art_uri(self, item):
...
+            # pylint: disable = redefined-variable-type
+            sleep_times = datetime.timedelta(seconds=int(sleep_time_seconds))
+            sleep_times = str(sleep_times).split(':')
```
KennethNielsen: I'm not quite sure. Maybe this was what you referred to in your own comment. It looks a little redundant. You mentioned something about the ours being on digit or two. My point is, that if the device doesn't care then neither should we. If it does, there may be a cleverer way to do the formatting. I think on that a little.

KennethNielsen:
That was it for the rest of the comments for the code. Most of it is discussion points or nit picks. Let me know what you think.

soco/core.py
americanwookie: This string manipulation code is ugly and seeing as I'm new to Python, I'm open to any examples you can offer.

Also, as I was preparing this comment, I realized my code doesn't distinguish between 0 and '' for ```sleep_time_seconds```. These values are different, '' will stop the current sleep timer, but '' will cause the speaker to immediately start tapering the sound, a potentially useful feature. I wonder if you can work this into such an example, or if we should have a separate method just for this feature.

soco/core.py
americanwookie: Sleep timer is disabled by calling ```set_sleep_timer``` with ```sleep_time_seconds``` set to 0. I've updated the doc in ```set_sleep_timer``` to reflect this and  ```get_sleep_timer``` to reflect that 0 will be returned when no sleep timer is set.

soco/core.py
americanwookie: Wait! As I read your comment two steps down, I realize, now I know how you're thinking this can be restructured . . . lemme know what you think.

soco/core.py
americanwookie: No, in my limited testing I found the device doesn't care. Maybe down the road it does. But there's so much string gymnastics here that I'm inclined to just clean it up and count on the device never caring.

soco/core.py
americanwookie: Wait, I don't like the above at all. 0 and '' should be consistent. I'm updating the code to return None when no sleep timer is set.

americanwookie:
I'm nearly finished revising based on this feedback. I'll be pushing a new commit in a few hours, after dinner.

soco/core.py
```diff
@@ -1490,6 +1494,70 @@ def get_item_album_art_uri(self, item):
...
+        # useful feature, while None means cancel the current timer
+        if sleep_time_seconds == 0:
+            sleep_time = '00:00:00'
+        elif not sleep_time_seconds:
+            sleep_time = ''
+        elif sleep_time_seconds != int(sleep_time_seconds):
+            raise ValueError('invalid sleep_time_seconds, must be integer \
+                value between 0 and 86399 inclusive')
+        else:
+            sleep_time = str(
```
americanwookie: Okay, so I'm failing because I'm redefining the type of ```sleep_time``` from ```unicode``` to ```str```. This makes no sense to me, as ```sleep_time``` was previously undefined. Help please.

americanwookie:
I've placed a comment in the "Files Changed" section regarding some help I need in passing the tests. Your feedback will be appreciated!

KennethNielsen:
Really nice work. I thinks it looks real good with the refactoring you did. I will leave a few in-line comments.

soco/core.py
```diff
@@ -1490,6 +1494,70 @@ def get_item_album_art_uri(self, item):
         else:
             return None
 
+    @only_on_master
+    def set_sleep_timer(self, sleep_time_seconds):
+        """Sets the sleep timer.
+
+        Args:
+            sleep_time_seconds (`int/NoneType`): How long to wait before
```
KennethNielsen: You don't need to mark the types as code with ``. Sphinx and Napoleon will take care of that.

soco/core.py
```diff
@@ -1490,6 +1494,70 @@ def get_item_album_art_uri(self, item):
...
+
+        Raises:
+            SoCoException: Upon errors interacting with Sonos controller
+            ValueError: Argument/Syntax errors
+
+        """
+        # Note: A value of None for sleep_time_seconds is valid, and needs to
+        # be preserved distinctly separate from 0. 0 means go to sleep now,
+        # which will immediately start the sound tappering, and could be a
+        # useful feature, while None means cancel the current timer
```
KennethNielsen: Good with a comment to explain the difference between of and 0. I do however this we do the check just one bit better. In general (as you noticed) it can give some problems to check whether something is None, simply by checking if it evaluates to False, if there is another valid input that also evaluates to False (like 0). However, a slightly cleaner way to do is to simply check whether variable is None, that way the checks can be rewritten to:

```python
if sleep_time_seconds is None:
    sleep_time = ''
elif sleep_time_seconds != int(sleep_time_seconds):
    ....
else:
    ....
```

soco/core.py
KennethNielsen: I'm really not sure. One issue however is that I don't **think** that the str representation is guaranteed to remain the same. I think a more reliable way to do it, is to rely on the timedelta objets ability to format itself:
```python
sleep_time = format(datetime.timedelta(seconds=int(sleep_time_seconds)))
```

Maybe see if that gets rid of the error. In any case I think it is better, since we expect unicode everywhere internally in SoCo and this will in Python 2.7 convert convert to Python 2.7 str which is not unicode.

soco/core.py
```diff
@@ -1490,6 +1494,70 @@ def get_item_album_art_uri(self, item):
...
+                datetime.timedelta(seconds=int(sleep_time_seconds))
+            )
+
+        try:
+            self.avTransport.ConfigureSleepTimer([
+                ('InstanceID', 0),
+                ('NewSleepTimerDuration', sleep_time),
+            ])
+        except SoCoUPnPException as err:
+            if re.match(r'Error 402 received', str(err)):
```
KennethNielsen: Maybe a regex match is a little heavy handed. If you can find the field on the exception that has the error message, you could do a simple `in` test, i.e: `substring in error_message`

soco/core.py
```diff
@@ -1490,6 +1494,70 @@ def get_item_album_art_uri(self, item):
...
+            else:
+                raise
+
+    @only_on_master
+    def get_sleep_timer(self):
+        """Retrieves remaining sleep time, if any
+
+        Returns:
+            int or NoneType: Number of seconds left in timer. If there is no
+            sleep timer currently set it will return None.
```
KennethNielsen: there should be an extra 4 space indent of the second line of the description

americanwookie:
Thanks for the feedback, it was very helpful, I'll learn Python yet! All requested changes are in, with a rework on the exception handling to provide a more consistent exception message, and improved tests.

KennethNielsen:
@americanwookie great work. I'll merge it now. Thanks for your contribution.

Can you add a comment in #332 (for the release notes) that describes your changes and includes a link to this PR?

PR #421, opened 2016-08-07
Add playbar night mode support

christiandt:
This PR adds night mode for Sonos players that support this feature. Night mode will make the dialogue more clear, while reducing the intensity of loud sounds.

https://sonos.custhelp.com/app/answers/detail/a_id/1959/~/improving-speech-clarity-and-night-time-listening-through-the-playbar

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,6 +147,7 @@ SoCo supports the following controls amongst others:
 -  Get or set the speaker’s bass EQ
 -  Get or set the speaker’s treble EQ
 -  Toggle the speaker’s loudness compensation
+-  Toggle the speaker's night mode
 -  Turn on (or off) the white status light on the unit
 -  Switch the speaker’s source to line-in or TV input (if the Zone Player
    supports it)
--- a/soco/core.py
+++ b/soco/core.py
@@ -20,7 +20,9 @@ from .data_structures import (
     DidlObject, DidlPlaylistContainer, DidlResource,
     Queue, from_didl_string, to_didl_string
 )
-from .exceptions import SoCoSlaveException
+from .exceptions import (
+    SoCoSlaveException, NotSupportedException
+)
 from .groups import ZoneGroup
 from .music_library import MusicLibrary
 from .services import (
@@ -162,6 +164,7 @@ class SoCo(_SocoSingletonBase):
         bass
         treble
         loudness
+        night_mode
         cross_fade
         status_light
         player_name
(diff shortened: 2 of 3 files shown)


DPH:
Hi @christiandt  - good addition. 
Works well on a sound bar, but errors if used on another player. 
It's back to the issue of how much do we protect users from creating their own errors!

You could:
-  add a test "is_playing_tv" as it it only meaningful if playing TV and the sound bar is the only one that can
-  test if it's a sound bar before making the call. 
-  trap the error and return

Cheers David

christiandt:
Hi. Thanks for the feedback :)
I would argue that setting night mode without playing TV is a valuable feature. If one wants to automate night mode (e.g. every week-day after 23), it would be irritating to have to change the source to TV, enable night mode, then set the source back again.

I instead created a is_playbar method that will check the model_name of the speaker. This will not send NightMode packets to the Sonos if it is not identified as a Playbar. I find this to be a better solution compared to sending incompatible packets to Sonos devices and responding to the Invalid Args exception afterwards, as we decrease unnecessary traffic on the network.

DPH:
Good solution - works well.
Cheers David

christiandt:
Is there anything missing from my side for this change (and #422 ) to get merged? This change will not modify existing functionality, but add missing functionality of the library.

KennethNielsen:
Hallo @christiandt 

I think this looks good. I do however have a suggestion for the interface.

I would prefer it if we could expose the information about whether the feature is supported by the speaker or or not, directly in the night_mode property instead of **having** to go through is_playbar. So I would suggest:

 * From the getter return None if the mode is not supported
 * In the setter raise an exception if the mode is not supported, but make it possible to disable the error if needed. Something along the lines of:

```python
@night_mode.setter
def night_mode(self, night_mode, disable_not_supported_error=False):
    """Switch on/off the speaker's night mode."""
    ...
```

Regards Kenneth

darkskiez:
I think changing the getter to None is a good idea, and a one word change that conveys the real state.

I think raising an exception of the setter if used inappropriately is fine, but disabling the exception handling through a (double negative!) parameter is strange  ... exceptions are disabled by catching them in the caller :) 

I'd really like to use this feature in my home-assistant automation to enable night mode automatically, as where I live this is very important.

darkskiez:
I traced the other settings that are missing here..

EqTypes:
SurroundEnable 0/1
SurroundMode 0/1 [Ambient / Full]
SurroundLevel -15 to 15 [Tv Surround Level]
MusicSurroundLevel  -15 to 15
SubEnable 0/1 
SubPolarity 0/1 [Placement Adjust off/on]
SubGain -15 to 15
AudioDelay 0 to 5 [Tv Dialog Lip Sync]

KennethNielsen:
@darkskiez agreed. I think it was late with not enough coffee, since there is no way to pass an optional argument to a property anyway. *facepalm*.

@christiandt So, my lack of caffeine induced errors aside. My suggestions would be:

 * From the getter return None if the mode is not supported
 * In the setter raise a custom NotSupported exception (which you will have to add to exceptions) if the mode is not supported

I would prefer not to add the is_playbar property at this point. I already regret that we have three properties for `is_playing_tv`, `is_playing_radio`, `is_playing_line_in` instead of one is_playing that returns strings. I think the same applies here. The information is already exposed via `get_speaker_info`, however if it is really important we could think about translating them and serving them as nice clean strings through a `player_type` or similar, but I think that should be a separate PR.

So if you can move the check semantics into the properties (get and setter) I think we are good to go.

christiandt:
@KennethNielsen All requested changes should now be in place.
Travis is failing due to a PEP-8 issue introduced in #444 , in other words not related to this PR.

KennethNielsen:
Looks good.

Just a final note, nitpicking really. I think the exception will be more generally useful if it is simply "NotSupportedException" and also could I persuade you to squash the commits, so that we will have a cleaner history. If you are not familiar with how to do that with git feature I can also just do it for you when I merge.

christiandt:
No problem. I have now changed the exception to NotSupportedException and squashed commits :)

KennethNielsen:
@christiandt merged. Thanks for your contribution.

Can you add a note about the change to the release notes issue for the next release (#440)

PR #423, opened 2016-08-10
Hard deprecation of spotify plugin. Fixes #401.

### LINKED ISSUE #401: Spotify Metadata API now end-of-life (0.11.1)
The Spotify Metadata API has gone end-of-life, breaking the lookups in the current plugins.spotify module. 

Here's the API URL currently used that's been deprecated by Spotify (line 190, plugins.spotify.py):

`api_lookup_url = 'http://ws.spotify.com/lookup/1/.json'`

...and here's a webpage from Spotify on migrating to the new API... https://developer.spotify.com/web-api/migration-guide/

I'm going to have a go at fixing it, but others will likely be able to do it a whole load faster that me!

Best,
Matt

KennethNielsen:
This PR is for the hard deprecation of the Spotify plugin. By hard deprecation I mean replacing all the code in the module with a RuntimeError describing the situation. The API docs will also show the text from the error.

The reason for this move is that the metadata API from Spotify, that this was originally written for, has reached end-of-life (see #401 for details). Given that we now have general code in the music_services sub-package that makes it possible to search Spotify, I think it is better to focus on that.

If anyone wishes to bring the dedicated plugin up to date with the newer  API (and fix a few bugs) then they can always pull it out of git.

In any case, as the current state of things are, the plugin is completely broken and we have a substitute, so I proposed this hard deprecation.

### DIFF
--- a/soco/plugins/spotify.py
+++ b/soco/plugins/spotify.py
@@ -1,238 +1,34 @@
 # -*- coding: utf-8 -*-
-# pylint: disable = redefined-variable-type
 
-"""Spotify Plugin."""
+"""The Spotify plugin has been DEPRECATED
 
-from __future__ import unicode_literals
+The Spotify Plugin has been immediately deprecated (August 2016),
+because the API it was based on (The Spotify Metadata API) has been
+ended. Since this rendered the plug-in broken, there was no need to
+forewarn of the deprecation.
 
-import requests
+Please consider moving to the new general music services code (in
+soco.music_services.music_service), that makes it possible to
+retrived information about the available media from all music
+services. A short intro for how to use the new code is available
+in the API documentation:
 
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Made an amend commit with a few lint and doc fixes.

KennethNielsen:
Any comments on this?

robwebset:
If it's gone ... get rid of it ... I'm with you on this one :)

KennethNielsen:
@robwebset (or anyone else) any comments on the code changes?

robwebset:
I don't really use any of these bits - but what you have stated makes sense to me.

KennethNielsen:
Ok. Since this is in reality fairly undramatic, I'll merge it.

Issue #388, opened 2016-03-25
Error parsing AVTransport events when playing Deezer Tracks


pfischi:
An error in event handling rountine occurs, when a Deezer track is played. This happens because of an unknown item class "object.container.album.musicAlbum.#DEFAULT" in the event xml string.

As a workaround, i've added a new class to "data_structures.py" to solve the problem.
```
  class DidlMusicAlbumDefault(DidlAlbum):

    item_class = 'object.container.album.musicAlbum.#DEFAULT'
    tag = ''
    _translation = DidlAudioItem._translation.copy()
    _translation.update(
        {
            'artist': ('upnp', 'artist'),
            'genre': ('upnp', 'genre'),
            'producer': ('upnp', 'producer'),
            'toc': ('upnp', 'toc'),
            'album_art_uri': ('upnp', 'albumArtURI'),
        }
    )


```

With best regards,

Stefan

KennethNielsen:
Hi @pfischi and thanks for reporting this.

I'm fairly certain that this class is not part of the DIDL Lite specification and I think that we have seen it before i.e. a particular music service implementing some class that isn't in the spec.

It could probably get messy real fast if we add these music service specific classes every time, so I'm wondering if we couldn't come up with some general solution. It would probably involve an option to lessen the restriction in the from_didl_string function according to one of these options:

 1) In the event of an unknown class, simple say that everything is an object and return that while issuing a warning
 2) Assume that most likely the extra class is a simple extension upon the class one level up in the hierarchic and therefore simple remove the classes one level at a time until we get something we know, return that and issue a warning
 3) Or, to make it real obvious that we got an unknown class, make a UnkonwnObject class and use that (this time without a warning)

I tend to lean towards 2. It is a little bit of guesswork, but I think it matches both cases we have seen and it would probably be what would return the user with the most useful object.

What do you think? @lawrenceakka ?

KennethNielsen:
Hallo @pfischi 

I finally git around to try and fix this in a more general fasion by the method mentioned by @tagdara in #399.

I have tried to implement the fix in this branch (https://github.com/KennethNielsen/SoCo/tree/pound_subclass_fix) requested for pull in #425.

Can you test it?

Regards Kenneth

KennethNielsen:
This should hopefully be fixed with PR #425 which is now merged. Thanks for reporting it.

PR #425, opened 2016-08-24
Ignore .# specified subclasses in Didl xml

### LINKED ISSUE #399: Apple Music using #AlbumView causes crashes
Apple Music has a number of issues with Sonos, and this one seems to also affect the Windows Desktop Controller. 

When you Queue up an entire album, Soco fails on the first track change with an unknown UPNP type of object.container.album.musicAlbum.#AlbumView

Since I convert everything down to regular dicts in my app, it's very frustrating every time a new service is added which does something new.  Incidentally, the use of #AlbumView is even breaking the Windows controller, causing the player to stop and the queue to jump to the first track and loop the first 1-2 seconds over and over.  So this is clearly not just a Soco issue but rather a larger implementation issue with Sonos and Didl.

I'm assuming this has the same attributes as musicAlbum, so I monkeypatched data_structures with this which seems to resolve it for me, but I'm not sure if there are better ways to handle the continously changing Didl datastructures.  

class DidlMusicAlbumView(DidlAlbum):

    """Class that represents a music library album. """

    item_class = 'object.container.album.musicAlbum.#AlbumView'
    # According to the spec, all musicAlbums should be represented in
    # XML by a <container> tag. Sonos sometimes uses <container> and
    # sometimes uses <item>. Set the tag type to '' to indicate that
    # either is allowed.
    tag = ''
    # name: (ns, tag)
    # pylint: disable=protected-access
    _translation = DidlAudioItem._translation.copy()
    _translation.update(
        {
            'artist': ('upnp', 'artist'),
            'genre': ('upnp', 'genre'),
            'producer': ('upnp', 'producer'),
            'toc': ('upnp', 'toc'),
            'album_art_uri': ('upnp', 'albumArtURI'),
        }
    )

KennethNielsen:
Several music services seem to use a scheme for subclassing Didl classes by writing a subclass after the real class separated by .# e.g. object.container.album.musicAlbum.#AlbumView. This is not part of the Didl spec and so our implementation fails. This commit fixes this by stripping the sub-class, hence completely ignoring it. Fixes #399 and #388.

If anyone ever needs this information it should be simple enough to store it on the object while still using the real Didl class.

I consider this a bugfix, so it could be pulled without review, but since this has a bit of a design component to it, I though I would make a PR anyway. Also to allow the bug reporters to try the fix.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -86,6 +86,12 @@ def from_didl_string(string):
     for elt in root:
         if elt.tag.endswith('item') or elt.tag.endswith('container'):
             item_class = elt.findtext(ns_tag('upnp', 'class'))
+
+            # In case this class has an # specified unofficial
+            # subclass, ignore it by stripping it from item_class
+            if '.#' in item_class:
+                item_class = item_class[:item_class.find('.#')]
+
             try:
                 cls = _DIDL_CLASS_TO_CLASS[item_class]
             except KeyError:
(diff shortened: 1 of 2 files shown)


KennethNielsen:
As a fix similar to this has previously been tested I'll go ahead and merge it.

PR #437, opened 2016-10-24
Event listener thread safety

bjarniivarsson:
EventListener.start would sometimes blow up if attempting to subscribe to multiple services from different threads. This happens when for example you have a thread per-speaker subscribing to events. This also fixes a thread crash that would occur when receiving events that you have not subscribed for (no service for sid). This may happen when restarting applications based on SoCo.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -228,8 +228,10 @@ class EventNotifyHandler(BaseHTTPRequestHandler):
             variables = parse_event_xml(content)
             # Build the Event object
             event = Event(sid, seq, service, timestamp, variables)
-            # pass the event details on to the service so it can update its cache.
-            if service is not None:  # It might have been removed by another thread
+            # pass the event details on to the service so it can update its
+            # cache.
+            if service is not None:
+                # It might have been removed by another thread
                 # pylint: disable=protected-access
                 service._update_cache_on_event(event)
             # Find the right queue, and put the event on it
@@ -317,7 +319,8 @@ class EventListener(object):
         with self._start_lock:
             if not self.is_running:
                 temp_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-                temp_sock.connect((any_zone.ip_address, config.EVENT_LISTENER_PORT))
+                temp_sock.connect((any_zone.ip_address,
+                                   config.EVENT_LISTENER_PORT))
                 ip_address = temp_sock.getsockname()[0]
                 temp_sock.close()
                 # Start the event listener server in a separate thread.
--- a/soco/events.py
+++ b/soco/events.py
@@ -220,23 +220,26 @@ class EventNotifyHandler(BaseHTTPRequestHandler):
         # find the relevant service from the sid
         with _sid_to_service_lock:
             service = _sid_to_service.get(sid)
-        log.info(
-            "Event %s received for %s service on thread %s at %s", seq,
-            service.service_id, threading.current_thread(), timestamp)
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Hi @bjarniivarsson 

Great with thread safety fixes. I must admit that I always find the combination of network and threaded code complicated and I therefore find it a little difficult to follow the fix. Can you explain, in a little more detail, for each of the two fixes, what would go wrong, how it would manifest itself, and how this fixes it?

Thanks in advance.

bjarniivarsson:
Hey @KennethNielsen 

This fixes two issues:
- First one (line 223, "if service:") fixes an issue where if you have a SoCo based application that has subscribed to events from a speaker and you restart it, you will get events from your previous subscription sent to you, and the sid will not resolve to a service instance. When this happens SoCo would spam the log with an exception (basically that NoneType has no property called "service_id" when trying to log the event). The fix is to simply ignore the event, but still return 200 to the speaker with the assumption that a new subscription will be made when the process has discovered the speaker again. 

- The second one (line 318, "self._start_lock) occurs when you have a multiple threads making subscriptions at the same time. You could easily get into a situation where one thread had started the server, but not set self._is_running to True when the second thread comes along and also tries to start the server (causing a an exception on socket.bind())

I'm getting these issues when working on improvements to Sonos handing in home-assistant, see this PR: https://github.com/home-assistant/home-assistant/pull/4063

bjarniivarsson:
This is the stack trace for the first one:
Exception happened during processing of request from ('192.168.1.58', 55463)
  File "/usr/lib/python3.4/socketserver.py", line 613, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.4/socketserver.py", line 344, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.4/socketserver.py", line 669, in __init__
    self.handle()
  File "/usr/lib/python3.4/http/server.py", line 398, in handle
    self.handle_one_request()
  File "/usr/lib/python3.4/http/server.py", line 386, in handle_one_request
    method()
  File "/srv/homeassistant/lib/python3.4/site-packages/soco/events.py", line 225, in do_NOTIFY
    service.service_id, threading.current_thread(), timestamp)
AttributeError: 'NoneType' object has no attribute 'service_id'

KennethNielsen:
Got it. Thanks.

soco/events.py
```diff
@@ -220,23 +220,28 @@ def do_NOTIFY(self):  # pylint: disable=invalid-name
...
+            log.info(
+                "Event %s received for %s service on thread %s at %s", seq,
+                service.service_id, threading.current_thread(), timestamp)
+            log.debug("Event content: %s", content)
+            variables = parse_event_xml(content)
+            # Build the Event object
+            event = Event(sid, seq, service, timestamp, variables)
+            # pass the event details on to the service so it can update its
+            # cache.
+            if service is not None:
```
KennethNielsen: I guess, strictly speaking, the if check on line 233 is no longer needed, since by your addition it is checked on line 223.

bjarniivarsson: Yes, you are correct, will fix

KennethNielsen:
Ah ok. One more (but a follow up), then I'm done nit-picking. I guess the comment that went along with the if check you just removed:
```python
# It might have been removed by another thread
```
should now be moved up to before ```if service:```.

KennethNielsen:
Merged. Thanks. Can you add a note about the change to the release notes issue #440?

PR #444, opened 2016-11-02
Allow events to use config IP

lwis:


lawrenceakka:
Hi,

Thanks for this but please can you explain what it is for? What particular problem are you trying to solve?

lwis:
@lawrenceakka this was created off the back of home-assistant/home-assistant#4177, essentially when running in Docker containers the host IP isn't obtainable/exposed - so when SoCo binds and advertises itself for the event stream to other speakers we end up in a situation where it's sending out the IP of the Docker container, which the speakers cannot connect to.

soco/config.py
```diff
@@ -31,6 +31,16 @@
 """
 
 
+EVENT_LISTENER_IP = None
+"""The IP on which the event listener listens.
+
+The default is None. If unset it will detect.
```
lawrenceakka: Please change to "The default of None means that the relevant IP address will be detected automatically"

soco/events.py
```diff
@@ -317,11 +317,16 @@ def start(self, any_zone):
         # Sonos net, see http://stackoverflow.com/q/166506
         with self._start_lock:
             if not self.is_running:
-                temp_sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-                temp_sock.connect((any_zone.ip_address,
-                                   config.EVENT_LISTENER_PORT))
-                ip_address = temp_sock.getsockname()[0]
-                temp_sock.close()
+                if config.EVENT_LISTENER_IP:
```
lawrenceakka: Please add a comment just before this line (We like comments 😄 ):

#Use configured IP address if there is one, else detect automatically

lwis:
@lawrenceakka done, cheers bud!

lawrenceakka:
No other comments received, so I'm committing this. Thanks very much.

PR #450, opened 2016-11-20
Event advertisement IP config.

lwis:
Sorry, I apparently missed this piece of the puzzle from #444.

### DIFF
--- a/soco/__init__.py
+++ b/soco/__init__.py
@@ -17,7 +17,7 @@ from .exceptions import SoCoException, UnknownSoCoException
 
 # Will be parsed by setup.py to determine package metadata
 __author__ = 'The SoCo-Team <python-soco@googlegroups.com>'
-__version__ = '0.12.0'
+__version__ = '0.12'
 __website__ = 'https://github.com/SoCo/SoCo'
 __license__ = 'MIT License'
 
--- a/soco/__init__.py
+++ b/soco/__init__.py
@@ -17,7 +17,7 @@ from .exceptions import SoCoException, UnknownSoCoException
 
 # Will be parsed by setup.py to determine package metadata
 __author__ = 'The SoCo-Team <python-soco@googlegroups.com>'
-__version__ = '0.11.1'
+__version__ = '0.12.0'
 __website__ = 'https://github.com/SoCo/SoCo'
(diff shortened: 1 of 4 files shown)


lwis:
@lawrenceakka thoughts?

lawrenceakka:
Sorry - haven't had time to test or think about this yet. 

1) I'm not sure why you would ever want a different listener address (which is the address on which Soco listens for events) and the advertisement address (which is, as far as I can see, the address which Soco tells the Sonos devices it is using to listen for events)!

2) Does it actually work, especially if you different values for the two IP address config variables. Just looking at the code, it doesn't look to me like it will. 

3) Also, it looks like you are failing some of the travis checks due to extra blank lines!

lwis:
@lawrenceakka use case;

I'm in a Docker container, when it tries to work out it's IP it will get the container IP, which will ultimately send the container IP to the Sonos devices (which is only reachable locally on the machine).

This does work, I'll do some more testing and fix the failing builds. Thanks for getting back to me.

lawrenceakka:
Should event_listener_ip default to event_advertisement_ip (so that if you have set the first, the second by default is the same)?

lwis:
@lawrenceakka not exactly, as containers cannot bind to the host IP, however can listen on it (by proxy).

soco/__init__.py
```diff
@@ -17,7 +17,7 @@
 
 # Will be parsed by setup.py to determine package metadata
 __author__ = 'The SoCo-Team <python-soco@googlegroups.com>'
-__version__ = '0.11.1'
+__version__ = '0.12'
```
ghcs27: Just remove this change to resolve the conflicts.

adamcstephens:
Can we resurrect this PR? I'm running into the same issue as @lwis with a docker container on a bridged network that can't use events because bind IP != advertise IP.

ghcs27:
Unfortunately, we can't reopen this since the @lwis' branch does not exist anymore.
You can create your own branch, do the [changes](https://github.com/SoCo/SoCo/pull/450/files) and submit a new PR.

Juggels:
@ghcs27, I have just created PR https://github.com/SoCo/SoCo/pull/604 for this

Issue #440, opened 2016-10-26
Release notes for 0.13


KennethNielsen:
For every change that goes into the 0.13 release. Please make a small description here, link to the PR (any issues) and mention whether it is a bug fix or not.

bjarniivarsson:
https://github.com/SoCo/SoCo/pull/437
- Fixes an issue where restarting an application that had subscribed to events sometimes causes an error when the events are delivered to the new instance.
- Fixes an issue where multiple threads trying to subscribe to events in parallel would sometimes cause SoCo to attempt to create multiple event listener servers and fail on socket.bind().

bjarniivarsson:
https://github.com/SoCo/SoCo/pull/438
- Fixes an issue where SoCo would not recognize object.container.playlistContainer.sonos-favorite when receiving events.

KennethNielsen:
#443 
 * Fixes a bug in play_uri where it would not play a http og https prefixed radio stream due to a change in the Sonos API (issue #434). This change fixes it by replacing the two http type prefixes with Sonos' x-rincon-mp3radio:// prefix.

KennethNielsen:
1ee005a8931e6c18d803ab4d589e6f45544ddb1c
 * Fixes an exception being raised on Windows with discover. The error was caused by socket.getsockname raising an exception on Windows with and unconnected unbound socket. Fixed by now simply logging the sockets.

lawrenceakka:
#444 The IP address used by the events listener can be configured

christiandt:
- #421 Add support for night mode on devices supporting the feature.
- #422 Add support for dialog mode on devices supporting the feature.
- Add NotSupportedException when mode is not supported by the device.

KennethNielsen:
* #455 Add queue-able data structures for the music service items (see #455 for details)

ghcs27:
- #470 Add a method for queueing multiple items with a single request.

KennethNielsen:
* #481 Added a FAQ section to the docs with play_uri and play local files answers

KennethNielsen:
* #484 A few queue related micro examples was added to the examples page in the docs

ghcs27:
- #471 Enhance `add_to_queue()` to with optional position argument

KennethNielsen:
* #487 Added by_name function to discovery to be able to get a device by player_name

DPH:
- #489 Snapshot: added documentation about using only once and fixed Python 3 potential issue

DPH:
[...]

PR #459, opened 2016-12-27
Updating the example mp3 file because the old one was redirecting with a 302

ersiko:


KennethNielsen:
Hi @ersiko 

Merged.

Can you include a comment about the change in #440.

Regards Kenneth

PR #300, opened 2015-04-17
Added support play line in from another Sonos player

davey-g-uk:
Current switch_to_line_in method does not support playing a line in source from another Sonos player. Amendments below allow method to be called with a SoCo instance as an argument to specify source line in to play. In same manner join method is called.

If Instance is omitted default behaviour will be to play current instances line in.

PR #460, opened 2017-01-01
Support to play line in from another Sonos player with test

ghcs27:
This is a copy of #300 with the requested test case for the parameter.
Can someone please check my test code (it's my first test with Mock)?

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -925,20 +925,16 @@ class SoCo(_SocoSingletonBase):
             source (SoCo): The speaker whose line-in should be played.
                 Default is line-in from the speaker itself.
 
-        Returns:
-            True if the Sonos speaker successfully switched to line-in.
-
-        If an error occurs, we'll attempt to parse the error and return a UPnP
-        error code. If that fails, the raw response sent back from the Sonos
-        speaker will be returned.
-
-        Raises SoCoException (or a subclass) upon errors.
+        Raises:
+             SoCoException: Upon errors, we'll attempt to parse the error and
+                return a UPnP error code. If that fails, the raw response sent
+                back from the Sonos speaker will be returned.
 
         """
-        if not source:
-            uid = self.uid
-        else:
+        if source:
             uid = source.uid
+        else:
+            uid = self.uid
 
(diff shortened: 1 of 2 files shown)


soco/core.py
```diff
@@ -918,11 +918,15 @@ def unjoin(self):
...
 
-    def switch_to_line_in(self):
+    def switch_to_line_in(self, source=None):
         """ Switch the speaker's input to line-in.
 
+        Args:
+            source (SoCo): The speaker whose line-in should be played.
+                Default is line-in from the speaker itself.
+
         Returns:
```
KennethNielsen: This part of the docstring appears to obsolete. We switched a ling time ago away from returning booleans in indicate success. I know you didn't introduce it, but might as well update while we are at it. You can just delete the returns part.

soco/core.py
```diff
@@ -931,10 +935,14 @@ def switch_to_line_in(self):
         Raises SoCoException (or a subclass) upon errors.
```
KennethNielsen: The Raises stuff is also Napoleon syntax but should be written in the same way as the Returns section i.e:

```python
"""
...
Raises:
    SoCoException: (or a subclass) upon errors.
"""

and this raises section I think should be placed after the args sections

File: soco/core.py [hunk 3]
```diff
@@ -931,10 +935,14 @@ def switch_to_line_in(self):
...
 
         """
+        if not source:
```
KennethNielsen: Since we are relying on the boolean value of a SoCo instance anyway, why not turn it around:
```python
if source:
    ...
else:
    ....
```

tests/test_core.py
```diff
@@ -499,6 +499,16 @@ def test_switch_to_line_in(self, moco_zgs):
                 ('CurrentURIMetaData', '')]
         )
 
+        moco_zgs.avTransport.reset_mock()
```
KennethNielsen: This tests here look good.

KennethNielsen:
Looks good, merged. Thanks for the contribution. Can you add a comment about this change to the release notes issue #440

PR #466, opened 2017-01-06
Fix #453 avTransport

Cereal2nd:
Attached is a fix, this one at least works for google play

Issue #453, opened 2016-12-08
avTransport events throw exception when connected via spotify direct


maru-sama:
The latest stable sonos update finally supports spotify direct connect. While this works fine it apparently breaks avTransport events. If I play a track via spotify direct events throw following exception

Exception happened during processing of request from ('192.168.0.14', 35521)
<snip>
File "/Users/maru/source/python/virtual/kivy/lib/python2.7/site-packages/soco/data_structures.py", line 504, in from_element
    raise DIDLMetadataError("Missing restricted attribute")
DIDLMetadataError: Missing restricted attribute

For debugging purposes I put a debug print to see the returned element. Apparently I get THREE 

`<item xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" id="-1" parentID="-1"><res duration="0:04:10" protocolInfo="http-get:*:audio/x-spotify:*">x-sonos-spotify:spotify%3atrack%3a3zmOCkma2GQdMV4AZYLfZB?sid=9&amp;flags=0&amp;sn=9</res><upnp:albumArtURI>https://i.scdn.co/image/131f47a6be760db779ea6b0789235a224f19e9cb</upnp:albumArtURI><upnp:class>object.item.audioItem.musicTrack</upnp:class><dc:title>Fashionable People</dc:title><dc:creator>Joel Plaskett Emergency</dc:creator><r:albumArtist>Joel Plaskett Emergency</r:albumArtist><upnp:album>Ashtray Rock</upnp:album><r:tiid>591805489</r:tiid></item>
`

[...]

Issue #448, opened 2016-11-14
Empty 'title' in function add_uri_to_queue() caused an AvTransport error


pfischi:
An error is thrown by the utils.py function 'really_unicode()' if an uri without metadata (especially with empty title) is added to the queue by 'add_uri_to_queue()'. I my case, i've added a local mp3-file tothe queue as follows:

```soco.add_uri_to_queue('x-rincon-mp3radio://192.168.178.22:8802/ZGVfX0hhbGxvIEhhbGxvIER1.mp3')```

The file is played correctly but this track caused errors triggered by AVTransport events.

```Traceback (most recent call last):
  File "/usr/lib/python3.4/socketserver.py", line 617, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib/python3.4/socketserver.py", line 344, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib/python3.4/socketserver.py", line 673, in __init__
    self.handle()
  File "/usr/lib/python3.4/http/server.py", line 398, in handle
    self.handle_one_request()
  File "/usr/lib/python3.4/http/server.py", line 386, in handle_one_request
    method()
  File "/home/sb/projects/shSonos/server.sonos/soco/events.py", line 229, in do_NOTIFY
    variables = parse_event_xml(content)
  File "/home/sb/projects/shSonos/server.sonos/soco/events.py", line 130, in parse_event_xml
    value = from_didl_string(value)[0]
  File "/home/sb/projects/shSonos/server.sonos/soco/data_structures.py", line 99, in from_didl_string
    items.append(cls.from_element(elt))
  File "/home/sb/projects/shSonos/server.sonos/soco/data_structures.py", line 513, in from_element
    title = really_unicode(title_elt.text)
  File "/home/sb/projects/shSonos/server.sonos/soco/utils.py", line 43, in really_unicode
    raise ValueError('%s is not a string at all.' % in_string)
ValueError: None is not a string at all.
```

If i change the line in "add_uri_to_queue()" from 

```item = DidlObject(resources=res, title='', parent_id='', item_id='')```
to
```item = DidlObject(resources=res, title='here a title', parent_id='', item_id='')```

everything seems to be fine.

Stefan

KennethNielsen:
Thanks for reporting this. @ghcs27 made a PR to fix this (#467). If you have the option to test that branch that would be great.

DPH:
@pfischi - not heard back for some time so closing this issue. 
[...]

PR #409, opened 2016-07-07
Event bugfixes.

bosim:
When using events two bugs occur
- When using Queue you get a exception when recieving messages.
- When subscribing to transport or similar event you get exception since title is missing.

I don't say it is the right fixes, but it works at least.

PR #467, opened 2017-01-06
Handle none values properly and fix avTransport events

### LINKED ISSUE #458: ValueError when handling event
I have an avTransport subscription from a Sonos speaker, and every now and again I get a value error:
```
Exception happened during processing of request from ('192.168.123.16', 4911)
  File "usr/lib/python3.4/socketserver.py", line 613, in process_request_thread
    self.finish_request(request, client_adress)
  File "usr/lib/python3.4/socketserver.py", line 344, in finish_request
    self.RequestHandlerClass(request, client_adress, self)
  File "usr/lib/python3.4/http/server.py", line 398, in handle
    self.handle_one_request()
  File "usr/lib/python3.4/http/server.py", line 386, in handle_one_request
    method()
  File "usr/local/lib/python3.4/dist-packages/soco/events.py", line 227, in do_NOTIFY
    variables = parse_event_xml(content)
  File "usr/local/lib/python3.4/dist-packages/soco/events.py", line 130, in parse_event_xml
  File "usr/local/lib/python3.4/dist-packages/soco/data_structures.py", line 93, in from_didl_string
    items.append(cls.from_element(elt))
  File "usr/local/lib/python3.4/dist-packages/soco/data_structures.py", line 501, in from_element
    title = really_unicode(title_elt.text)
  File "usr/local/lib/python3.4/dist-packages/soco/utils.py", line 43, in really_unicode
    raise ValueError('%s is not a string at all.' % in_string)
ValueError: None is not a string at all.
```
It seems like in line 501 of data_structures.py the title tag is present in the XML, but has no content. Maybe title_elt.text should be checked for None values and replaced with an empty string.
Would someone please take a look at this?

I am using soco 0.11.1 on Python 3.4.

ghcs27:
This is a combined PR for #465 and #466 because both change the same code.
It fixes #458, #448 and #453.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -491,14 +491,16 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
                 "UPnP class is incorrect. Expected '{0}',"
                 " got '{1}'".format(cls.item_class, item_class))
 
-        # parent_id, item_id  and restricted are stored as attibutes on the
+        # parent_id, item_id  and restricted are stored as attributes on the
         # element
-        item_id = really_unicode(element.get('id', None))
+        item_id = element.get('id', None)
         if item_id is None:
             raise DIDLMetadataError("Missing id attribute")
-        parent_id = really_unicode(element.get('parentID', None))
+        item_id = really_unicode(item_id)
+        parent_id = element.get('parentID', None)
         if parent_id is None:
             raise DIDLMetadataError("Missing parentID attribute")
+        parent_id = really_unicode(parent_id)
         restricted = element.get('restricted', None)
         if restricted is None:
             raise DIDLMetadataError("Missing restricted attribute")
@@ -510,7 +512,7 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
         if title_elt is None:
             raise DIDLMetadataError(
                 "Missing title element")
-        title = really_unicode(title_elt.text)
+        title = really_unicode('' if title_elt.text is None else title_elt.text)
 
         # Deal with any resource elements
         resources = []
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -512,7 +512,10 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
         if title_elt is None:
             raise DIDLMetadataError(
                 "Missing title element")
-        title = really_unicode('' if title_elt.text is None else title_elt.text)
+        title = title_elt.text
+        if title_elt.text is None:
+            title = ''
+        title = really_unicode(title)
 
         # Deal with any resource elements
         resources = []
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -501,10 +501,13 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
         if parent_id is None:
             raise DIDLMetadataError("Missing parentID attribute")
         parent_id = really_unicode(parent_id)
+
+        # CAUTION: This implementation deviates from the spec.
+        # Elements are normally required to have a `restricted` tag, but
+        # Spotify direct violates this. To make it work, a missing restricted
+        # tag is interpreted as `restricted = True`.
         restricted = element.get('restricted', None)
-        if restricted is None:
-            raise DIDLMetadataError("Missing restricted attribute")
-        restricted = True if restricted in [1, 'true', 'True'] else False
+        restricted = False if restricted in [0, 'false', 'False'] else True
 
         # There must be a title. According to spec, it should be the first
         # child, but Sonos does not abide by this


Cereal2nd:
let hope this get merged quickly
as we really need this fix in domogik

ghcs27:
can please test if this works for you?

Cereal2nd:
no this one is not working for me ...

title is always empty, with my changes only title was filled in

ghcs27:
realized my mistake: I checked for `not title_elt` instead of `title_elt is None`
Now it should work.

Cereal2nd:
This one works

ghcs27:
Thankyou for checking

ghcs27:
The flake8 complaints are not related to the code I have changed. Is there anything that's keeping this from getting merged?

ghcs27:
Is there anything that's keeping this from getting merged?

soco/data_structures.py
```diff
@@ -491,26 +491,26 @@ def from_element(cls, element):     # pylint: disable=R0914
...
+        item_id = really_unicode(item_id)
+        parent_id = element.get('parentID', None)
         if parent_id is None:
             raise DIDLMetadataError("Missing parentID attribute")
+        parent_id = really_unicode(parent_id)
         restricted = element.get('restricted', None)
-        if restricted is None:
-            raise DIDLMetadataError("Missing restricted attribute")
-        restricted = True if restricted in [1, 'true', 'True'] else False
+        restricted = False if restricted in [0, 'false', 'False'] else True
```
KennethNielsen: Didn't look through all the bug reports that this is supposed to fix yet. But this changes the behaviour so that a missing restricted field will no longer generate an error and will default to True. Is this the intended behaviour?

Cereal2nd:
Yes this is intended, as Google play service is not sending it ...

KennethNielsen:
In that case I think this should be merged. I'll wait a few days to allow some of the bug reporters time to test it. If no testing turns up I'll merge it as is.

ghcs27:
Should I squash the commits?

KennethNielsen:
@ghcs27 that's your call. The commits such as they are, show the progression of the patch pertaining to actual functionality (except for the "too long line one", which is purely style). Whether you want to squash to get rid of that one commit is your call. Whether you want to squash more is really up to your view of the work. Commits should preferably contains "small isolated packages of work".

Do note however, if you do, you do not need to open a new PR for it. Just squach in your local branch and force push to the ghcs27:handle_none_values and make a comment here about it, that's enough.

ghcs27:
OK, then I'll just keep the commits.

KennethNielsen:
This is all good then. I will let it sit for a week in total from I commented in the other bugs asking for testing and then I'll merge it.

ghcs27:
I think this can be merged now!

erinxocon:
Is this pull request ready to be merged?  I'm trying to use this library in a project, but I connect to the Sonos speakers via Spotify connect.  Thanks!

maru-sama:
Hello, this apparently fixes the #453 for me.
That said there is still an error afterwards since apparently the title attribute is missing, but the initial reported problem

DIDLMetadataError: Missing restricted attribute

has been fixed.

Can this be merged?

ghcs27:
@KennethNielsen could you please merge?

KennethNielsen:
@ghcs27 uhh, sorry about that. When I saw this pop up in my email queue again I had a thought. While I have no problem meriging this as is, I feel like events are forcing the events and data structures code into the wild west.  So 2 thoughts that I wanted to share/discuss before merging:

As far as I remember from reviewing it, the existing data structures code was fairly spec compliant. In general, this is a good thing, because it means that you have somewhere to check if the implementation is correct. So any changes to the code, that changes it away from being spec compliant, really should be:

 1) Documenter. Can I get you to add very clear comments in the places where you are changing semantics away from the spec?
 2) Tested. Ideally, although maybe a little too late now, we should collect test cases from the issues and add those to the test suites, so that we can ensure that future changes will still allow these non-compliant events through. I don't know if you feel up to that. My fear is, that if I merge, probably it will be a little more difficult to convince bug reporters to collect test cases. On the other hand, this has also been sitting around for so long, that it might also end up annoying people that it doesn't just get merged. I'll take your input on this.

Regards Kenneth

jake:
FWIW, I ws able to get it to crash using the [sample event subscription code](https://github.com/SoCo/SoCo/blob/11515c44332b98b4011c3e07509ba6f54fd32d96/soco/events.py#L50) and playing from Spotify Connect (get on the same wifi as your sonos, and in the spotify app, select your sonos as a speaker option). Not sure how to turn that into a test case, or if that helps.

maru-sama:
Hello Kenneth,

Instead of creating exceptions what about the possibility to let the event handler drop/ignore non-conformant events or flag them as broken. This could be made a configuration setting when you subscribe for events

strict .... Current behaviour
ignore .... Ignore non conformant events
relaxed ... Try to get as much information out of the xml and forward this with a "broken" flag. 

In my specific case (spotify direct) I get 3 events for one song change so my idea is to ignore the content of the events and poll the player for the metadata instead.

KennethNielsen:
@jake I could help you turn that into a test case, but I would need someone that has the time to write it up first, so let's see what happens with this discussion.

KennethNielsen:
@maru-sama that isn't a bad idea although it would add a little extra complexity. Unfortunately, I think we are knee-deep in "relaxed" already, so the idea of "strict", as in spec compliant is probably lost. It might be possible to allow switching between "relaxed" and "ignore" with relatively little complexity. Could you post that idea as a separate issue?

KennethNielsen:
@ghcs27 let me know if you think I'm being a pain here. If you prefer to have this merged quicker, since it's been sitting around for so long, then just address 1 and I will merge. Then we can try to make a separate issue about collecting non-compliant events.

jake:
I'd vote for getting this in instead of waiting for an ideal fix, since it fixes a pretty much fatal error, rendering master unusable in this scenario.

KennethNielsen:
@jake yes, we always want the fix, but we also want maintainable code 2 years down the road ;)

Cereal2nd:
true, but now the code is not use able for some services.
I think we should merge this, and create a new task issue to implement the
real and maintainable fix

maru-sama:
Well this patch still does not fully fix the spotify direct issue because the title attribute is missing too. I mentioned this in one of my earlier comments. It "just" fixes the issue of the missing restricted tag.

ghcs27:
@KennethNielsen I've added a warning label concerning the non-compliance to the spec, I hope it fits your expectation.
I won't be able to write the test soon, so please consider merging without it, but I'll get around to it sometime.
As for @maru-sama's suggestion, making it a separate issue at #536 is probably the best, as it does not only concern the changes in this PR

@maru-sama For the missing title attribute, please see #409.

ghcs27:
@maru-sama For the missing title attribute, please see #409.

amelchio:
Hmm, the data_structures.py change in #409 appears very similar to the one in this PR? They actually conflict so both cannot be applied at the same time.

The problem is a "Missing title element" exception because `title_elt is None` and neither PR seems to handle that situation.

KennethNielsen:
Ok. I will merge this and then we can sort out test cases and #409 afterwards.

KennethNielsen:
@ghcs27 can you add a comment to #440 about the change

KennethNielsen:
@Cereal2nd @erinxocon @maru-sama @jake @amelchio I would very much like help, in collecting these non-spec compliant events that make the code break, if any of you feel up to it. Please see #545 for instruction about how to collect them.

KennethNielsen:
Also, please see #536 for a drafted solution to make the events code more resilient all over against non-spec compliant events.

PR #470, opened 2017-01-14
Add multiple items to queue

ghcs27:
This PR adds a `SoCo.add_multiple_to_queue()` method, which takes a list of queueable items and adds them to the queue with a single request. For spec compatibility, it also has a `container` parameter, which should be the superordinate Didl container, but since it also works without a container, the default is None.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -138,6 +138,7 @@ class SoCo(_SocoSingletonBase):
         get_current_transport_info
         add_uri_to_queue
         add_to_queue
+        add_multiple_to_queue
         remove_from_queue
         clear_queue
         get_favorite_radio_shows
@@ -1283,6 +1284,39 @@ class SoCo(_SocoSingletonBase):
             ('EnqueuedURIMetaData', metadata),
             ('DesiredFirstTrackNumberEnqueued', 0),
             ('EnqueueAsNext', 1)
+        ])
+        qnumber = response['FirstTrackNumberEnqueued']
+        return int(qnumber)
+
+    def add_multiple_to_queue(self, items, container=None):
+        """Adds a sequence of items to the queue.
+
+        Args:
+            items (list): A sequence of items to the be added to the queue
+            container (DidlObject, optional): A container object which
+                includes the items.
+
+        Returns:
+            int: The index at which the items were enqueued.
+        """
+        uris = ' '.join([item.resources[0].uri for item in items])
+        uri_metadata = ' '.join([to_didl_string(item) for item in items])
+        if container is not None:
+            container_uri = container.resources[0].uri
+            container_metadata = to_didl_string(container)
+        else:
+            container_uri = ''
+            container_metadata = ''  # pylint: disable=redefined-variable-type
+        response = self.avTransport.AddMultipleURIsToQueue([
+            ('InstanceID', 0),
+            ('UpdateID', 0),
+            ('NumberOfURIs', len(items)),
+            ('EnqueuedURIs', uris),
+            ('EnqueuedURIsMetaData', uri_metadata),
+            ('ContainerURI', container_uri),
+            ('ContainerMetaData', container_metadata),
+            ('DesiredFirstTrackNumberEnqueued', 0),
+            ('EnqueueAsNext', 0)
         ])
         qnumber = response['FirstTrackNumberEnqueued']
         return int(qnumber)


ghcs27:
Is there anything I can do to move this forward?

KennethNielsen:
Hallo @ghcs27 

This looks great.

I can confirm that this works for me on both tracks and albums from the music library. Also, which is really nice, it works directly on the search results (which I guess is a result of them being iterable), that is:
```python
z = any_soco()
result = z.music_library.get_albums(search_term="black", start=1, max_items=3)
z.add_multiple_to_queue(result)
```
works, just as well as had it been an ordinary list:
```python
z = any_soco()
result = list(z.music_library.get_albums(search_term="black", start=1, max_items=3))
z.add_multiple_to_queue(result)
```

Will fo a review shortly.

soco/core.py
```diff
@@ -1287,6 +1288,35 @@ def add_to_queue(self, queueable_item):
         qnumber = response['FirstTrackNumberEnqueued']
         return int(qnumber)
 
+    def add_multiple_to_queue(self, item_list, container=None):
```
KennethNielsen: suggest to make the argument name "items", not item list, since the only requirement of the code is that it is a sequence of items.

This applies also to the entire docstring, where all references to this argument should list it as a sequence of items, with the only exceptions being the argument type, where it makes sense to list it as a list.

soco/core.py
```diff
@@ -1287,6 +1288,35 @@ def add_to_queue(self, queueable_item):
...
+        """Adds a list of items to the queue.
+
+        :param item_list: The list of items to be added to the queue
```
KennethNielsen: We are in the process of migrating all doc strings to fit with the Napoleon extension for  Sphinx and its syntax (although it isn't always that obvious, because we are not that long into the process. But all new doc strings should follow it e.g:
```python
Args:
    items (list): A sequence of items to the be added to the queue
```
etc.

soco/core.py
```diff
@@ -1287,6 +1288,35 @@ def add_to_queue(self, queueable_item):
...
+            ('NumberOfURIs', len(item_list)),
+            ('EnqueuedURIs', uris),
+            ('EnqueuedURIsMetaData', uri_metadata),
+            ('ContainerURI', container_uri),
+            ('ContainerMetaData', container_metadata),
+            ('DesiredFirstTrackNumberEnqueued', 0),
+            ('EnqueueAsNext', 0)
+        ])
+        qnumber = response['FirstTrackNumberEnqueued']
+        return int(qnumber)
```
KennethNielsen: Remember to document the return type

KennethNielsen:
If you are up to it, it would also be cool if you could squash the commits after making the revisions. 4 commits for a change this size is a bit much. It would probably fit nicely into 1.

KennethNielsen:
If you are up to it, it would also be great with a tests for it. You can probably draw inspiration from add_to_queue

ghcs27:
@KennethNielsen Thanks for the review, I'm on it.
Could you please give me a hint where the test for `add_to_queue()` is?

ghcs27:
@KennethNielsen Could you please take a look at my commits? I would not mind if I need to change something.

soco/core.py
```diff
@@ -1287,6 +1288,39 @@ def add_to_queue(self, queueable_item):
...
+    def add_multiple_to_queue(self, items, container=None):
+        """Adds a list of items to the queue.
+
+        Args:
+            items (list): A sequence of items to the be added to the queue
+            container (DidlObject): (optional) A container object which includes
+                the items.
+
+        Returns:
+            The index at which the items were enqueued.
```
KennethNielsen: You need a type on the return as well:
```python
Returns:
    int: The index at which ...

File: soco/core.py [hunk 5]
```diff
@@ -1287,6 +1288,39 @@ def add_to_queue(self, queueable_item):
         qnumber = response['FirstTrackNumberEnqueued']
         return int(qnumber)
 
+    def add_multiple_to_queue(self, items, container=None):
+        """Adds a list of items to the queue.
```
KennethNielsen: sequence of items

KennethNielsen:
With regards to the tests for the ordinary add_to_queue, it seems like there are none, but it would still be great if you feel like writing some this method.

How you would probably structures it, is to mock out the service method `self.avTransport.AddMultipleURIsToQueue` and check that the mock gets called with the expected input for some prespecified inputs. The tests for core are in https://github.com/SoCo/SoCo/blob/master/tests/test_core.py,

KennethNielsen:
You can probably draw some inspiration from `test_add_item_to_sonos_playlist`test_add_item_to_sonos_playlist`

KennethNielsen:
Hi @ghcs27 

I'd like your thoughts on how you want to continue here.

 * Regarding the tests. We do not have a policy of requiring them on new work (merely requesting), so if you want to have this merged and push that of to another time, that's no problem. If you would like to do the work, but want more concrete help that I have given so far, just let me know.
 * There are a few lint issue you will have to take a look at. Let me know if you need help.
 * Regarding squashing the commits. If you don't already know the procedure this requires an interactive re-base of the last part of the git history, which is a slightly advanced feature. If you don't feel up to it, I can do it during the merge. If you would like to try, but need instructions, let me know.

Looking forward to hear from you.

ghcs27:
I have squashed the commits and fixed one lint issue, but I just can't seem to fix the remaining pylint complaint: 
```
************* Module soco.core
R:1390,12: Redefinition of container_metadata type from str to unicode (redefined-variable-type)
```
Could you please give me a clue what I should do to fix that?

KennethNielsen:
@ghcs27 I can certainly understand why it confuses. I think there is a bug in to_didl_string with returning different types in python 2.7 vs. 3, but I'm to tired to check for sure right now. The return value type in Python 3 is definitely right, and that passes lint. So I'm ok with you just disabling the warning until I get it cleared up.

ghcs27:
Added the pylint ignore.

KennethNielsen:
@ghcs27 merged. Thanks for the contribution.

Can you also please add a comment about the improvement in this issue #440 for the release notes of the next version.

PR #471, opened 2017-01-14
add_to_queue() at given position

ghcs27:
This PR adds optional `position` and `as_next` arguments to `add_uri_to_queue()` and `add_to_queue()` to specify at which position the item should be inserted into the queue and whether this item should be played next.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1261,28 +1261,36 @@ class SoCo(_SocoSingletonBase):
                                                                 **kwargs)
 
     @only_on_master
-    def add_uri_to_queue(self, uri):
+    def add_uri_to_queue(self, uri, position=0, as_next=False):
         """Adds the URI to the queue.
 
         :param uri: The URI to be added to the queue
         :type uri: str
+        :param position: The index (1-based) at which the URI should be added.
+            Default is 0 (add URI at the end of the queue).
+        :type position: int
+        :param as_next: Whether this URI should be played as the next track.
+            This can be used to set the next track even when
+            `play_mode=SHUFFLE`.
+        :type as_next: bool
         """
         # FIXME: The res.protocol_info should probably represent the mime type
         # etc of the uri. But this seems OK.
         res = [DidlResource(uri=uri, protocol_info="x-rincon-playlist:*:*:*")]
         item = DidlObject(resources=res, title='', parent_id='', item_id='')
-        return self.add_to_queue(item)
+        return self.add_to_queue(item, position, as_next)
 
     @only_on_master
-    def add_to_queue(self, queueable_item):
-        """Adds a queueable item to the queue."""
+    def add_to_queue(self, queueable_item, position=0, as_next=False):
+        """Adds a queueable item to the queue.
+        For Arguments see `add_uri_to_queue`"""
         metadata = to_didl_string(queueable_item)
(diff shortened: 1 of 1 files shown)


soco/core.py
```diff
@@ -1261,28 +1261,36 @@ def get_sonos_playlists(self, *args, **kwargs):
                                                                 **kwargs)
 
     @only_on_master
-    def add_uri_to_queue(self, uri):
+    def add_uri_to_queue(self, uri, position=0, as_next=False):
         """Adds the URI to the queue.
```
KennethNielsen: While it doesn't pertain to this PR at all, could I convince you to convert the docstring to Napoleon syntax and also to exchange the bulk of the two, so that the main (full) docstring is in add_to_queue and the docstring to add_uri_to_queue then refers to the other one.

KennethNielsen:
I can't seem to get the `as_next` argument to work (even in normal play mode). Both with and without a `position` argument, the `as_next` doesn't do anything for me. I did something like this, with an album already playing in the queue:

```python
import soco

for z in soco.discover():
    if z.player_name == "Stue":
        break

def print_queue(z):
    for item in z.get_queue():
        print(item.title)
    print()

print_queue(z)

t = z.music_library.get_tracks(search_term="Enter")[0]
print(t.title)
print()

z.add_to_queue(t, as_next=True)

print_queue(z)

```

ghcs27:
I just did another test and realized `as_next` works only if shuffle is enabled.
It does not change the queue in any way. Rather, in shuffle mode the player selects the next title in beforehand, and `as_next=True` sets the new item as the next track.
Sonos uses this feature for the 'Play Next' command.
I have updated the docstring to point out that the argument only works in shuffle mode, and converted it to Napoleon syntax.
Please have a look at the changes and check whether, with this clarification, `as_next` also works for you.

soco/core.py
```diff
@@ -1283,7 +1275,15 @@ def add_uri_to_queue(self, uri, position=0, as_next=False):
     @only_on_master
     def add_to_queue(self, queueable_item, position=0, as_next=False):
         """Adds a queueable item to the queue.
-        For Arguments see `add_uri_to_queue`"""
+
+        Args:
+            queueable_item (str): The item to be added to the queue
```
KennethNielsen: Thanks for doing the doc string work as well. Just two detail, the argument type for `queueable_item` is no longer `str` after you moved it. It should now be `DidlObject or MusicServiceItem`.

soco/core.py
```diff
@@ -1283,7 +1275,15 @@ def add_uri_to_queue(self, uri, position=0, as_next=False):
...
+            position (int): The index (1-based) at which the URI should be
+            added.
```
KennethNielsen: I think the "added." was somehow left hanging. It should either be moved to the line above if that doesn't make it too long or indent it one more level to work correctly as Napoleon.

KennethNielsen:
Thanks for clarification on the `as_next` argument. I think a doc string clarification for that will do just fine. I made a new review for a few more nit-picks concerning the doc string updates I coerced you into doing ;) Thanks.

ghcs27:
No problem 😄

KennethNielsen:
Merged. Thanks for your contribution.

Can you please add a comment to the release notes thread #440?

PR #474, opened 2017-02-06
Bugfix wrong internal states on snapshot after some group changes

pvizeli:
I found this bug after I implementation this to Home-Assistant. We use it as member variable in a internal object. But after some change It dosn't work anymore. So I found out, that the internal states going wrong and will not set to defaults after a new snapshot.

### DIFF
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -81,6 +81,10 @@ class Snapshot(object):
         media_info = self.device.avTransport.GetMediaInfo([('InstanceID', 0)])
         self.media_uri = media_info['CurrentURI']
 
+        # reset internal states
+        self.is_coordinator = False
+        self.is_playing_queue = False
+
         # extract source from media uri
         if self.media_uri.split(':')[0] != 'x-rincon':
             self.is_coordinator = True


KennethNielsen:
@robwebset if I (and the git history) remember correctly, you are one of the main persons behind the snapshot module. Can you review this?

robwebset:
Hi @KennethNielsen 

It's been a while since I wrote that. I think the root of the problem is (I think) that the snapshot class is being re-used in this case. When I originally wrote it I saw it as a throw-away class that was created - snapshot, restored, then destroy the class and create a new one when it was next required.

However, I do not see any issue with the changes proposed and would say that it's safe to merge in.

Thanks, Rob

pvizeli:
@robwebset exactly describe our problem. I can change the handle to a throw-away in our code. Don't see that on documentation. My change make it possible to use the same snapshot object over multible snap/restore calls.

robwebset:
@pvizeli No worries - I think we should be OK to merge this one - it makes it better overall :)

DPH:
+1

KennethNielsen:
@robwebset @DPH @pvizeli 

I read through the code for snapshot and this *seems* ok. It is however not completely trivial to convince one-self that this is the only change that is required in order to make the Snapshot object re-usable. I.e. implemented this as a use-and-throw-away is much easier that re-use.

Given the discussion here, I tend to lean towards that what is really missing here is a documentation clarification, especially given that more functionality is not necessarily better if sacrificing simplicity. What do you think @pvizeli?

DPH:
I have this running for my door bell 24X7 (Raspberry Pi Python 2.7) but on each doorbell press I do create a new class, so I use and throw away. If we want it to be reusable then we would need to reset a number of the variables (ie the __init__) to be safe. If a zone was a master on a previous snap and is now a slave I think it would give unpredictable results. 

Should we have classes that can long live, or is it better to use and destroy? Agree we should document but I expect people will expect you can re-use it. 

If we want to make this re-usable we should extend @pvizeli change to include all the  init code into the snapshot function (and then remove from init).
Cheers Daivd

KennethNielsen:
@DPH agreed, except I don't know if people will expect to be able to re-use it if it is documented. Snapshotting is an exotic enough feature that you will have to read the documentation anyway.

Creating new instances seems like a waste, but the actual over head is very small and in any case, completely insignificant compared to the work the class has to anyway each time it (re)-snapshots.

The more I think about it, I think I prefer to document and "keep it simple"™.

DPH:
OK - Lets keep it simple and include some documentation (assume in the code) will include some when I look at #475 issue  issue next week.

pvizeli:
I can change our code to "throw-away" handle. That is not a problem. You can close this PR

robwebset:
I can't say I have a strong opinion about this either way - so will go with whatever the consensus is.

KennethNielsen:
Waiting for doc PR from @DPH and closing this one. Thanks.

PR #478, opened 2017-02-22
Didl class for Sonos favorites

ghcs27:
This PR makes browsing Sonos favorites with the existing music library methods possible.
This is archieved by adding a data structure for the `object.itemobject.item.sonos_favorite` Didl class. The metadata of the favorites contains a `<resMD>`  tag, which encloses the metadata of the item to which the favorite refers. I have added a `reference` property to make this metadata accessible as a Didl object.
Unlike the existing implementation, one can now handle Sonos and radio favorites just like other music items - the specific methods are no longer required.
I would really like to integrate this functionality into SoCo, so please let me know if I should change something to get this merged!

### DIFF
--- a/soco/music_library.py
+++ b/soco/music_library.py
@@ -37,7 +37,8 @@ class MusicLibrary(object):
                           'playlists': 'A:PLAYLISTS',
                           'share': 'S:',
                           'sonos_playlists': 'SQ:',
-                          'categories': 'A:'}
+                          'categories': 'A:',
+                          'sonos_favorites': 'FV:2'}
 
     # pylint: disable=invalid-name, protected-access
     def __init__(self, soco=None):
--- a/soco/music_library.py
+++ b/soco/music_library.py
@@ -142,6 +142,34 @@ class MusicLibrary(object):
         args = tuple(['playlists'] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
+
+    def get_sonos_favorites(self, *args, **kwargs):
+        """Convenience method for `get_music_library_information`
+        with ``search_type='sonos_favorites'``. For details of other arguments,
+        see `that method
+        <#soco.music_library.MusicLibrary.get_music_library_information>`_.
+        """
+        args = tuple(['sonos_favorites'] + list(args))
+        return self.get_music_library_information(*args, **kwargs)
+
+    def get_favorite_radio_stations(self, *args, **kwargs):
+        """Convenience method for `get_music_library_information`
+        with ``search_type='radio_stations'``. For details of other arguments,
+        see `that method
+        <#soco.music_library.MusicLibrary.get_music_library_information>`_.
+        """
+        args = tuple(['radio_stations'] + list(args))
+        return self.get_music_library_information(*args, **kwargs)
+
+    def get_favorite_radio_shows(self, *args, **kwargs):
+        """Convenience method for `get_music_library_information`
+        with ``search_type='radio_stations'``. For details of other arguments,
+        see `that method
+        <#soco.music_library.MusicLibrary.get_music_library_information>`_.
+        """
+        args = tuple(['radio_shows'] + list(args))
+        return self.get_music_library_information(*args, **kwargs)
+
         # pylint: disable=too-many-locals, too-many-arguments,
         # too-many-branches
 
--- a/soco/music_library.py
+++ b/soco/music_library.py
@@ -142,7 +142,6 @@ class MusicLibrary(object):
         args = tuple(['playlists'] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
-
     def get_sonos_favorites(self, *args, **kwargs):
         """Convenience method for `get_music_library_information`
         with ``search_type='sonos_favorites'``. For details of other arguments,
--- a/soco/music_library.py
+++ b/soco/music_library.py
@@ -38,7 +38,9 @@ class MusicLibrary(object):
                           'share': 'S:',
                           'sonos_playlists': 'SQ:',
                           'categories': 'A:',
-                          'sonos_favorites': 'FV:2'}
+                          'sonos_favorites': 'FV:2',
+                          'radio_stations': 'R:0/0',
+                          'radio_shows': 'R:0/1'}
 
     # pylint: disable=invalid-name, protected-access
     def __init__(self, soco=None):
--- a/pylintrc
+++ b/pylintrc
@@ -6,7 +6,7 @@
 # duplicate code check disabled whilst refactoring of wimp plugin is in
 # progress
 disable=too-many-lines,locally-disabled,duplicate-code,too-few-public-methods,
-    bad-option-value,no-else-return
+    bad-option-value,no-else-return,cyclic-import
 
 
 [REPORTS]
--- a/soco/core.py
+++ b/soco/core.py
@@ -142,9 +142,6 @@ class SoCo(_SocoSingletonBase):
         add_multiple_to_queue
         remove_from_queue
         clear_queue
-        get_favorite_radio_shows
-        get_favorite_radio_stations
-        get_sonos_favorites
         create_sonos_playlist
         create_sonos_playlist_from_queue
         remove_sonos_playlist
@@ -1450,6 +1447,7 @@ class SoCo(_SocoSingletonBase):
             ('InstanceID', 0),
         ])
 
+    @deprecated('0.13', "soco.music_library.get_favorite_radio_shows", '0.15')
     def get_favorite_radio_shows(self, start=0, max_items=100):
         """Get favorite radio shows from Sonos' Radio app.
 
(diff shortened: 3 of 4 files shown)


ghcs27:
Can someone please help me with this complaint:
`soco/data_structures.py:772:15: F821 undefined name 'from_didl_string'`
The code seems to work fine, so I don't understand what this is about.

KennethNielsen:
Hi @ghcs27 

This sounds real interesting and I will review this next.

ghcs27:
I just figured out where `soco/data_structures.py:772:15: F821 undefined name 'from_didl_string'` comes from: `from_didl_string()` was moved to data_structures_entry.py.
However, I cannot just do
```python
from .data_structures_entry import from_didl_string
```
because that would create a cyclic import. Could someone please help me to resolve this?

ghcs27:
I just figured out where `soco/data_structures.py:772:15: F821 undefined name 'from_didl_string'` comes from: `from_didl_string()` was moved to data_structures_entry.py.
However, I cannot just do
```python
from .data_structures_entry import from_didl_string
```
because that would create a cyclic import. Could someone please help me to resolve this?

KennethNielsen:
Hi @ghcs27 

Sorry about the long time away. Busy with everything else. To try and make sure it wont be as long again, I will ask you to do a little more work on this one (like explaining stuff to me) instead of trying to figure it out on my own.

Moving `from_didl_string` in datastructures_entry.py was done precisely to avoid **other** circular imports. So in general the idea is this.

 1. Code (including music library) should in general rarely have to instantiate stuff from data_structures directly, but instead use `from_didl_string` in `data_structures_entry.py` (hence the name).
 2. `from_didl_string` is in charge of figuring out the correct class for the item, and then to instantiate the datastructure (directly from XML) by calling `from_element` on it with the XML as argument
 3. The `from_element` method on DildObject (which is inherited) extracts all the useful information from the XML and calls __init__ on itself with it.

This pattern is desinged for the patterns that every other datastructure seems to follow, namely that the returned XML is (at most) a list of DidlObjects. However, for these favorites things seem to be different, since they have a reference to a DidlObject. A few questions:

 1. Are these favorites playable themselves?
 2. Are the references?
 3. Can you describe the role of the two as you understand it and give an example on the maetadata being held in both?

Regards Kenneth

ghcs27:
I am currently trying to understand the music services code myself so I can hopefully restructure it to avoid circular imports.

**Answers:**
1. Yes
2. The xml in the `<resMD>` tag is not, but my code makes it playable by adding the `<res>` from the favorite.
3. - The favorite contains the information you need when browsing favorites, like given title (the one that can be changed in the official controller), a description (mouseover in the controller) and its index. It also contains the `<res>` tag.
   - The reference contains the metadata of the library item, but not the `<res>` tag. This includes the title as specified in the mp3.

**Some example XML:**

Favorite:
```XML
<item id="FV:2/53" parentID="FV:2" restricted="false">
    <dc:title>Test Favorite</dc:title>
    <upnp:class>object.itemobject.item.sonos-favorite</upnp:class>
    <r:ordinal>10</r:ordinal>
    <res protocolInfo="x-rincon-playlist:*:*:*">x-rincon-playlist:RINCON_000E5875680601400#A:ALBUMARTIST/Adele</res>
    <r:type>instantPlay</r:type>
    <r:description>Interpret</r:description>
    <r:resMD>[...]</r:resMD>
</item>
```
Reference:
```XML
<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/">
    <item id="A:ALBUMARTIST/Adele" parentID="A:ALBUMARTIST" restricted="true">
        <dc:title>Adele</dc:title>
        <upnp:class>object.container.person.musicArtist</upnp:class>
        <desc id="cdudn" nameSpace="urn:schemas-rinconnetworks-com:metadata-1-0/">RINCON_AssociatedZPUDN</desc>
    </item>
</DIDL-Lite>
```
For comparison, the referenced item in the music library:
```XML
<container id="A:ALBUMARTIST/Adele" parentID="A:ALBUMARTIST" restricted="true">
    <dc:title>Adele</dc:title>
    <upnp:class>object.container.person.musicArtist</upnp:class>
    <res protocolInfo="x-rincon-playlist:*:*:*">x-rincon-playlist:RINCON_000E5855B28401400#A:ALBUMARTIST/Adele</res>
</container>
```

ghcs27:
What is the relationship of ms_data_structures.py and music_services\data_structures.py?

KennethNielsen:
Please don't re-factor that at will. As I mentioned, the current structure is a result of a refactoring to avoid circular imports. I think I have an idea of how to fix it. I will write about it soon.

KennethNielsen:
Ok. So first things first. ms_data_structures.py is a obsolete version of music_services/data_structures.py. It should have been marked as obsolete a while ago, but I can see that it isn't.

KennethNielsen:
@ghcs27 what happens when you browse the favorite? Do you then get the normal music library items back, as you would have if you had browsed the music library item?

ghcs27:
You cannot browse the favorite. `player.music_library.browse(favorite)` returns an empty SearchResult.

soco/music_library.py
```diff
@@ -139,6 +142,33 @@ def get_playlists(self, *args, **kwargs):
         args = tuple(['playlists'] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
+    def get_sonos_favorites(self, *args, **kwargs):
+        """Convenience method for `get_music_library_information`
```
KennethNielsen: I know the other docstrings in this class are broken too, but can you fix up the new ones to follow the docstring convention of:
```python
def myfunction():
    """Short preferably one-line description

    Longer description

    Args (not applicable here)
    """
```

it may even have been me that wrote them that way to begin with, anyway, we should get them fixed up.

KennethNielsen:
Finally got around to reviewing this. I think it looks good. I thinks I found a solution for the circular import problem, mentioned in the review. A thought struck me while reading through this and testing it: Normally, anything that you get out of the MusicLibrary class is browsable, such as they are, but for these favorites, we have to go via the reference and browse those. Did you consider hacking modifying browse, to account for this. Something along the lines of, of ml_item is instance of favorite then replace ml_item with ml_item.reference. Would that make sense?

ghcs27:
Just tried `import .data_structures_entry` instead of `from .data_structures_entry import from_didl_string`,
but pylint still complains about the cyclic import. A solution that feels a bit more like a hack would be injecting `from_didl_string` into data_structures.py like this:
```python
# data_structures_entry.py
import .data_structures

def from_didl_string(string):
    ...

data_structures.from_didl_string = from_didl_string
```

```python
# data_structures.py
# no data_structures_entry import

from_didl_string = None

class DidlFavorite(DidlItem):
    do_something_with(from_didl_string)
```

ghcs27:
Got the import fixed, but it still requires turning off cyclic import checking (globally!).
Please also have a look at [this refactoring](https://github.com/ghcs27/SoCo/commit/d3829d8d22b46f88bbf35f8a93244930cdc63f02) which tries to make the import structure more elegant, by using the trick described above (turns out it is actually a valid code pattern, if applied carefully).

KennethNielsen:
I must admit I don't care much for injecting objects into modules, as I don't imagine it will much good for readability. I have looked a little around and I think the pylint error is a bug, or, the circular relative imports aren't supported in earlier python 3 versions. Could you try and disable the warning globally, so we can see whether the tests run on all python version?

ghcs27:
Hmm... it's a python error if I do `from . import data_structures_entry`.

ghcs27:
Neither does `import soco.data_structures_entry as data_structures_entry`.

ghcs27:
Nor just `import soco.data_structures_entry`.
It might be that we can only resolve the cyclic import problem by changing all imports along the cycle
(or refactor so we don't need to).

KennethNielsen:
As discussed on IRC. For now, I suggest you formalize you original implementation, so that the import is saved in a module level variable, which is initially set to None, then in the code that use it, check whether the function is imported and if not import it and save it to the global variable. It is the less intrusive way to get this merged soon.

If this can be done without triggering the cyclic import pylint warning, then disabling that warning should be removed, if not, then it should be left in and a separate issue should be opened to fix the code better.

ghcs27:
Sorry, I forgot.

KennethNielsen:
No problem.

ghcs27:
Apparently pylint would like uppercase naming for the module level variable, and doesn't like the global statement:
```
C: 44, 0: Constant name "_from_didl_string" doesn't conform to UPPER_CASE naming style (invalid-name)
W:822, 8: Using the global statement (global-statement)
C:822, 8: Constant name "_from_didl_string" doesn't conform to UPPER_CASE naming style (invalid-name)
```
Should I rename it? Is there an alternative to `global`? Or should I just disable the warnings?

KennethNielsen:
Yes, rename to _FROM_DIDL_STRING_FUNCTION and disable the global warning on the line where you use it

ghcs27:
OK, *now* it's done.
pylint only complains about files I didn't touch.

KennethNielsen:
It seems the failing builds are due to a new pylint version on TravisCI. I'll merge this now and look into the pylint warnings afterwards. Thanks for the contributions. Remember to write and entry in the release notes issue #551.

Issue #472, opened 2017-01-14
Shortcut for DidlObject.resources[0].uri


ghcs27:
I have noticed that I use `DidlObject.resources[0].uri` quite often and it is also present in the SoCo code multiple times. Hardcoding the index `[0]` in so many places does not seem very good practice to me, I would propose adding a `uri` property to DidlObject which returns `resources[0].uri`:

If, for instance, we were to encounter metadata with two `<res>` tags of which only the second one contains the URI, code at many different places would have to be updated. With an adaptable property, this wouldn't be a problem.

What do you think?

KennethNielsen:
Sounds like a good idea. Do you want to write it up? If you do, make sure there is no name clash for uri.

ghcs27:
While writing the property, I came across the question: Should it have a setter?
Setting the uri of a DidlObject which does not yet have a resource object would create a new `DidlResource` with a generic protocol info.

ghcs27:
The line `DidlResource(uri=[uri], protocol_info="[uri_prefix]:*:*:*")` is quite frequent. I reckon that `[uri_prefix]:*:*:*` should be the default protocol info, which would make a setter for `DidlObject.uri` simpler.

KennethNielsen:
I think, let's see if a need arises for a setter. At least that is a lot more rare, so maybe it is ok if people will have to manipulate the `.resources` directly for that.

ghcs27:
Well, I just had a look at my own code:
```python
res = [DidlResource(uri=uri, protocol_info="x-rincon-playlist:*:*:*")]
item = DidlObject(resources=res, title=title, parent_id='', item_id='')
```
isn't as intuitive or readable as
```python
item = DidlObject(title=title, parent_id='', item_id='')
item.uri = uri
```
or even better
```python
item = DidlObject(title=title, parent_id='', item_id='', uri=uri)
```
The first code block even crops up in the SoCo code: see add_uri_to_queue() and create_sonos_playlist.

ghcs27:
@KennethNielsen Drafts of my changes are at

- [`ghcs27:didl_object_uri`](https://github.com/ghcs27/SoCo/tree/didl_object_uri) without setter or protocol info default
- [`ghcs27:didl_object_uri_with_setter`](https://github.com/ghcs27/SoCo/tree/didl_object_uri) including setter and protocol info default

Can you please take a look at them?

[...]

PR #479, opened 2017-02-22
Property for DidlObject.resources[0].uri

ghcs27:
I have noticed that I use DidlObject.resources[0].uri quite often and it is also present in the SoCo code multiple times. Hardcoding the index [0] in so many places does not seem very good practice to me.

Therefore, I have added a uri property to DidlObject which returns resources[0].uri.

> Note: This was discussed before in #472.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -692,7 +692,12 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
 
     @property
     def uri(self):
+        """The uri to use for playing this item. This is currently the uri of
+        the first resource."""
         return self.resources[0].uri
+        # According to the spec, a DidlObject can have multiple resources and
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Hi @ghcs27 

The more I think about this, the more I would prefer it that it was a method. As I ended with in the other thread, there is no rule that things must be properties. We just make them that, if it is appropriate, because they get and set something simple.

What do you think about a get and set with this def:
```python
def get_uri(resource=0):
    pass
def set_uri(resource=0, protocol_info="x-rincon-playlist:*:*:*"):
    pass
```
Wiht that structure I would merge it in a heart beat.

ghcs27:
That isn't a bad idea actually, so I have opened #482 for `get_uri` and `set_uri`.

PR #484, opened 2017-03-13
Added a few queue micro examples

KennethNielsen:
A have added a few queue related examples to the micro examples page.

Let me know what you think.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -1021,8 +1021,9 @@ class ListOfMusicInfoItems(list):
     the total number of results to the query and :attr:`~number_returned` is
     the number of results actually returned. If the two differ, paging is
     required. Paging is typically performed with the ``start`` and
-    ``max_items`` arguments to the query method.
-
+    ``max_items`` arguments to the query method. See e.g. the
+    :meth:`~soco.music_library.MusicLibrary.get_music_library_information`
+    method for details.
     """
 
     def __init__(self, items, number_returned, total_matches, update_id):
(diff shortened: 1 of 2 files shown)


ghcs27:
I think it is very good to expand the examples page. I like the examples you added, but the examples page doesn't seem to me to be the best place for describing `total_matches` and `number_returned`. Does this already exist somewhere? If not, I'd suggest moving your description to the `ListOfMusicInfoItems` docstring and linking from the example page to there.

KennethNielsen:
@ghcs27 no, I think I remember thinking that it didn't feel correctly placed there, but I could quite figure out what to do with it. I think your suggestion of moving it to the ListOfMusicInfoItems docstring makes sense. I'll make the change soon. Let me know if there are any comments for grammar, spelling or formulations.

KennethNielsen:
How about that? You think that is an appropriate level of the example?

ghcs27:
Looks great 👍

soco/data_structures.py
```diff
@@ -1012,7 +1012,18 @@ class DidlMusicGenre(DidlGenre):
...
+    """Abstract container class for a list of music information items.
+
+    Instances of this class are returned from queries into the music library
+    or to music services. The attributes :attr:`~total_matches` and
+    :attr:`~number_returned` are used to ascertain whether paging is required
+    in order to retrive all elements of the query. :attr:`~total_matches` is
+    the total number of results to the query and :attr:`~number_returned` is
+    the number of results actually returned. If the two differ, paging is
+    required. Paging is typically performed with the ``start`` and
+    ``max_items`` arguments to the query method.
```
ghcs27: Just one thing: A link to `music_library.get_music_library_information` would be good.

KennethNielsen: Agreed. Thanks.

PR #487, opened 2017-03-19
Added by_name function to discovery, with test and doc changes

KennethNielsen:
This PR adds a very simple `by_name` function to discovery that returns a device by its name.

```python
from soco.discovery import by_name
device = by_name("Kitchen")
```

I don't know why I haven't done this earlier, since the code that it contains, I have been writing by hand a lot of time while testing different things.

Let me know what you think.

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -49,18 +49,14 @@ one e.g. to query for music library information::
 Getting a named device
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Getting a device by name is done by searching through the devices
-returned by :func:`soco.discover`::
+Getting a device by player name can be done with the
+:func:`soco.discovery.by_name` function::
 
-  >>> import soco
-  >>> for device in soco.discover():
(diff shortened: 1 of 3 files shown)


ghcs27:
So simple and yet so ingenious. 😶

KennethNielsen:
@ghcs27 Thanks.

Can I convince you to do a review? I already spotted a singular/plural "..If none is/are found.." error, that I will correct.

ghcs27:
- Re "none is/are found": Actually, it seems that even native English speakers couldn't agree on whether "is" or "are" would be correct in this context [[link]](https://motivatedgrammar.wordpress.com/2009/01/08/none-is-none-are-grammar-according-to-clarkson/), so it's probably fine either way.
- Re by_name(): Tested, works.
- Re docs: Grammar/spelling fine, maybe add a full stop at the end of the docstring "sentences"?

KennethNielsen:
@ghcs27 Thanks. Merged.

Issue #485, opened 2017-03-13
add_multiple_to_queue fails with too many items (>16)


KennethNielsen:
While answering #223 I was playing around with the new add_multiple_to_queue method and I noticed that it kept failing after trying a few times to add a queue to the queue. After a little bit of investigation it appears to be related (somehow) to the number of items added.

To reproduce run the follow:

```python
import soco

for device in soco.discover():
    if device.player_name == "Stue":
        break

device.clear_queue()
print(device)

for n in range(1, 100):
    print(n)
    device.clear_queue()
    sandman = device.music_library.get_tracks(search_term="Enter Sandman")[0]
    items = [sandman] * n
    device.add_multiple_to_queue(items)
```
(while changing the zone and track name to something that works at your place).

On my system it fails when n hits 17. My initial thought was that the metadata got too big to be contained in whatever data type the Sonos strings has, so I printed out the size of the uris and metadata strings while testing:

```text
<SoCo object at ip 192.168.1.18>
1
Metadata 917 uris 93
2
Metadata 1835 uris 187
3
Metadata 2753 uris 281
4
Metadata 3671 uris 375
5
Metadata 4589 uris 469
6
Metadata 5507 uris 563
7
Metadata 6425 uris 657
8
Metadata 7343 uris 751
9
Metadata 8261 uris 845
10
Metadata 9179 uris 939
11
Metadata 10097 uris 1033
12
Metadata 11015 uris 1127
13
Metadata 11933 uris 1221
14
Metadata 12851 uris 1315
15
Metadata 13769 uris 1409
16
Metadata 14687 uris 1503
17
Metadata 15605 uris 1597
Traceback (most recent call last):
  File "test_multiple.py", line 15, in <module>
    device.add_multiple_to_queue(items)
  File "/home/kenni/venv/soco/SoCo/soco/core.py", line 1401, in add_multiple_to_queue
    ('EnqueueAsNext', 0)
  File "/home/kenni/venv/soco/SoCo/soco/services.py", line 181, in _dispatcher
    return self.send_command(action, *args, **kwargs)
  File "/home/kenni/venv/soco/SoCo/soco/services.py", line 408, in send_command
    self.handle_upnp_error(response.text)
  File "/home/kenni/venv/soco/SoCo/soco/services.py", line 469, in handle_upnp_error
    error_xml=xml_error
soco.exceptions.SoCoUPnPException: UPnP Error 402 received: Invalid Args from 192.168.1.18
```

[...]

PR #488, opened 2017-03-20
Fix add_multiple_to_queue fails with too many items

ghcs27:
This is a fix for #485.
`add_multiple_to_queue()` now breaks the list into chunks of 16 items and makes separate requests to add each one to the queue.
Multiple requests, however, means that we cannot return the 'index at which the items were enqueued' anymore, so I have removed the return value (I cannot see much value in knowing the position at which the first of many items was enqueued anyway).

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1388,31 +1388,31 @@ class SoCo(_SocoSingletonBase):
             items (list): A sequence of items to the be added to the queue
             container (DidlObject, optional): A container object which
                 includes the items.
-
-        Returns:
-            int: The index at which the items were enqueued.
         """
-        uris = ' '.join([item.resources[0].uri for item in items])
-        uri_metadata = ' '.join([to_didl_string(item) for item in items])
         if container is not None:
             container_uri = container.resources[0].uri
             container_metadata = to_didl_string(container)
         else:
-            container_uri = ''
+            container_uri = ''  # Sonos seems to accept this as well
             container_metadata = ''  # pylint: disable=redefined-variable-type
-        response = self.avTransport.AddMultipleURIsToQueue([
-            ('InstanceID', 0),
-            ('UpdateID', 0),
-            ('NumberOfURIs', len(items)),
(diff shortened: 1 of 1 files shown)


soco/core.py
```diff
@@ -1388,31 +1388,28 @@ def add_multiple_to_queue(self, items, container=None):
             items (list): A sequence of items to the be added to the queue
```
KennethNielsen: So I think I suggested some of the formulation that describes `items`. The reason I suggested "sequence" is that the original version of this code really only required a sequence, since the only that you were doing with it was to iterate over it. This meant e.g. that it would have worked fine with a set. But now you also need indexing (slicing really), which means that you should probably either change this formulation to explicitly say "list" or change the code to turn items into a list when it enters.

soco/core.py
```diff
@@ -1388,31 +1388,28 @@ def add_multiple_to_queue(self, items, container=None):
...
-            ('EnqueuedURIs', uris),
-            ('EnqueuedURIsMetaData', uri_metadata),
-            ('ContainerURI', container_uri),
-            ('ContainerMetaData', container_metadata),
-            ('DesiredFirstTrackNumberEnqueued', 0),
-            ('EnqueueAsNext', 0)
-        ])
-        qnumber = response['FirstTrackNumberEnqueued']
-        return int(qnumber)
+        csize = 16
```
KennethNielsen: So in the nitpicking category.

We usually prefer to write variables out, unless they are only used on the same line. The loop expression maybe also got a bit complex. How about:
```python
chunk_size = 16
for index in range(0, len(items), chunk_size):
    chunk = items[index, index + chunk_size]

KennethNielsen:
Hi @ghcs27 

Thanks for nailing this down. I have just a few comments/requests for changes. Let me know what you think.

Regards Kenneth

ghcs27:
Fixed and added some code comments.

KennethNielsen:
Great. Thanks for the contribution. Merged.

PR #489, opened 2017-03-21
Added one use documentation and fixed Python 3 portnetial issue

### LINKED ISSUE #475: snapshot code confusion over self.playlist_position
While reading through the snapshot code I stumbled on this:

```python
            if self.is_playing_queue and self.playlist_position > 0:
                # was playing from playlist

                if self.playlist_position is not None:
```

which I saw as a potential Python 3 bug. Since, if `self.playlist_position` can be `None`, `self.playlist_position > 0` will fail on Python 3 (unorderable types), while it was allowed under Python2.

When you then look at how the value is retrieved it goes:

```python
                position = track_info['playlist_position']
                if position != "":
                    # save as integer
                    self.playlist_position = int(position)
```

So it is indeed possible, that the property will not be changed away from init value of None.

Maybe the order of the checks in the first code snippet should be changed. This should be investigated.

DPH:
Second attempt at pull request to fix #475 

This fixes a python 3 potential issue, and adds documentation to snapshot to explain it should only be used once as on re-use variables are not initialised.

### DIFF
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -4,7 +4,12 @@
 """Functionality to support saving and restoring the current Sonos state.
 
 This is useful for scenarios such as when you want to switch to radio
-and then back again to what was playing previously.
+or an announcement and then back again to what was playing previously.
+
+Warning:
+    This class is designed to be created used and destroyed. It is not
+    designed to be reused or long lived. The init sets up defaults for
+    one use.
 """
 
 
@@ -55,7 +60,7 @@ class Snapshot(object):
         # For coordinator zone playing from Queue:
         self.play_mode = None
         self.cross_fade = None
-        self.playlist_position = None
+        self.playlist_position = 0
         self.track_position = None
 
         # For coordinator zone playing a Stream:
--- a/soco/__init__.py
+++ b/soco/__init__.py
@@ -7,6 +7,7 @@
 # unicode_literals
 # https://github.com/SoCo/SoCo/issues/98
 #
+# Davids test
 
 import logging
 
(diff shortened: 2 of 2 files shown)


soco/snapshot.py
```diff
@@ -4,7 +4,12 @@
 """Functionality to support saving and restoring the current Sonos state.
 
 This is useful for scenarios such as when you want to switch to radio
-and then back again to what was playing previously.
+or an announcement and then back again to what was playing previously.
+
+Warning:
+    This class is deigned to be created used and destroyed. It is not
```
KennethNielsen: deSigned
created, used and destroyed      I think

KennethNielsen:
Thanks for writing this up. Just a few minor comments.

DPH:
Updated as requested - I should be able to spell in English!!

DPH:
We could add a flag that raised an error if anyone tried to re-use - or is this being over cautious!

ghcs27:
IMO, raising a ValueError on invalid state is better than giving wrong results if the user disregards the documentation.

KennethNielsen:
@DPH @ghcs27 I tend to think that is being over cautious. I don't we should block it, Python is typically referred to as the "consenting adults" language, without locked doors. Of course checking for inconsistent state while trying to write the snap shot back could be done, but would be a bit of extra work. I think we should let it become a problem before doing that work. In any case it should be a different PR, so I think I'll pull this one as it is.

KennethNielsen:
Merged. @DPH can you add a comment about the change in the release notes issue #440.

Issue #457, opened 2016-12-27
The readme example is not working for me


ersiko:
I'm following the example but after I follow the steps, my speaker doesn't play anything
```
Python 3.5.2+ (default, Sep 22 2016, 12:18:14) 
[GCC 6.2.0 20160927] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import soco
>>> speakers = soco.discover()
>>> speaker = speakers.pop()
>>> speaker
SoCo("10.10.11.104")
>>> speaker.play_uri('http://archive.org/download/TenD2005-07-16.flac16/TenD2005-07-16t10Wonderboy_64kb.mp3')
>>> track = speaker.get_current_track_info()
>>> track['duration']
'0:00:00'
>>> track
{'duration': '0:00:00', 'uri': 'http://archive.org/download/TenD2005-07-16.flac16/TenD2005-07-16t10Wonderboy_64kb.mp3', 'album': '', 'artist': '', 'position': '0:00:00', 'metadata': '<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/"><item id="-1" parentID="-1" restricted="true"><res protocolInfo="http-get:*:application/octet-stream:*">http://archive.org/download/TenD2005-07-16.flac16/TenD2005-07-16t10Wonderboy_64kb.mp3</res><r:streamContent></r:streamContent><dc:title>TenD2005-07-16t10Wonderboy_64kb.mp3</dc:title><upnp:class>object.item</upnp:class></item></DIDL-Lite>', 'title': 'TenD2005-07-16t10Wonderboy_64kb.mp3', 'album_art': '', 'playlist_position': '1'}
>>> 

```

It says "duration: 0". And when I try to play on the mobile app it says "Unable to play 'TenD2005-07-16.flac16/TenD2005-07-16t10Wonderboy_64kb.mp3', the file can not be found".

The file is available, I can play it on my browser... Am I doing something wrong?

ghcs27:
[...]

PR #491, opened 2017-03-29
allow choice of how to play streams in play_uri

DPH:
#457 fixes bug, and enhances play_uri:
 - changed code
 - added tests for the new code
 - provided detailed documentation in the doc string about playing streams. Note I have updated this from   the comments in #457 based on the terminology used in the Sonos API guide.

This gives the user control to force a stream to be played as a radio stream when it defaults to a track stream. Some streams will error if forced, and some must be forced, so the user needs to know what they are doing. 

The release code helps the user in certain circumstances by forcing http:// &  https:// to play as radio. This may or may not be what the user wants depending on the stream.  This help will now be removed so people may initially hit a few problems with current code, until they understand they have the choice on how it is played. 

The new approach relies on informed users and gives them choices.  The only alternative to make the code fully backward compatible would be to keep the current translation and have another option to omit it.

Welcome views.
Regards Daivd

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -450,54 +450,57 @@ class SoCo(_SocoSingletonBase):
     @only_on_master
     def play_uri(self, uri='', meta='', title='', start=True,
                  force_radio=False):
-        """Play a given stream. Pauses the queue. If there is no metadata
-        passed in and there is a title set then a metadata object will be
-        created. This is often the case if you have a custom stream, it will
-        need at least the title in the metadata in order to play.
-        
-        On the Sonos controller music is shown with one of the following display
-        formats and controls:
-        - Radio format: shows the name of the radio station plus other available
-                        data. No seek, next, previous, or voting capability. 
-                        Examples: TuneIn, radioPup
-        - Smart Radio:  shows track name, artist, and album. Limited seek, next
-                        and sometimes voting depending on Music Service. 
-                        Examples: Amazon Prime Stations, Pandora Radio Stations.
-        - Track format: shows track name, artist, and album the same as play 
-                        from queue. Full seek, next and previous capabilities. 
-                        Examples: Spotify, Napster, Rhapsody.
+        """Play a URI
 
-        How it is displayed is determined by the URI prefix:
-        x-sonosapi-stream, x-sonosapi-radio, x-rincon-mp3radio, hls-radio
-        default to radio or smart radio format depending on the stream.
-        Others default to track format x-file-cifs, aac, http, https, 
-        x-sonos-spotify (used bySpotify), x-sonosapi-hls-static (Amazon Prime),
-        x-sonos-http (Google Play & Napster).
-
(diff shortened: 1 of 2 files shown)


KennethNielsen:
The travis continuous integration and test fails. You seem to have introduced a few lint warnings (mostly trailing whitespace). They should be easy to fix. You can run the linter (pylint) by running `make link` at the repo base, to check for code style locally. Let me know if you need help with it.

DPH:
@KennethNielsen think I have covered all comments. Have local sphinx running now - makes it much easier to see impact of changes. The `... ` notation adds warnings to the generator, however the ``...`` notation shows up in red and is not used elsewhere in soco but removes the warnings!

Pylint is built into pyCharm which I use on windows, and doesn't report any errors. Welcome advice here.

Cheers David

DPH:
Run pylint outside of Pycharm and fixed trailing spaces and lines over 79 characters. 
Is our standard 79 characters per line? 
Cheers David

ghcs27:
The new `test_soco_play_uri_force_radio` test is failing.
I tested the method and didn't find any unexpected behaviour, so it's probably a fault in the test.

DPH:
Hi @ghcs27 ,
Sorry - it worked when I did the PR - will look at it and check. 
The PR is changes to the core and the tests. 
Cheers David

DPH:
Hi @ghcs27, tested again myself and it works on my system. 
Not sure what is wrong if it doesn't work for you - welcome your advice
Cheers David

KennethNielsen:
@DPH The changes look good. The Travis check still fail, but that is due to the new version of pylint. This has been fixed in master. If you feel like rebasing it, feel free, otherwise, don't worry about it.

We do need to figure out the test though.

KennethNielsen:
@ghcs27 can you post Python version, platform and the output of pytest, please?

ghcs27:
Sorry - I should've said that the tests only fail on Python 2.6.
Actually I didn't run pytest myself, but looked at the [Python 2.6 Travis build](https://travis-ci.org/SoCo/SoCo/jobs/222544423).

It fails on the line
`uri = 'x-rincon-mp3radio{}'.format(uri[colon:])`
With the message
`ValueError: zero length field name in format`

It will probably pass on all versions when changed to
`uri = 'x-rincon-mp3radio{0}'.format(uri[colon:])`
using an explicit positional index.

KennethNielsen:
@DPH There is still a tiny pylint warning left. You can view the result of the Travis build by clicking the red cross next to the last commit. It seems to be complaining about the number of arguments, which is something we would usually disable if we are comfortable with it.

DPH:
All checks passed now!
I just suppressed the pylint - too many args as I don't see we can get rid of any, and it doesn't seem to make sense to merge any up.
Cheers David

KennethNielsen:
Looks good. Thanks for the contribution. Remember to add a comment about this addition to the release notes issue #440

PR #493, opened 2017-04-05
Documnetation: snapshot FAQ and examples

DPH:
A documentation pull request as discussed - an area with frequent question.
Please check I have changed the faq.rst correctly - not used that before.
Added 2 examples of using snapshot - tested on python 2.7 and 3.5 - feel free to recommend changes.
Cheers David

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -50,11 +50,10 @@ How can I save, then restore the previous playing Sonos state ?
 
 This is useful for scenarios such as when you want to switch to radio,
 an announcement or doorbell sound and then back to what was playing previously.
-Documentation of the Snapshot module `~soco.snapshot.Snapshot`
+Documentation of the Snapshot :mod:`~soco.snapshot` module.
 
 SoCo provides a snapshot module that captures the current state of a player and
 then when requested re-instates that state. Examples of it's use are:
- - `basic snap example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/basic_snap.py>`_
- - `multi zone example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/multi_zone_snap.py>`_
 
-
+ * `basic snap example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/basic_snap.py>`_
+ * `multi zone example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/multi_zone_snap.py>`_
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -44,3 +44,17 @@ two options:
        started, but that no security precautions has been taken to
        e.g. prevent serving other files out into the local
        network. Take appropriate actions if this is a concern.
+
+How can I save, then restore the previous playing Sonos state ?
+---------------------------------------------------------------
+
+This is useful for scenarios such as when you want to switch to radio,
+an announcement or doorbell sound and then back to what was playing previously.
+Documentation of the Snapshot module `~soco.snapshot.Snapshot`
+
+SoCo provides a snapshot module that captures the current state of a player and
+then when requested re-instates that state. Examples of it's use are:
+ - `basic snap example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/basic_snap.py>`_
+ - `multi zone example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/multi_zone_snap.py>`_
+
+
--- a/examples/snapshot/basic_snap.py
+++ b/examples/snapshot/basic_snap.py
@@ -1,0 +1,47 @@
+""" Sample script to demonstrate use of SoCo snapshot
+
+This is useful for scenarios such as when you want to switch to radio,
+an announcement or doorbell sound and then back to what was playing previously.
+
+To use it just change the IP address to one of your Sonos players. 
+
+This script will:
+ - start playing a radio station
+ - take a snapshot of the status of the sonos player (what playing, volume etc.)
+ - play an alert mp3 for a few seconds (a poem!)
+ - re-instate the sonos player to it's previous state (the radio station)
+ 
+Note: the snap function is designed to be used once, Having taken a snap to 
+take another re-instantiate the class.
+"""
+
+import time
+import soco
+from soco.snapshot import Snapshot
+
+# something to play on a Sonos player to start (a radio station)
+start_uri = u'x-sonosapi-stream:s2846?sid=254&amp;flags=32'
+# alert sound to interrupt the above (a poem) - use amy file Sonos can play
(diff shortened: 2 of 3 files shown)


doc/faq.rst
```diff
@@ -44,3 +44,17 @@ two options:
...
+How can I save, then restore the previous playing Sonos state ?
+---------------------------------------------------------------
+
+This is useful for scenarios such as when you want to switch to radio,
+an announcement or doorbell sound and then back to what was playing previously.
+Documentation of the Snapshot module `~soco.snapshot.Snapshot`
+
+SoCo provides a snapshot module that captures the current state of a player and
+then when requested re-instates that state. Examples of it's use are:
+ - `basic snap example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/basic_snap.py>`_
```
KennethNielsen: I think the content looks good.

There is a minor syntax error. You should make a blank line after the colon and don't indent the list and I think use * as lists, so it becomes:

```rst
SoCo provides a snapshot module that captures the current state of a player and
then when requested re-instates that state. Examples of it's use are:

* `basic snap example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/basic_snap.py>`_
* `multi zone example  <https://github.com/SoCo/SoCo/blob/master/examples/snapshot/multi_zone_snap.py>`_
```

If you have sphinx and napoleon installed you can test the compilation by doing "make html" in the doc directory.

examples/snapshot/basic_snap.py
```diff
@@ -0,0 +1,47 @@
...
+This script will:
+ - start playing a radio station
+ - take a snapshot of the status of the sonos player (what playing, volume etc.)
+ - play an alert mp3 for a few seconds (a poem!)
+ - re-instate the sonos player to it's previous state (the radio station)
+ 
+Note: the snap function is designed to be used once, Having taken a snap to 
+take another re-instantiate the class.
+"""
+
```
KennethNielsen: A comment about py2 and 3 compatible code. I guess this was developed on Py3 as you use print as a function. However, the `print()` only work like it is supposed to in Py2, if there is only one argument. If doing `print("more", "strings")` it will print out `("more", "strings")`. The proper way to do it is to start the file with: `from __future__ import print_function`. Then print will actually be a function also in Py2.

While you are at it, you might as well also add `from __future__ import print_function, unicode_literals`. Then you will not need to preface your strings with u to make sure they are unicode.

examples/snapshot/basic_snap.py
```diff
@@ -0,0 +1,47 @@
...
+import time
+import soco
+from soco.snapshot import Snapshot
+
+# something to play on a Sonos player to start (a radio station)
+start_uri = u'x-sonosapi-stream:s2846?sid=254&amp;flags=32'
```
KennethNielsen: Here, remove the u after having imported unicode_literals

examples/snapshot/basic_snap.py
```diff
@@ -0,0 +1,47 @@
...
+
+To use it just change the IP address to one of your Sonos players. 
+
+This script will:
+ - start playing a radio station
+ - take a snapshot of the status of the sonos player (what playing, volume etc.)
+ - play an alert mp3 for a few seconds (a poem!)
+ - re-instate the sonos player to it's previous state (the radio station)
+ 
+Note: the snap function is designed to be used once, Having taken a snap to
```
KennethNielsen: 3 things:

Capital after ":"

Calling if function might confuse since it isn't actually a function, so either change to functionality or say that the "Snapshot instance" is only designed to be used once.

"once, Having", change to full stop or make lowercase. Also the last part if that sentence is maybe a little to compactly formulated. What do you think?

examples/snapshot/multi_zone_snap.py
```diff
@@ -0,0 +1,82 @@
...
+This script has a play_alert function that will:
+ - take a snapshot of the current status of all sonos players (whats playing, 
+   volume etc.)
+ - play an alert_uri file on each group coordinator
+ - re-instate the sonos players to it's previous state
+ 
+This script does not group or un-group any Sonos players. Group management is
+a separate subject. For alerts grouping causes delays, so is avoided here.
+
+Note: the snap function is designed to be used once, Having taken a snap to
```
KennethNielsen: Same as above.

examples/snapshot/multi_zone_snap.py
```diff
@@ -0,0 +1,82 @@
...
+"""
+
+import time
+import soco
+from soco.snapshot import Snapshot
+
+
+def play_alert(alert_uri, alert_volume=20, alert_duration=0, fade_back=False):
+    """
+    Demo function using soco.snapshot accrss multiple Sonos players.
```
KennethNielsen: across

examples/snapshot/multi_zone_snap.py
```diff
@@ -0,0 +1,82 @@
...
+    
+    :param alert_uri: file that Sonos can play as an alert
```
KennethNielsen: Consider using the Napoleon style for sphinc docstrings, which we are trying to switch to in general. Let me know if you need help.

examples/snapshot/multi_zone_snap.py
```diff
@@ -0,0 +1,82 @@
...
+        # For every Sonos player set volume and mute for every zone
+        zone.volume = alert_volume
+        zone.mute = False
+
+    # play the sound (uri) on each sonos coordinator
+    print('will play: {} on all coordinators'.format(alert_uri))
+    for zone in zones:
+        if zone.is_coordinator:
+            zone.play_uri(uri=alert_uri, title='Sonos Alert')
+            a_coordinator = zone  # remember last coordinator for use next
```
KennethNielsen: I can't seem to see this being used.

examples/snapshot/multi_zone_snap.py
```diff
@@ -0,0 +1,82 @@
...
+Note: the snap function is designed to be used once, Having taken a snap to 
+take another re-instantiate the class.
+"""
+
+import time
+import soco
+from soco.snapshot import Snapshot
+
+
+def play_alert(alert_uri, alert_volume=20, alert_duration=0, fade_back=False):
```
KennethNielsen: I'd maybe pass zones in as a parameter. What do you think.

DPH:
Have updated as suggested. Not sure if I have the Napoleon structure right in the multi_zone_snap.py example. Welcome any further comments

Cheers Daivd

KennethNielsen:
Looks good. There are still a lot of pylint warnings, but they seem to be caused by a new version of pylint, which included new checks. I'll see about fixing those tonight.

KennethNielsen:
Merged. Can you add a comment about the changes to the release notes #440?

PR #495, opened 2017-04-13
Add to the git revision number to the version string when in developm…

### LINKED ISSUE #486: Adding "+" to the version in development
I want to add a new convention to our development process. I think we should denote the version in the development archive with whatever version was released last with a "+" suffixed.

That will make it easier to deduce from the version string alone, what kind of installation a user that reports an issue has.

I already changed the string the __init__.py file and unless someone objects wildly I will update the "how to do a release" section of the docs.

KennethNielsen:
…ent.

With more and more issues it has become necessary to know what version of the code (even in development) is being used. This PR adds the git revision (if it can be detected) to the version string.

This change means that when releasing, beside from changing the ``__version__`` in ``__init__.py``, it is also required to set ``_RELEASE`` to ``True``. This has been documented in release howto in the docs.

This closes #486.

I do need one little bit of information from someone on Windows, so shout of if you can help with a test on that platform.

Let me know what you think.

### DIFF
--- a/doc/development/release-howto.rst
+++ b/doc/development/release-howto.rst
@@ -27,7 +27,8 @@ Create and Publish
 * Verify that all tests pass.
 
 * Update the version number in ``__init__.py`` (see
-  `example <https://github.com/SoCo/SoCo/commit/d35171213eabbc4>`_).
+  `example <https://github.com/SoCo/SoCo/commit/d35171213eabbc4>`_)
+  and set _RELEASE to ``True``.
 
 * Tag the current commit, eg
 
@@ -57,6 +58,9 @@ Create and Publish
 
 Wrap-Up
 -------
+
+* In ``__init__.py`` set _RELEASE back to ``False`` and commit and
+  push the change to github.
 
 * Create the milestone for the next release (with the most likely version
   number) and close the milestone for the current release.
--- a/doc/development/release-howto.rst
+++ b/doc/development/release-howto.rst
@@ -27,8 +27,7 @@ Create and Publish
 * Verify that all tests pass.
 
 * Update the version number in ``__init__.py`` (see
-  `example <https://github.com/SoCo/SoCo/commit/d35171213eabbc4>`_)
-  and set _RELEASE to ``True``.
+  `example <https://github.com/SoCo/SoCo/commit/d35171213eabbc4>`_).
 
 * Tag the current commit, eg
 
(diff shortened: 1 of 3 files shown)


stefankoegl:
I like the idea of including the git revision, but somehow it feels odd that each release requires two commits - one for setting ``_RELEASE`` to ``True``, and one for setting it back to ``False``.

What do you think of treating something as a release only if it was prepared according to the release procedures (so essentially setting the flag within the produced package), and treating something as a non-release whenever it is run from source?

DPH:
Agree we need to know what version issues are raised on. 
Ideal is any exception could include the version in the output - not sure if that is possible. 
I can test on Windows - just let me know 
Cheers Daivd

KennethNielsen:
@stefankoegl you have a point. I was pondering this while I wrote it as well. I will make a change to the PR.

KennethNielsen:
@stefankoegl done with a97421895abb37e4817cd36f607f2a699c9685cc, what do you think?

KennethNielsen:
@DPH I'm not sure if the include version in exceptions ideas you mention is possible. Although it could be practical, I fear it would be too invasive.

Can I get you to test something simple for me on Windows. It's actually a git thing more than Python. The code that gets the revision number tries to do it by calling the git command, but if it cannot find it, it falls back to trying to read it out the git files directly. For this I need to know whether ``.git/HEAD`` also uses forward slashes on Windows for folder separation.

So, in your SoCo repository, at the root level of it, there should be a hidden folder called .git, can I get you to paste the content of the HEAD file inside that directory? You must be on an ordinary local branch in order for it to have the information I need.

DPH:
@KennethNielsen, 

The content of HEAD in C:\Users\David\gitSpace\SoCo\.git on Branch force_stream is:
`ref: refs/heads/force_stream`

and on the master branch contents is:
`ref: refs/heads/master`

Let me know if you need more or I have interpreted it incorrectly. 
Cheers David

KennethNielsen:
@DPH that was what I was looking for. Now I feel more comfortable that the fall back will work on Windows as well.

KennethNielsen:
@stefankoegl what do you think about it now?

stefankoegl:
To be honest, I am still not really sure. It somehow seems odd to either release a state that was never committed or commit twice for a single release. Also having an import-time call to a complex function (even if its just for non-release) doesn't really look nice.

However, I did not yet come up with a solution that approves on those points.

I'm -0 on merging at the moment.

KennethNielsen:
@stefankoegl you know what: The more I think about this, the more I think it was better the first time around. I really don't see the issue with having two commits per release. One will be "release" the next will be "development open". I doesn't take up any space in the git archive and the number of release related commits is still negligible.

As far as calling a complex function of import I do agree that is not an idea situation, but it is only when in development. One way to mitigate some of that risk would be to envelop it in a broad try-except, that way, the only thing left to worry about, would be if something hangs, but there are only a few things in that function that potentially could, so I think all-in-all it is worth the risk.

I will do a revision changing back to the original version and make a few changes along the lines of mitigation.

PR #496, opened 2017-04-14
Add `TRANSITIONING` to possible values in `current_transport_state`

pmatos:
During a transition between states sonos can return `TRANSITIONING` for `current_transport_state`. This change is just to document the value.


ghcs27:
The changed line now exceeds the 79 character limit. Could you please insert a line break?

pmatos:
@ghcs27 I have now updated the file.

KennethNielsen:
Looks good. Thank you @pmatos for the doc fix and @ghcs27 for review.

PR #501, opened 2017-04-19
Add Python 3.6 to travis builds

KennethNielsen:
Python 3.6 has been out for a while. Let's see if it builds, and if it does, add it to the Python versions Travis CI runs our builds on.


KennethNielsen:
It all checks out. If I can get someones ack on this I'll merge.

DPH:
Fully support adding Python 3.6.
Have not tested it but if it causes no problems lets include it. 

Cheers David

KennethNielsen:
@DPH well, causes no problems with the unit tests anyway. I think that should be good enough.

KennethNielsen:
With no objections and no problems with unit tests I think it seems ok. I'll close this.

PR #506, opened 2017-04-25
Add ramp_to_volume() method to smoothly change the volume

ghcs27:


DPH:
Hi @ghcs27,
this looks good - I have often thought of adding it, rather than writing it in my own code.

Only suggested addition is a little more documentation to help prevent questions - something like:
```
Smoothly change volume
      
from the current volume to the desired value. Volume can be changed up or down.
The ramp rate is linear at 1.25 steps per second. For example: to change from 
volume 50 to volume 30 would take 16 seconds. This rate is selected by Sonos.
This method is non blocking and has no network overhead once sent.
```
There are other Ramp Types but I don't think it is worth making them available.
`ALARM_RAMP_TYPE` resets to zero, waits about 30secs, then ramps at 2.5 steps per second !
`AUTOPLAY_RAMP_TYPE` resets to zero then ramps very quickly at 50 steps per second.

I would vote to ignore these and leave code functionality as you have it. What do others think?
Cheers David

KennethNielsen:
Looks good.

@ghcs27 I'll second @DPH's comment about a little more explanation of the mechanics to get ahead of questions.

@DPH did you test it?

ghcs27:
Thank you @DPH for the documentation. The other ramp types do not seem very useful, so I'll leave them out.

ghcs27:
I just added a test for the new method, so could you please have a look at that too?

DPH:
Tested both ramp_to_volume function and it's use in snapshot. All work as expected from my point of view.
Looked at the test code added and it looks OK, but I am not skilled in that area (yet!)
With documentation enhanced this looks good to go
+1
Cheers David

ghcs27:
Thanks!

robwebset:
Great addition, I do have one comment - as David has highlighted, there are other values for the RampType.

As they have been identified now, it would be nice to add them to the documentation comment and also have a default argument to the method:

def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE')

Which would allow users to override the default value if they wish to. (Seems a shame not to do that little extra bonus bit that really polishes the method).

Thanks, Rob

ghcs27:
Now with optional `ramp_type` parameter and corresponding documentation.

KennethNielsen:
The tests look good to me. @DPH you reviewed this. You ok with the updates to the rest?

DPH:
Have retested - all works

Sorry @ghcs27 one minor change in the doc string:
`To change from volume 50 to volume 50 would take 16 seconds.`
should read:
`To change from volume **30** to volume 50 would take 16 seconds.`

Also in my tests `AUTOPLAY_RAMP_TYPE` does not wait at all before ramping.

Almost there - a great addition
Cheers Daivd

ghcs27:
Sorry for the docstring slip - corrected now.
It is strange that `AUTOPLAY_RAMP_TYPE` does not wait on your system (it could be that we have different Sonos firmware versions, mine is 5.3) - in any case, I have removed the statement on the 3 second waiting time.
Whether there is a waiting time should not affect any code using SoCo, so I believe it's perfectly okay to leave it out of the docstring.

DPH:
Hi @ghcs27 there may be a small delay in `AUTOPLAY_RAMP_TYPE` but I am not sure if it is the network or within Sonos by design. Mine show version as 7.2 is that the firmware version? If thats the wrong figure how do you get the firmware version?

In terms of the PR all good and tested - ready to go. Thank you.
Cheers David

KennethNielsen:
Looks great. Thanks @ghcs27 for the contribution and @DPH for reviewing. @ghcs27 can you add a comment to the release notes issue #440 please?

Issue #513, opened 2017-06-23
play_from_queue always makes call to self.avTransport.SetAVTransportURI


relevitt:
The play_from_queue function in soco.core always makes a call to self.avTransport.SetAVTransportURI. As the comment in the python code explains, this is to set the queue as the source URI on the device.

In some circumstances this may be a necessary step (e.g. if the device has been set to play from line-in), but they seldom apply. 

The downside of this approach is that every time a user wishes to jump to another track in the queue, the queue is first reinstated on the device. This is an expensive operation which triggers additional events and causes delay. In an application which is monitoring events to keep the user interface updated, these additional events are undesirable. 

For example, there will be an unnecessary transitional event that the queue position has jumped to the first track in the queue. A user interface monitoring these events will jump to the first track before jumping to the track being set by the call to play_from_queue.

As a result, I'm no longer using this soco function, but rather making a direct call to self.avTransport.Seek. This avoids the unnecessary overhead of reinstating the queue on the device.

This solves my problem, but I wonder if soco.core might be updated to address the concern. Could a check be made to see if the queue is already on the device before reloading it, so that reloading occurs only if the queue is not loaded? Or could there be another soco.core function (jump_to_position) which only calls self.avTransport.Seek to set the position? This would be consistent with the previous and next functions, which don't reinstate the queue.

DPH:
Hi @relevitt,
if you are subscribed to the AV events then you know what is currently playing, but without subscription you would have to issue a request to find out if it was playing from the queue, which is probably a similar in impact to just setting it to the queue.

We could split of the seek part and make it callable - say, ```jump_to_track(5, play=True) ``` and also use it in```play_from_queue``` to avoid duplication. It would then be necessary to handle errors if it wasn't already playing from the queue.

Easy to do, adds an additional function to SoCO - what do others think?
Cheers David

ghcs27:
Expect my PR.

PR #514, opened 2017-06-24
Add jump_to_track method

ghcs27:
As proposed in #513. I will gladly make any change that is requested.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -432,20 +432,34 @@ class SoCo(_SocoSingletonBase):
         return int(response['RampTime'])
 
     @only_on_master
+    def jump_to_track(self, index):
+        """ Jump to another track in the queue.
+
+        Arguments:
+            index (int): the index of the track to play; first item in the
+                queue is 0.
+
+        Raises:
+            SoCoUPnPException: UPnP Error 711 if the index is invalid or the
+                queue isn't currently playing.
+        """
+        self.avTransport.Seek([
+            ('InstanceID', 0),
+            ('Unit', 'TRACK_NR'),
+            ('Target', index + 1)
+        ])
+
+    @only_on_master
     def play_from_queue(self, index, start=True):
-        """ Play a track from the queue by index. The index number is
-        required as an argument, where the first index is 0.
+        """ Play a track from the queue by index.
 
-        index: the index of the track to play; first item in the queue is 0
-        start: If the item that has been set should start playing
+        Arguments:
+            index (int): the index of the track to play; first item in the
+                queue is 0.
+            start (bool): If the item that has been set should start playing
 
-        Returns:
-        True if the Sonos speaker successfully started playing the track.
-        False if the track did not start (this may be because it was not
-        requested to start because "start=False")
-
-        Raises SoCoException (or a subclass) upon errors.
-
+        Raises:
+             SoCoUPnPException: UPnP Error 711 if the index is invalid.
         """
         # Grab the speaker's information if we haven't already since we'll need
         # it in the next step.
@@ -461,16 +475,11 @@ class SoCo(_SocoSingletonBase):
         ])
 
         # second, set the track number with a seek command
-        self.avTransport.Seek([
-            ('InstanceID', 0),
-            ('Unit', 'TRACK_NR'),
-            ('Target', index + 1)
-        ])
+        self.jump_to_track(index)
 
         # finally, just play what's set if needed
         if start:
-            return self.play()
-        return False
+            self.play()
 
     @only_on_master
     def play(self):
--- a/soco/core.py
+++ b/soco/core.py
@@ -433,15 +433,24 @@ class SoCo(_SocoSingletonBase):
 
     @only_on_master
     def jump_to_track(self, index):
-        """ Jump to another track in the queue.
+        """Jump to another track in the queue.
+
+        Note:
+            This function can only be used if the player is already set to use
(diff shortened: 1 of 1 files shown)


ghcs27:
The Travis CI build errors on Python 2.6 with
`ERROR: You need Python 2.7 or 3.3+ to install the typing package.`

DPH:
Hi @ghcs27,
tested change and it all works as expected.

In the documentation:  it raises a 701 error if not already playing from the queue, and a 711 error if index out of range. Need to correct this.

I think it would be worth adding a bit of context:
'''This function  can only be used if the player is already set to use the queue. If not use play_from_queue. This is faster than play from queue if already using the queue, as it only make one call rather than 3. Note: if the queue is already playing it will continue to play using the new track after jump_to_track. If paused it will remain paused.
'''
Lets get input from others too
Cheers David

ghcs27:
Thanks for the doc infos! Updated now.

KennethNielsen:
Since @DPH already reviewed this, I will just give this a quick glance. I am annoyed about the Python 2.6 issue, since I prefer not to pull stuff that doesn't pass test. I will check where we are on deprecating Python 2.6.

KennethNielsen:
I think we are moving on removing 2.6 now (#526), so I would ask you to rebase it afterwards

KennethNielsen:
@ghcs27 #526 was merged. Can you rebase against master, so that we can see if the tests pass?

ghcs27:
Done.

soco/core.py
```diff
@@ -432,20 +432,43 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
         return int(response['RampTime'])
 
     @only_on_master
-    def play_from_queue(self, index, start=True):
-        """ Play a track from the queue by index. The index number is
-        required as an argument, where the first index is 0.
+    def jump_to_track(self, index):
+        """ Jump to another track in the queue.
```
KennethNielsen: No space after """ and include a blank line after the first single line and the larger explanation below

soco/core.py
```diff
@@ -432,20 +432,43 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
+            If the queue is already playing it will continue to play using the
+            new track after jump_to_track. If paused it will remain paused.
 
-        Returns:
-        True if the Sonos speaker successfully started playing the track.
-        False if the track did not start (this may be because it was not
-        requested to start because "start=False")
+        Arguments:
+            index (int): The index of the track to play; first item in the
+                queue is 0.
```
KennethNielsen: Usually no full stop if there is only one sentence

soco/core.py
```diff
@@ -432,20 +432,43 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
 
-        Raises SoCoException (or a subclass) upon errors.
+        Raises:
+            SoCoUPnPException: UPnP Error 701 if the queue isn't currently
+                playing.
+                UPnP Error 711 if the index is invalid.
```
KennethNielsen: While I think it is a good idea to separate the error numbers like this, have you checked how it looks when the docs compile. Sphinx sometimes does weird stuff.

soco/core.py
```diff
@@ -432,20 +432,43 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
+            ('Target', index + 1)
+        ])
 
+    @only_on_master
+    def play_from_queue(self, index, start=True):
+        """ Play a track from the queue by index.
+
+        Arguments:
+            index (int): the index of the track to play; first item in the
+                queue is 0.
```
KennethNielsen: As above

soco/core.py
```diff
@@ -432,20 +432,43 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
+            start (bool): If the item that has been set should start playing
+
+        Raises:
+             SoCoUPnPException: UPnP Error 711 if the index is invalid.
```
KennethNielsen: and here

ghcs27:
Sure! But can you please tell me how I can check how my docstrings will look in Sphinx?

ghcs27:
Sure! But can you please tell me how I can check how my docstrings will look in Sphinx?

KennethNielsen:
@ghcs27 of course. You will need to compile the docs locally, so you will need to install sphinx locally. If you develop SoCo in a virtual environment this should be fairly easy. If you install the development requirements:
```bash
pip install -r requirements-dev.txt
```

then enter the doc folder, and then do:
```bash
make html
```
then the docs will be built inside the `_build/html` sub-folder and so you should be able to open `_build/html/index.html` in a browser to view the results.

If you are on Windows you will need to use the make.bat. I don't know the syntax for it, but I suspect it is similar and easily googlable.

Let me know if it doesn't work for you.

ghcs27:
@KennethNielsen thanks for the instructions!

I took care of the docstring (I hope it's ok now).

But before you merge anything, please await another PR for comparison which integrates this functionality into seek(). This will (a) be closer to the implementation by Sonos (they both use the avTransport.Seek() method) and (b) allow you to construct a dict `{track=0, position='0:00:00'}` representing the absolute position and seek to this position with a single call (useful for snapshot et al.).

KennethNielsen:
Closed in favor of #541.

PR #522, opened 2017-10-28
Remove broken downloads badge

hugovk:
See https://github.com/badges/shields/issues/716


stefankoegl:
Good catch - thanks :)

PR #523, opened 2017-10-28
 Drop support for EOL Python 2.6 & 3.3, add support for 3.6

### LINKED ISSUE #325: Drop support for python 2.6
Quoted from the original description in #263

> From #255
> 
> The new versions of pylint and astroid have dropped support for Python 2.6 (and presumably also for python 3.2). At the moment, we have pinned to old versions in requirements-dev.txt. We will have to decide what to do about our own Python 2.6 support going forward. One option may be to have Travis use different versions of pylon etc for Py2.6, but this seems a bit messy.

For reference, please see the lengthy discussion about the issues pertaining to dropping support for Python 2.6 in #263.

hugovk:
Fixes #325.

PS It's great to see a  Music Hack Day hack live beyond the hack day!

---

Edit: Split out to separate PRs to allow more granular decision making:
* [x] https://github.com/SoCo/SoCo/pull/526 Drop 2.6
* [ ] https://github.com/SoCo/SoCo/pull/527 Drop 3.3
* [ ] https://github.com/SoCo/SoCo/pull/528 Add 3.6 and dev

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "2.7"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,11 @@ python:
   - "3.7-dev"
   - "nightly"
 
+matrix:
+  allow_failures:
+    - python: "3.7-dev"
+    - python: "nightly"
+
 cache:
   directories:
     - $HOME/.cache/pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 python:
   - "pypy" # Run slowest first so as not to hold up the build
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: python
 python:
+  - "pypy" # Run slowest first so as not to hold up the build
   - "2.7"
   - "3.3"
   - "3.4"
   - "3.5"
-  - "pypy"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
@@ -17,10 +16,8 @@ install:
   - pip install -e .
 
 script:
-  # flake8 no longer supports Python 2.6, so only run it if not 2.6
-  - if [[ $TRAVIS_PYTHON_VERSION != '2.6' ]]; then flake8 soco; fi
-  # Pylint no longer supports Python 2.6, so only run it if not 2.6
-  - if [[ $TRAVIS_PYTHON_VERSION != '2.6' ]]; then pylint soco; fi
+  - flake8 soco
+  - pylint soco
   - py.test --cov-config .coveragerc --cov soco .
 
 after_script:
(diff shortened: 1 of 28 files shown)


.travis.yml
```diff
@@ -1,11 +1,10 @@
 language: python
 python:
-  - "2.6"
+  - "pypy" # Run slowest first so as not to hold up the build
```
stefankoegl: Travis-CI runs test jobs in parallel, so the order should not matter.

hugovk: It makes a difference if there's more than one build going, as I had with these commits.

If PyPy is the last one to be started, you can be waiting longer to get the result of the whole build.

But I can undo it if it's a problem. :)

stefankoegl: No, it's not a problem. I just disagree on the conclusion, but I agree on the result.

As far as I can tell, jobs are started depending on capacity - in (for a user) unpredictable order, regardless of this sequence.

So in my understanding this order does not matter at all, so there is no reason for me to object to it.

robwebset:
I'm not sure of the merit of deliberately making it not compatible with old Python version - I don't see any plus side - I thing the original was about not worrying for additions moving forward. (Which I do sort of see the point of). Deliberately changing existing code so that it won't work on older python version (without giving any real user feature benefit) seems like the wrong way to go for me.

hugovk:
You get a cleaner codebase, less burden to support old versions which hardly anybody uses, no need to maintain clunky old code, and you can start to use new and modern features of Python. People can install an old version of SoCo if needed.

Here's the pip installs for SoCo from PyPI for the last month, supporting a drop of 2.6 and 3.3:
```
$ pypinfo --percent --pip soco pyversion
python_version percent download_count
-------------- ------- --------------
3.5              37.2%            823
3.6              29.0%            641
2.7              17.0%            377
3.4              15.3%            338
3.3               1.5%             33
2.6               0.0%              1
```

If you're not testing it, there's not much point maintaining it. Look to the future!

Having said that, if you prefer I can update this to retain the old code and just change the Travis config, docs and classifiers.

stefankoegl:
If we drop support for 2.6 we should use this to clean up the codebase.

If we would not do this right now, we would still introduce incompatible changes but would not take the chance to clean up properly.

> Look to the future!

+1

.travis.yml
```diff
@@ -1,11 +1,10 @@
...
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
-  - "pypy"
+  - "3.6"
```
stefankoegl: According to the [Travis docs](https://docs.travis-ci.com/user/languages/python/#Specifying-Python-versions), the following are also supported.

  - "3.5-dev" # 3.5 development branch
  - "3.6-dev" # 3.6 development branch
  - "3.7-dev" # 3.7 development branch
  - "nightly"
  - "pypy3" # Pypy3 2.4.0
  - "pypy-5.3.1"

I would suggest adding all of them to be prepared for upcoming versions.

hugovk: Good idea, see https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/

I've added all except "pypy-5.3.1", as "pypy" is actually now the newer 5.8.0 (the docs are usually a bit out of date; "pypy3" is also 5.8.0).

"3.7-dev" and "nightly" are failing pylint (lots of `Missing module docstring (missing-docstring)`), so they can be `allow_failures` for now.

KennethNielsen:
Has this PR be superseeded by the others?

hugovk:
It has, let's close this one.

PR #528, opened 2017-10-31
Add support for Python 3.6 and dev branches

hugovk:
Split out from PR https://github.com/SoCo/SoCo/pull/523 ("Drop support for EOL Python 2.6 & 3.3, add support for 3.6") to allow more granular decision making.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,11 @@ python:
   - "3.7-dev"
   - "nightly"
 
+matrix:
+  allow_failures:
+    - python: "3.7-dev"
+    - python: "nightly"
+
 cache:
   directories:
     - $HOME/.cache/pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 language: python
 python:
   - "pypy" # Run slowest first so as not to hold up the build
+  - "pypy3"
   - "2.7"
   - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.5-dev"
+  - "3.6-dev"
+  - "3.7-dev"
+  - "nightly"
 
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


KennethNielsen:
@hugovk I few questions. How does the allow failure work. If a commit fails on 3.7-dev, will it report back to Github as a fail or a pass?

hugovk:
Have a look at https://travis-ci.org/SoCo/SoCo/builds/295239453

You can see 3.7-dev and nightly down at the bottom. They're both run, but even if they fail (as they do now), they won't fail the whole build.

See:
* https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/
* https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail
* https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/

"3.7-dev" and "nightly" are failing pylint (lots of `Missing module docstring (missing-docstring)`). At some point, these should be fixed or pylint configured so pylint passes.

KennethNielsen:
@hugovk thanks. I guess this does raise a few questions, like the ones mentioned in that last link. If we allow the failures, will we ever notice that the tests failed? I guess it would be helpful if Travis supported an extra all_passed_badge. On the other hand, as long as the Travis folks are not complaining about the extra test run load (that might not yield much), then I don't see how it could do any harm to add it. If someone sufficiently ambitious about retporting upstream bugs feels like it, we could also add todo bugs for checking the status of these before the development branches are released.

In any case. I think the changes look good. The PR broke, because we merged another part of the mother PR. If you can try and rebase this against master, then this will get a +1 and we can merge.

Let me know if you need help with the rebase.

hugovk:
I've resolved the conflicts through the GitHub web UI and the PR passes, but let me know if you'd rather a cleaner rebase and I'll do it tomorrow.

hugovk:
PR rebased and cleaned up!

KennethNielsen:
+1 from me. Will see if there is any input from others (@stefankoegl, @DPH opinions?), otherwise I will merge in a couple of days.

DPH:
+1 Makes sense, and keeps us up to date.

KennethNielsen:
Merged. Thanks for your contribution.

Issue #519, opened 2017-10-04
snapshot throwing an error


rudyvan:
Hello,

i have a large sonos setup with a TV soundbar.
Sometimes (once every 50 calls) i get a snapshot() error.
The documentation states that snapshot should return True when the zone is a coordinator.
Why an error then returns that the cross_fade can only be called on a coordinator?
I am using the latest build.
See the error message below:

```python
2017/10/03 14:37:12 Crash of PI-Deploy
Traceback (most recent call last):
  File "/home/pi/PI_Voice.py", line 125, in sonos_thread
    if resume[zone.ip_address].snapshot(): # it returns True when the snapshot zone is a coordinator
  File "/usr/local/lib/python3.5/dist-packages/soco/snapshot.py", line 101, in snapshot
    self.cross_fade = self.device.cross_fade
  File "/usr/local/lib/python3.5/dist-packages/soco/core.py", line 103, in inner_function
    raise SoCoSlaveException(message)
soco.exceptions.SoCoSlaveException: The method or property "cross_fade" can only be called/used on the coordinator in a group'
```
Thanks for the help.
Kind Regards,
Rudy

DPH:
Hi Rudy,
does this occur on the sound bar when it is playing TV?
If so cross fade is not valid on TV input!

We would need to tweak the snapshot code to exclude setting this when TV in use. 
Cheers David

rudyvan:
Indeed it probably is the case.
Is it possible to modify the code and bypass cross fade when the device does not support it or would trow an error?
Kind Regards,
Rudy

DPH:
Hi @rudyvan ,
have looked at this and it is on getting the cross fade status that it errors. 
Are you re-initialising snapshot every time, or is the function reused. 
The current set up requires you reinitialise each time, otherwise some variables are not set correctly.
I think this is what is causing the problem. 

Check   that
```resume[zone.ip_address] = Snapshot(zone) ```
occurs every time before 
```if resume[zone.ip_address].snapshot():```

let me know if this fixes it, or if this is not the problem and I will dig deeper. 
Cheers David

**Comment by rudyvan** [comment]
Hello David,

i called SoCoCam() each time before calling snapshot().
```python
resume[zone.ip_address]=SoCoCam(zone);
```
i will now replace with your suggestion to call Snapshot(zone) instead and report back if it is not solved.

Thanks for your help!
Kind Regards,
Rudy

rudyvan:
Apologies David,

[...]

PR #529, opened 2017-10-31
Change to use SoCo method to determine coordinator

DPH:
Fixes way snapshot determines if device is coordinator. 
Previous method worked but appears to have created the odd error - see #519 
Don't understand why old approach sometimes failed, but this change uses the standard SoCo approach.


KennethNielsen:
Looks good. I'll merge it. This might make conflicts for PR #530, since the code changes are close. But they should be easy to fix.

KennethNielsen:
Please leave a comment about your changes in the release notes draft issue #440

Issue #521, opened 2017-10-09
Snapshot can't restart cloud queues started with Alexa


DPH:
In testing with Alexa and the new Sonos Voice activation I notice that queues started by Alexa fail to restart when using snapshot.  Sonos / Alexa use a cloud queue when starting tracks with voice which disappears once you play something else. Radio started with Alexa is OK.

Will enhance snapshot in the next few days to stop it trying to re-start these queues, as it currently throws and error. Over time we may be able to find a way to restart them . 

Assigned to self
Regards David

DPH:
PR #530 removes error.

PR #530, opened 2017-10-31
Prevent error when queue started from Alexa

DPH:
Sonos can now be controlled by Amazon Alexa, which creates 'cloud queues'. 
However, when playing a cloud queue started by Alexa, there is no was to restart that queue from SoCo.
This PR prevents the error, by not restarting cloud queues #521 
This will hopefully reduce support queries.

Over time I trust we can find a way to make this work. 
Have written to Sonos about it, but got their typical response of not a supported function!

Cheers David

### DIFF
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -97,8 +97,8 @@ class Snapshot(object):
         if self.media_uri.split(':')[0] != 'x-rincon':
             self.is_coordinator = True
         if self.media_uri.split(':')[0] == 'x-rincon-queue':
-            if self.media_uri.split('#')[1] \
-                    == '0':  # pylint: disable=simplifiable-if-statement
+            # pylint: disable=simplifiable-if-statement
+            if self.media_uri.split('#')[1] == '0':
                 # playing local queue
                 self.is_playing_queue = True
             else:
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -93,7 +93,16 @@ class Snapshot(object):
         media_info = self.device.avTransport.GetMediaInfo([('InstanceID', 0)])
         self.media_uri = media_info['CurrentURI']
 
-        # extract source from media uri
+        # Extract source from media uri - below some media URI value examples:
+        #  'x-rincon-queue:RINCON_000E5859E49601400#0'
+        #       - playing a local queue always #0 for local queue)
+        #
+        #  'x-rincon-queue:RINCON_000E5859E49601400#6'
+        #       - playing a cloud queue where #x changes with each queue)
+        #
+        #  -'x-rincon:RINCON_000E5859E49601400'
+        #       - a slave player pointing to coordinator player
+
         if self.media_uri.split(':')[0] != 'x-rincon':
             self.is_coordinator = True
         if self.media_uri.split(':')[0] == 'x-rincon-queue':
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -7,6 +7,12 @@ This is useful for scenarios such as when you want to switch to radio
 or an announcement and then back again to what was playing previously.
 
 Warning:
+    Sonos has introduced control via Amazon Alexa. A new cloud queue is
+    created and at present there appears no way to restart this
+    queue from snapshot. Currently if a cloud queue was playing it will
+    not restart.
+
+Warning:
     This class is designed to be created used and destroyed. It is not
     designed to be reused or long lived. The init sets up defaults for
     one use.
@@ -50,6 +56,7 @@ class Snapshot(object):
         self.media_uri = None
         self.is_coordinator = False
         self.is_playing_queue = False
+        self.is_playing_cloud_queue = False
 
         self.volume = None
         self.mute = None
@@ -90,7 +97,12 @@ class Snapshot(object):
         if self.media_uri.split(':')[0] != 'x-rincon':
             self.is_coordinator = True
         if self.media_uri.split(':')[0] == 'x-rincon-queue':
-            self.is_playing_queue = True
+            if self.media_uri.split('#')[1] == '0':
+                # playing local queue
+                self.is_playing_queue = True
+            else:
+                # playing cloud queue - started from Alexa
+                self.is_playing_cloud_queue = True
 
         # Save the volume, mute and other sound settings
         self.volume = self.device.volume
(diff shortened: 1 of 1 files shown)


DPH:
Not sure how to fix the continuous-intergration error:

KennethNielsen:
@DPH I think that pylint error weird. I usually hesistant about saying that those don't make sense, because usually it is just me that don't understand it, but ...

I mean, the value of `self.media_uri.split('#')[1]` would not determine the truth value in itself, since it is a string and they are true or false depending on whether they are non-empty or not, and so sin you are comparing to a one-letter string, we cannot simplify to just:
```python
self.is_playing_queue = bool(self.media_uri.split('#')[1])
```

Since you initialize both values to False in `__init__` I guess we might instead write:
```python
self.is_playing_queue = self.media_uri.split('#')[1] == '0'
self.is_playing_cloud_queue = not self.is_playing_queue
```
but I certainly wouldn't call that easier to read and I don't know if that would fix it.

Unless someone shows up with a better explanation, I suggest disabling that warning for now.

Actually googling it, it seems that at least in this thread, someone comments exactly on out type of example and suggests that it is wrong: https://github.com/PyCQA/pylint/issues/837#issuecomment-255109936. With this information I suggest disabling it along with a comment saying that it is likely buggy and that the disable could probably be removed sometime in the future.

soco/snapshot.py
```diff
@@ -90,7 +97,12 @@ def snapshot(self):
         if self.media_uri.split(':')[0] != 'x-rincon':
             self.is_coordinator = True
         if self.media_uri.split(':')[0] == 'x-rincon-queue':
-            self.is_playing_queue = True
+            if self.media_uri.split('#')[1] == '0':
```
KennethNielsen: Do you have some idea how robust this check is. I can't really tell from the code. Can you comment, maybe just here, what the uris might look like in the different cases

DPH: I think this is robust from analysis and current use. The is_playing_queue has been in use in Snapshot since it's inception. The addition here is to differentiate the types of queues now Sonos have introduced a cloud queue to work with Alexa.

Special URI's in Sonos:

x-sonosapi-stream: ....                   radio stream
x-sonosapi-radio: .....                     radio stream / Amazon Station / Google mixes

x-rincon-queue:RINCON_000E5859E49601400#0   - local queue (note: #0 always zero for local queue)

x-rincon-queue:RINCON_000E5859E49601400#6   - cloud queue (note: #x where x changes with each queue)

x-rincon:RINCON_000E5859E49601400   -  slave pointing to coordinator 

Although I have never seen a slave not start 'x-rincon:' the snapshot code has been changed to use device.is_coordinator in place of this #529 as there were reported issues #519 which on analysis appeared to show what snapshot thought was not a slave was a slave in SoCo!  As intermittent hard to debug, but this approach makes it consistent.

Hope this makes sense - may be worth adding this in a documentation area so we can build on it as we discover more....
Shout if you want more.

Cheers David

DPH:
Eventually got pylint to ignore line.

soco/snapshot.py
KennethNielsen: Ah makes sense and I agree should be robust. In general, for code that isn't immediately obvious, like string determination with split and indexing, I think it is always worth it to explain in a comment above with an example. So basically a small comment around your:

```text
x-rincon-queue:RINCON_000E5859E49601400#0 - local queue (note: #0 always zero for local queue)

x-rincon-queue:RINCON_000E5859E49601400#6 - cloud queue (note: #x where x changes with each queue)

x-rincon:RINCON_000E5859E49601400 - slave pointing to coordinator
```
observation would be great.

DPH:
@KennethNielsen good idea have added some documentation.
Had a bit of a problem with conflicts but think I have it sorted.
Cheers David

KennethNielsen:
Hi @DPH 

Looks great with the comments. I think ready to merge. However, the history of this one turned out a little complicated (see below) due to the conflicts etc. and I think also it would be nice to have the relatively small commit squashed into fewer. You wanna try your hand at a rebase, otherwise I can do it when I merge.

```text
*   d062923 (HEAD, DPH/snap_alexa) Merge branch 'master' into snap_alexa
|\  
* \   c90aa65 Merge branch 'master' into snap_alexa
|\ \  
| |/  
| *   5ad05f3 (upstream/master) Merge pull request #528 from hugovk/add-3.6
| |\  
| | * 2b8bdb1 allow_failures for 3.7-dev and nightly
| | * 533fbef Test PyPy3 and CPython dev branches
| | * 2e2bdac Add support for Python 3.6
| |/  
| *   8e1347e Merge pull request #529 from DPH/snap_coord
| |\  
| | * deea8ca (DPH/snap_coord) Change to use SoCo method to determine coordinator
| * |   11515c4 Merge pull request #526 from hugovk/rm-2.6
| |\ \  
| | |/  
| |/|   
| | * 18fda89 Label ImportError
| | * e17118d 'Changed in version 3.3: This module will use a fast implementation whenever available. The xml.etree.cElementTree module is deprecated.'
| | * 2f3d39c Tidy up OrderedDict import
| | * 78e6592 Update string formatters
| | * 82639c3 Use set literals
| | * e46388c Run slowest PyPy first so as not to hold up the build
| | * 1d8fe20 Drop support for EOL Python 2.6
* | | fe53c41 added media URI documentation 3
* | | 0aab625 added media URI documentation 2
* | | 918661e added medai URI documentation
* | | 070e950 Fix pylint error simplifiable-if-statement - line too long 2!
* | | 191079a Fix pylint error simplifiable-if-statement - line too long!
* | | 98dbbef Fix pylint error simplifiable-if-statement
* | | dcdf2bd Fix line too long....
* | | b2313ca Prevent error when queue started from Alexa
|/ /  
* |   4dd3ad1 (upstream/python-2.6) Merge pull request #522 from hugovk/patch-1
|\ \  
| |/  
|/|   
| * ac77262 Remove broken downloads badge
|/  
*   06a5e75 (origin/master, origin/HEAD, master) Merge pull request #491 from DPH/force_stream
```

DPH:
Hi @KennethNielsen, 
as you can see I got into a bit of a mess with Git doing this simple change!!
Need to practice more on git.

Had a quick look at rebase but will need to learn it and try it out on my own stuff before hitting SoCo.
If you can do this one when you merge I would appreciate it as I don't want to hold up the release. 
I will  learn it for next time. 

Thanks David

KennethNielsen:
@DPH no problem. Rebased and merged. Thanks for your contribution. Please add a comment about the change to #440

PR #532, opened 2017-11-04
Update is_playing_radio for AAC streams

amelchio:
I found this method for detecting a radio stream in home-assistant/home-assistant#4063 and thought it fit better within SoCo.

Since I do not know the inside of SoCo, I actually do not know whether moving from `GetPositionInfo` to `GetMediaInfo` is okay at this place?

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1104,12 +1104,14 @@ class SoCo(_SocoSingletonBase):
 
         return True or False
         """
-        response = self.avTransport.GetPositionInfo([
-            ('InstanceID', 0),
-            ('Channel', 'Master')
-        ])
-        track_uri = response['TrackURI']
-        return re.match(r'^x-rincon-mp3radio:', track_uri) is not None
+        response = self.avTransport.GetMediaInfo(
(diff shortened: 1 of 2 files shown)


amelchio:
This is the version currently used in Home Assistant, it includes even more schemes:
```
def _is_radio_uri(uri):
    """Return whether the URI is a radio stream."""
    radio_schemes = (
        'x-rincon-mp3radio:', 'x-sonosapi-stream:', 'x-sonosapi-radio:',
        'hls-radio:')
    return uri.startswith(radio_schemes)
```

Also note that it tests a provided URI, not the currently playing one. This is used when playing a favorite:
```
        if _is_radio_uri(uri):
            self.soco.play_uri(uri, title=source)
        else:
            self.soco.clear_queue()
            self.soco.add_to_queue(src.reference)
            self.soco.play_from_queue(0)
```

Am I doing this right? If so, would you accept a new function that works on an arbitrary string?

amelchio:
I am closing this in preference of #587.

PR #533, opened 2017-11-04
Avoid lost events during subscription call

amelchio:
I have experienced some inconsistent behavior during startup. This appears to be caused by a race in event registration: an event can come in before its handler is registered.

This PR corrects the issue by extending `_sid_to_service_lock` to cover the full registration.

That simple change is scary for two reasons:
1. The lock now covers a `requests.request` call that could take some time.
2. <s>It introduces nested locking with `_sid_to_event_queue_lock` so a deadlock could potentially happen down the road.</s> (now fixed, see comments)

A fix that addresses these issues will be more complex (which is also scary) but I will be happy to provide one if you prefer to avoid the drawbacks listed above.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -466,30 +466,32 @@ class Subscription(object):
         }
         if requested_timeout is not None:
             headers["TIMEOUT"] = "Second-{}".format(requested_timeout)
-        response = requests.request(
-            'SUBSCRIBE', service.base_url + service.event_subscription_url,
-            headers=headers)
-        response.raise_for_status()
-        self.sid = response.headers['sid']
-        timeout = response.headers['timeout']
-        # According to the spec, timeout can be "infinite" or "second-123"
-        # where 123 is a number of seconds.  Sonos uses "Second-123" (with a
-        # capital letter)
-        if timeout.lower() == 'infinite':
-            self.timeout = None
-        else:
-            self.timeout = int(timeout.lstrip('Second-'))
-        self._timestamp = time.time()
-        self.is_subscribed = True
-        log.info(
-            "Subscribed to %s, sid: %s",
-            service.base_url + service.event_subscription_url, self.sid)
-        # Add the queue to the master dict of queues so it can be looked up
-        # by sid
-        with _sid_to_event_queue_lock:
-            _sid_to_event_queue[self.sid] = self.events
-        # And do the same for the sid to service mapping
+
+        # Lock out EventNotifyHandler during registration
         with _sid_to_service_lock:
+            response = requests.request(
+                'SUBSCRIBE', service.base_url + service.event_subscription_url,
+                headers=headers)
+            response.raise_for_status()
+            self.sid = response.headers['sid']
+            timeout = response.headers['timeout']
+            # According to the spec, timeout can be "infinite" or "second-123"
+            # where 123 is a number of seconds.  Sonos uses "Second-123" (with
+            # a capital letter)
+            if timeout.lower() == 'infinite':
+                self.timeout = None
+            else:
+                self.timeout = int(timeout.lstrip('Second-'))
+            self._timestamp = time.time()
+            self.is_subscribed = True
+            log.info(
+                "Subscribed to %s, sid: %s",
+                service.base_url + service.event_subscription_url, self.sid)
+            # Add the queue to the master dict of queues so it can be looked up
+            # by sid
+            with _sid_to_event_queue_lock:
+                _sid_to_event_queue[self.sid] = self.events
+            # And do the same for the sid to service mapping
             _sid_to_service[self.sid] = self.service
         # Register this subscription to be unsubscribed at exit if still alive
         # This will not happen if exit is abnormal (eg in response to a
--- a/soco/events.py
+++ b/soco/events.py
@@ -227,11 +227,11 @@ class EventNotifyHandler(BaseHTTPRequestHandler):
         content_length = int(headers['content-length'])
         content = self.rfile.read(content_length)
         # Find the relevant service and queue from the sid
-        with _sid_mapping_lock:
-            service = _sid_to_service.get(sid)
-            event_queue = _sid_to_event_queue.get(sid)
+        with _subscriptions_lock:
+            subscription = _subscriptions.get(sid)
         # It might have been removed by another thread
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@lawrenceakka the events module and thread safety, would you have time to look at this PR. Otherwise let me know, then I will muddle through.

lawrenceakka:
I’ll take a look,  it I’m travelling for the rest of this week, so it’ll be next week, hopefully early. 

Lawrence

amelchio:
I don't know how to add a proper test for this but I have made a script showing the problem. With this PR included it will consistently print my number of players. Without the fix it sometimes prints a lower number and logs this message for each missing count:
```
INFO:soco.events:No service registered for uuid:RINCON_XXXXXXXXXXXXX1400_sub0000000524
```

This issue is very important to me so I hope you can include a fix in 0.13.

```python
#!/usr/bin/env python3
import soco
import logging
import time
import threading

players = soco.discover()

logging.basicConfig()
logging.getLogger('soco.events').setLevel(logging.INFO)

counter = 0
class Queue():
    def put(self, item, block=True, timeout=None):
        global counter
        counter = counter+1

def work():
    x = 0
    while x < 1000000:
        x = x+1    

def subscribe(player):
    sub = player.avTransport.subscribe(event_queue=Queue())
    time.sleep(2)
    sub.unsubscribe()

for _ in range(0, 10):
    threading.Thread(target=work).start()

threads = []
for player in players:
    threads.append(threading.Thread(target=subscribe, args=(player,)))
[t.start() for t in threads]
[t.join() for t in threads]

print(counter)
```

KennethNielsen:
@lawrenceakka thanks.

@amelchio my immediate thoughts on this, is that we want to avoid the possibility of a dead lock at all costs, so if you want to look into a solution that does that, while you wait for feedback from @lawrenceakka that might be an idea. But in any case I'll leave the rest of the feedback to @lawrenceakka .

amelchio:
To be clear: this PR does not introduce a potential dead-lock. It just establishes a lock hierarchy that could be violated by another code change in the future.

I will await @lawrenceakka feedback before working more on this, so I know which direction to take.

amelchio:
@lawrenceakka Ping?

amelchio:
Can I do something to move this forward?

ghcs27:
I'm no maintainer, but I have two questions about this PR:
- If you introduce a lock hierarchy, how do you want to make sure future contributors don't violate it?
  - Document it in a prominent place. Future developers working with the locks should know what to do.
  - If you want to be absolutely sure: [Enforce your rule.](http://chimera.labs.oreilly.com/books/1230000000393/ch12.html#deadlock)
- Is there a reason you cannot move `_sid_to_service[self.sid] = self.service` before the subscription call? `self.service` is already known then, and the dictionary is only used in `do_NOTIFY`, so no harm done.
  See [here](https://github.com/ghcs27/SoCo/commit/92056c4e6e50c04e2f0b73673a8e8575941f5551) for my suggestion.

amelchio:
Thanks for reviewing! To answer your questions:
1. That is actually also _my_ question. There are several ways to handle this but they all have drawbacks and I do not know the SoCo preferred way to do it.
Since the two mappings are always used in tandem, it may be simpler to cover both of them with a single lock – that will certainly remove the lock hierarchy issue. I have pushed an additional commit implementing this proposal.
2. The reason is that `self.sid` is unknown until we get the response from the subscription call. Unfortunately, the `do_NOTIFY` handler might get its first event before we even see the subscription response and that is the issue I am trying to fix.

ghcs27:
1. I'd say this solves the lock hierarchy issue. While it still feels strange to hold the lock during a network call, I'd say it's the right behavior.
2. Yeah, I didn't notice that `self.sid` was missing.
- I'd recommend you have a closer look at what happens if event_queue is None in [line 232](https://github.com/SoCo/SoCo/pull/533/files#diff-b90b8922e8da075f0daf49298faedbc6R232). It's probably okay like this, but better to be sure.

amelchio:
That is not a problem because the two mapping are always updated at the same time. So if the service is set, so is the event_queue. To make this more clear, I made another commit that removes one of the mappings and makes the other one point at the Subscription (which has both the service and the event_queue). I quite like how that turned out 😃

I think we can avoid locking during the network call by having an internal buffer of undelivered events and flushing it when a subscription completes. However, subscribing is rare so I prefer the coarse lock which is much simpler.

ghcs27:
Yeah, this looks goot to me. :+1:
I don't know how the maintainers see it, though - Coluld you please have a look, @lawrenceakka and/or @KennethNielsen?

ghcs27:
I just saw that the build fails, but it isn't related to this PR.

*For the maintainers:
Pylint got updated, and it now complains if a function with a return value doesn't reach a return statement in all cases (using an implicit `return None` at the end of the function). We can disable this warning, but it sounds like a valid code style complaint and we only violate it in 4 places, so we might as well make our code compliant.*

amelchio:
@ghcs27 Thanks so much for reviewing this, the code is much better now.

I have updated the PR title to better reflect the current proposal.

amelchio:
I have a big rework of the Sonos integration in Home Assistant that is currently bit rotting because it needs this fix. I would like some indication whether this PR will move forward towards a PyPI release or I should consider alternative ways of making the fix available to Home Assistant.

KennethNielsen:
@amelchio I'll see about getting this through now. It seems you and @ghcs27 had some constructive discussions. I will review to see where we are at.

KennethNielsen:
Thanks @amelchio and @ghcs27 for your work on this.

I must admit that the exchanging two WeakValueDicts and Locks with one, had me somewhat uneasy at first, because I was worried about missing some subtlety, that would make it a bad idea. But I have read it a few times now and I can't find it. So I will merge this as is.

@lawrenceakka if you get a little time, you can look over the change mentioned in the paragraph above and complain if we missed something

Nice work.

@amelchio remember to add a comment to the release notes thread #551 describing the change.

amelchio:
Thank you! I added a short comment for the release note.

KennethNielsen:
@amelchio BTW. We are trying for more regular timed releases. The plan is to release the next version on feb. 17 (as per #538).

PR #534, opened 2017-11-04
Decrease logging level

amelchio:
This error message shows up frequently for me but it appears to not be an error at all.

### DIFF
--- a/soco/data_structures_entry.py
+++ b/soco/data_structures_entry.py
@@ -59,7 +59,7 @@ def from_didl_string(string):
             # causes problems.
             raise DIDLMetadataError("Illegal child of DIDL element: <%s>"
                                     % elt.tag)
-    _LOG.error(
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Ouch:
```text
$ git blame data_structures_entry.py -L62,+4
e6c52050 (Kenneth Nielsen 2016-10-19 22:40:58 +0200 62)     _LOG.error(
e6c52050 (Kenneth Nielsen 2016-10-19 22:40:58 +0200 63)         'Created data structures: %.20s (CUT) from Didl string "%.20s" (CUT)',
e6c52050 (Kenneth Nielsen 2016-10-19 22:40:58 +0200 64)         items, string,
e6c52050 (Kenneth Nielsen 2016-10-19 22:40:58 +0200 65)     )
```

That error level looks like it was left around after being used during development :$

Merged. Thanks.

PR #537, opened 2017-11-19
Alexa support

tagdara:
The Alexa update added new metadata components and changed the use of several previously mandatory fields.

These new formats seem to only be used when an Alexa request for music has been placed.  Music queued through the apps or through the API like Soco retains the old format.

These changes are used to prevent exceptions on those fields which kill subscription threads.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -194,9 +194,14 @@ class DidlResource(object):
         content = {}
         # required
         content['protocol_info'] = element.get('protocolInfo')
-        if content['protocol_info'] is None:
-            raise Exception('Could not create Resource from Element: '
-                            'protocolInfo not found (required).')
+        
+        # With the Alexa update, the 'protocolInfo' tag does not appear on data containers where it 
+        # possibly should and probably did in the past.  This execption should be ignored with the new formats.
+          
+        #if content['protocol_info'] is None:
+        #    raise Exception('Could not create Resource from Element: '
+        #                    'protocolInfo not found (required).')
+        
         # Optional
         content['import_uri'] = element.get('importUri')
         content['size'] = _int_helper('size')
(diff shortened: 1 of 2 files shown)


KennethNielsen:
Uff. This Alexa stuff is horrible. I find it weird that Sonos isn't any better than this to get their partners to follow spec and I somewhat expect that they will push for it. So the changes to SoCo must be reversible.

With regard to the specific changes, they have been marked by comments as being Alexa specific, but I think they should be Alexa specific in code if possible.

Can you look into whether it is possible to detect Alexa items, possibly by looking at the uri, and then make these changes specific to Alexa items only.

KennethNielsen:
Besides, the branch now has conflicts so can you rebase of of master?

ghcs27:
This PR has become a lot less pressing now that we have #567. I'd say we should wait until we figure out exactly how the Alexa integration works.

KennethNielsen:
@ghcs27 agreed, however, one of the changes suggested here are in events and I'm not sure covered by #567, we will see, we can always do a point release if SoCo remain broken for Aleaxa

ghcs27:
What I mean is:

> These changes are used to prevent exceptions on those fields which kill subscription threads.

With #567, the exceptions will still be thrown, but they won't kill the thread.
The change in events.py would just exclude the `av_transport_uri_meta_data` field from parsing.

relevitt:
@stefankoegl Please let me know if this pull request is to be merged in 0.15, as I'll need to include it the work I'm doing on events?

stefankoegl:
What do we need to proceed with this?

KennethNielsen:
As far as I can see then problem with the missing restricted tag is already taken care of and I just merged a proposed fix for the missing protocolInfo, so unless I'm mistaken this should now be redundant. I will close it, but feel free to re-open if that is not the case.

PR #539, opened 2017-11-21
Small bugfix in core.py

### LINKED ISSUE #548: Exception thrown when trackinfo not found
I'm seeing this when using the Sonos component with Home Assistant, with version 0.12 of SoCo:
```...
  File "/srv/homeassistant/lib/python3.4/site-packages/homeassistant/components/media_player/sonos.py", line 504, in update
    track_info = self._player.get_current_track_info()
  File "/srv/homeassistant/lib/python3.4/site-packages/soco/core.py", line 1051, in get_current_track_info
    index = trackinfo.find(' - ')
AttributeError: 'NoneType' object has no attribute 'find'
```

Looks like sometimes the Sonos can return metadata which doesn't contain the trackinfo, which breaks when we try to extract the index.

relevitt:
This is a small bugfix to stop an error where None would be returned by metadata.findtext. Instead, an empty string is returned.
Closes #548.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1215,7 +1215,7 @@ class SoCo(_SocoSingletonBase):
             metadata = XML.fromstring(really_utf8(metadata))
             # Try parse trackinfo
             trackinfo = metadata.findtext('.//{urn:schemas-rinconnetworks-com:'
-                                          'metadata-1-0/}streamContent')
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@relevitt thanks for the fix. Remember to add a comment to the release notes thread #551 about the change.

@ghcs27 thanks for the review.

PR #540, opened 2017-11-21
 Major docstring fixes and reformats (Napoleon Style FTW!) 

ghcs27:
Lots of core.py docstrings, reformatted into Napoleon style.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -431,9 +431,6 @@ class SoCo(_SocoSingletonBase):
 
         index: the index of the track to play; first item in the queue is 0
         start: If the item that has been set should start playing
-
-        Raises SoCoException (or a subclass) upon errors.
-
         """
         # Grab the speaker's information if we haven't already since we'll need
         # it in the next step.
@@ -461,10 +458,7 @@ class SoCo(_SocoSingletonBase):
 
     @only_on_master
     def play(self):
-        """Play the currently selected track.
-
-        Raises SoCoException (or a subclass) upon errors.
-        """
+        """Play the currently selected track."""
         self.avTransport.Play([
             ('InstanceID', 0),
             ('Speed', 1)
@@ -488,9 +482,6 @@ class SoCo(_SocoSingletonBase):
             title (str): The title to show in the player (if no meta).
             start (bool): If the URI that has been set should start playing.
             force_radio (bool): forces a uri to play as a radio stream.
-
-        Raises:
-            SoCoException: (or a subclass) upon errors.
 
         On a Sonos controller music is shown with one of the following display
         formats and controls:
@@ -559,10 +550,7 @@ class SoCo(_SocoSingletonBase):
 
     @only_on_master
     def pause(self):
-        """Pause the currently playing track.
-
-        Raises SoCoException (or a subclass) upon errors.
-        """
+        """Pause the currently playing track."""
         self.avTransport.Pause([
             ('InstanceID', 0),
             ('Speed', 1)
@@ -570,10 +558,7 @@ class SoCo(_SocoSingletonBase):
 
     @only_on_master
     def stop(self):
-        """Stop the currently playing track.
-
-        Raises SoCoException (or a subclass) upon errors.
-        """
+        """Stop the currently playing track."""
         self.avTransport.Stop([
             ('InstanceID', 0),
             ('Speed', 1)
@@ -583,9 +568,6 @@ class SoCo(_SocoSingletonBase):
     def seek(self, timestamp):
         """ Seeks to a given timestamp in the current track, specified in the
         format of HH:MM:SS or H:MM:SS.
-
-        Raises SoCoException (or a subclass) upon errors.
-
         """
         if not re.match(r'^[0-9][0-9]?:[0-9][0-9]:[0-9][0-9]$', timestamp):
             raise ValueError('invalid timestamp, use HH:MM:SS format')
@@ -599,8 +581,6 @@ class SoCo(_SocoSingletonBase):
     @only_on_master
     def next(self):
         """Go to the next track.
-
-        Raises SoCoException (or a subclass) upon errors.
 
         Keep in mind that next() can return errors
         for a variety of reasons. For example, if the Sonos is streaming
@@ -616,8 +596,6 @@ class SoCo(_SocoSingletonBase):
     @only_on_master
     def previous(self):
         """Go back to the previously played track.
-
-        Raises SoCoException (or a subclass) upon errors.
 
         Keep in mind that previous() can return errors
         for a variety of reasons. For example, previous() will return an error
@@ -1030,8 +1008,6 @@ class SoCo(_SocoSingletonBase):
         Seems to work ok even if you remove what was previously the group
         master from it's own group. If the speaker was not in a group also
         returns ok.
-
-        Raises SoCoException (or a subclass) upon errors.
         """
 
         self.avTransport.BecomeCoordinatorOfStandaloneGroup([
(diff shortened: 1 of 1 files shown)


KennethNielsen:
Without reviewing this. I imaging most of this is non-problematic and **much** needed. The only exception if removing the "Raises SoCoException" from the docs, which I'm not sure I agree on, and which I think should be discussed as matter of policy.

But I would like to get all the rest of your work in right away, to prevent this rather large PR to go stale. So how about cherrypicking commit e81bd71 into a different branch (and PR) and then rebasing this one, without that commit?. That way we can continue with the large part here immediately.

ghcs27:
Sorry, I messed up my Git files, so getting out this PR is not as easy as hoped. Will resolve that soon.

KennethNielsen:
@ghcs27 let me know if you need help. Remember you can always reset your local version of this branch to the one on github, that this PR is created from, to start working from this starting point. In any case, let me know if you need help with git.

ghcs27:
It will probably be faster to make a 'put raises back in' commit than to rebase the three following commits (lots of conflicts, some of them not easily resolvable).
Anyway, it's not that much work, I just have to find time to do it.

ghcs27:
I just put the 'Raises SoCoException' lines back in, but I just can't come to terms with it.
It's only in some seemingly random docstrings, although it is true of all of them. The current string also not compliant with the Napoleon style.
To make the docstrings consistent, we'll have to either put a Napoleon version of it in *all* the docstrings or remove it.

soco/core.py
```diff
@@ -686,7 +680,7 @@ def mute(self, mute):
 
     @property
     def volume(self):
-        """The speaker's volume.
+        """int: The speaker's volume.
 
         An integer between 0 and 100.
```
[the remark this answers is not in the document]
KennethNielsen: if it fits in a line anyway

soco/core.py
```diff
@@ -759,8 +753,9 @@ def treble(self, treble):
 
     @property
     def loudness(self):
-        """The Sonos speaker's loudness compensation. True if on, otherwise
-        False.
+        """bool: The Sonos speaker's loudness compensation.
+
+        True if on, False otherwise.
```
KennethNielsen: this might also be merged into the top line

soco/core.py
```diff
@@ -1113,9 +1088,7 @@ def is_playing_radio(self):
 
     @property
     def is_playing_line_in(self):
-        """ Is the speaker playing line-in?
-
-        return True or False
+        """bool: Is the speaker playing line-in?
         """
```
KennethNielsen: no newline

soco/core.py
```diff
@@ -1412,8 +1374,9 @@ def queue_size(self):
         return queue_size
 
     def get_sonos_playlists(self, *args, **kwargs):
-        """ Convenience method for:
-            get_music_library_information('sonos_playlists')
+        """ Convenience method for
```
KennethNielsen: No space before Con...

soco/core.py
```diff
@@ -1494,13 +1460,8 @@ def remove_from_queue(self, index):
         """ Remove a track from the queue by index. The index number is
```
KennethNielsen: Let's loose the space before Remove here as well

soco/core.py
```diff
@@ -1528,9 +1483,9 @@ def get_favorite_radio_shows(self, start=0, max_items=100):
         """Get favorite radio shows from Sonos' Radio app.
 
         Returns:
-        A list containing the total number of favorites, the number of
-        favorites returned, and the actual list of favorite radio shows,
-        represented as a dictionary with `title` and `uri` keys.
+            A list containing the total number of favorites, the number of
```
KennethNielsen: I suspected that this was wrong, so I checked the code. That first "list" should be "dict". But shouldn't that return type be documented properly. I guess these have disappeared anyway with one of your other PR's

soco/core.py
```diff
@@ -1707,9 +1649,10 @@ def remove_sonos_playlist(self, sonos_playlist):
     def add_item_to_sonos_playlist(self, queueable_item, sonos_playlist):
         """Adds a queueable item to a Sonos' playlist.
 
-        :param queueable_item: the item to add to the Sonos' playlist
-        :param sonos_playlist: the Sonos' playlist to which the item should
-            be added
+        Args:
+            queueable_item: the item to add to the Sonos' playlist
```
KennethNielsen: I guess both of these have types that could be referenced properly

soco/core.py
```diff
@@ -433,19 +470,14 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
-        required as an argument, where the first index is 0.
-
-        index: the index of the track to play; first item in the queue is 0
-        start: If the item that has been set should start playing
-
-        Returns:
-        True if the Sonos speaker successfully started playing the track.
-        False if the track did not start (this may be because it was not
-        requested to start because "start=False")
+        """ Play a track from the queue by index.
```
KennethNielsen: No space before Play

soco/core.py
```diff
@@ -1513,13 +1474,7 @@ def remove_from_queue(self, index):
...
     @only_on_master
     def clear_queue(self):
-        """Removes all tracks from the queue.
-
-        Returns:
-        True if the Sonos speaker cleared the queue.
-
-        Raises SoCoException (or a subclass) upon errors.
-        """
+        """Removes all tracks from the queue."""
```
KennethNielsen: It should probably be """Remove all..."""

KennethNielsen:
Just finished the review. Just a bunch of nit-pick comments. Nice work.

ghcs27:
Thanks for the review, seems like I made quite some formatting blunders.
The one thing I'm not sure I agree on is putting the value range into the first line. I feel this makes the first line less clear and concise. But if you feel strongly about this please let me know.

KennethNielsen:
@ghcs27 no I don't feel strongly about the range in the first line, so if you think it is better your way, let's go with that.

soco/core.py
```diff
@@ -1261,15 +1223,15 @@ def get_speaker_info(self, refresh=False, timeout=None):
...
+            timeout: How long to wait for the server to send
+                data before giving up, as a float, or a
+                `(connect timeout, read timeout)` tuple
+                e.g. (3, 5). Default is no timeout.
 
         Returns:
-        Information about the Sonos speaker, such as the UID, MAC Address, and
-        Zone Name.
+            dict: Information about the Sonos speaker, such as the UID,
+            MAC Address, and Zone Name.
```
KennethNielsen: does that format correctly. I thought the second line "MAC Add...." should be indented.

soco/core.py
```diff
@@ -1527,9 +1484,9 @@ def get_favorite_radio_shows(self, start=0, max_items=100):
         """Get favorite radio shows from Sonos' Radio app.
 
         Returns:
-        A list containing the total number of favorites, the number of
-        favorites returned, and the actual list of favorite radio shows,
-        represented as a dictionary with `title` and `uri` keys.
+            dict: A dictionary containing the total number of favorites, the
+            number of favorites returned, and the actual list of favorite radio
```
KennethNielsen: Same here. But I might be mistaken. If you have the compiled docs from this branch at hand, maybe you can just check if I'm correct

ghcs27:
It's actually correct like this: The 'Returns' section has to be the same indentation, it's a quirk of the napoleon convention. It took me a while to figure that out when I started...

KennethNielsen:
Great.Thanks for the work.

Merging.

Remember to add a note to the release notes.

PR #541, opened 2017-11-21
Allow specification of track number in 'seek()'

ghcs27:
This expands the seek function to handle both relative positions in a track and track number in the queue.

It is even possible to seek to a tuple or dict containing the absolute position (track and relative pos.):
```
t = ('0:00:00', 0)
player.seek(*t)
d = {'position': '0:00:00', 'track': 0}
player.seek(**d)
```
This will be useful for snapshot et al.

Closes #514 and #513


KennethNielsen:
I think I tend to favor this implementation, because it stays closer to exposed service functions. We might consider, if anyone asks for it to add a convenience `jump_to_track` method  that calls this one without a timestamp. We should proceed with this and close the other PR. Can I get you to rebase and possibly update this PR if you want to add changes and then I will review.

PR #367, opened 2015-09-22
Fix tests raising unwanted Deprecation Warnings

dundeemt:
fixed up tests that were calling recently moved .music_library methods. Closes issue #366.

PR #542, opened 2017-11-21
Remove the methods that were deprecated in 0.12

ghcs27:
Lots of methods were moved from core.py to music_library.py and scheduled for deprecation in the upcoming release 0.14 - this PR removes them as planned.
Closes #367.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -33,7 +33,7 @@ from .services import (
     zone_group_state_shared_cache,
 )
 from .utils import (
-    really_utf8, camel_to_underscore, deprecated
+    really_utf8, camel_to_underscore
 )
 from .xml import XML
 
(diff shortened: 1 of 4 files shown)


KennethNielsen:
Hi @ghcs27 

Of topic for this particular PR, but would you consider joining our IRC channel (#SoCo on freenode). Since you are quite active, it would be convenient for me to be able to reach you live sometimes, when we are both online.

Regards

ghcs27:
@KennethNielsen Joined!

KennethNielsen:
The changes to core are good to go, but the changes to the tests bother me. It actually has nothing to do with this PR, as those tests should have been moved along with moving the actual music library code. However, I don't think we should delete the test cases. So either they should be:

 1. Put in a test_mulibrary and properly ported
or
 2. Placed there and mark the whole file as "should be skipped" by pytest, until they have been ported and an help wanted issue should be created for actually porting the tests

ghcs27:
Oh... I assumed they had been copied to test_music_library.py, but never checked. I'll try to port them.

ghcs27:
Just realized that #367 ports some of the tests (without moving them), I can use that.

KennethNielsen:
@ghcs27 yeah, but they should still be moved, so that we keep the one test file per source code file correlation. Let me know (on IRC preferably) when it is ready for review.

KennethNielsen:
Merged. Remember to leave a comment about the change in the release notes issue.

PR #546, opened 2017-11-24
Handle none values, the 2nd

### LINKED ISSUE #453: avTransport events throw exception when connected via spotify direct

ghcs27:
Closes  #453.
Code was copied from @bosim's PR, so credit goes to him.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -456,7 +456,7 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
             # way.
             setattr(self, key, value)
 
-    # pylint: disable=too-many-locals
+    # pylint: disable=too-many-locals, too-many-branches
     @classmethod
     def from_element(cls, element):     # pylint: disable=R0914
         """Create an instance of this class from an ElementTree xml Element.
@@ -504,21 +504,18 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
 
         # CAUTION: This implementation deviates from the spec.
         # Elements are normally required to have a `restricted` tag, but
-        # Spotify direct violates this. To make it work, a missing restricted
+        # Spotify Direct violates this. To make it work, a missing restricted
         # tag is interpreted as `restricted = True`.
         restricted = element.get('restricted', None)
         restricted = False if restricted in [0, 'false', 'False'] else True
 
-        # There must be a title. According to spec, it should be the first
-        # child, but Sonos does not abide by this
+        # Similarily, all elements should have a title tag, but Spotify Direct
+        # does not comply
         title_elt = element.find(ns_tag('dc', 'title'))
-        if title_elt is None:
-            raise DIDLMetadataError(
-                "Missing title element")
-        title = title_elt.text
-        if title_elt.text is None:
+        if title_elt is None or not title_elt.text:
             title = ''
-        title = really_unicode(title)
+        else:
+            title = really_unicode(title_elt.text)
 
         # Deal with any resource elements
         resources = []
--- a/soco/events.py
+++ b/soco/events.py
@@ -25,6 +25,7 @@ from .xml import XML
 log = logging.getLogger(__name__)  # pylint: disable=C0103
 
 
+# pylint: disable=too-many-branches
 def parse_event_xml(xml_event):
     """Parse the body of a UPnP event.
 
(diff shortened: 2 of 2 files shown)


amelchio:
This fixes Spotify Direct for me 👍

soco/data_structures.py
```diff
@@ -504,21 +504,20 @@ def from_element(cls, element):     # pylint: disable=R0914
...
-        # child, but Sonos does not abide by this
+        # Similarily, all elements should have a title tag, but Spotify Direct
+        # does not comply
         title_elt = element.find(ns_tag('dc', 'title'))
         if title_elt is None:
-            raise DIDLMetadataError(
-                "Missing title element")
-        title = title_elt.text
-        if title_elt.text is None:
+            title = None
```
KennethNielsen: The one bit I don't like about this PR is line 516. According to the docs for the DIDL-Lite elements, the title is mandatory and it says nothing about a possible None value, so I think it should default to the empty string in this case as well.

ghcs27: I intentionally set title to None, because in this case, the title really nonexistent, not empty. Defaulting to an empty string would make it impossible to distinguish between missing and empty title.
But if you want the empty string as a default (might integrate better with other code?), that's also fine with me.

KennethNielsen: Yeah, I got that, and that might make sense. My point is just that the docs for the DidlLite Element says that the title is mandatory and a string. So either we should default to a string, or, change the DidlLiteElement docstring away from spec, to document the possibility of a None value for the title.

Initially I was against the idea of changing DidlLite implementation too much away from spec, but since we are already down that rabbit hole, I am now thinking that it might actually be the better option.

So if we go with that option, I would ask you to:

 1. Change the docstring in the appropriate DidlLite class
 2. Highlight in the release notes thread, that this fix introduced a change of possibly data types for a core data type

ghcs27:
Returning None or the default value if the attribute/tag is missing is actually not an easy decision, and it should be consistent (the restricted attribute should behave the same way), and I'm no longer sure which way is the best.
Returning None gives the user control of what should happen if the element is missing, but it's also dangerous because he might not be prepared for a None value. I'm

KennethNielsen:
Yes. The restricted tag is easier, because there we can default to a sensible value with the same type. I still fell that broken events is not worth it braking the DIDL spec or making a backwards incompatible change for. What you could do is if you default to an empty string, leave the extra if/elif in and write a comment about the change, which will make it easier to change later.

ghcs27:
Yeah, defaulting to an empty string is probably better - there will be very few situations when one has to differentiate between empty title and no title.
And of course: *Special cases aren't special enough to break the rules.* - The Zen of Python

KennethNielsen:
If you make a commit to that effect I'll merge right away

ghcs27:
Done.

KennethNielsen:
Great. Thanks for the contribution. Merged, remember the change log if it applies here.

ghcs27:
Great!
Is there a release notes thread already?

KennethNielsen:
There is now #551

PR #555, opened 2017-12-30
don't throw exception if alarm is not associated to any room

philippkeller:
I got this error:

```
/usr/local/lib/python3.5/dist-packages/soco/alarms.py in get_alarms(zone)
    308         instance.enabled = values['Enabled'] == '1'
    309         instance.zone = [z for z in zone.all_zones
--> 310                          if z.uid == values['RoomUUID']][0]
    311         instance.program_uri = None if values['ProgramURI'] ==\
    312             "x-rincon-buzzer:0" else values['ProgramURI']
IndexError: list index out of range
```

The problem is that some of my alarms have no rooms associated. This PR now skips all alarms with no associated room because in `save()` you do `self.zone.uid` which requires every room to have a zone. If you'd like to also allow alarms without zone, then please add a comment and I'll try to rework my PR.

### DIFF
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -307,7 +307,7 @@ def get_alarms(zone=None):
         instance.recurrence = values['Recurrence']
         instance.enabled = values['Enabled'] == '1'
         instance.zone = next((z for z in zone.all_zones
-                         if z.uid == values['RoomUUID']), None)
+                              if z.uid == values['RoomUUID']), None)
         # some alarms are not associated to zones -> filter these out
         if instance.zone is None:
             continue
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -306,8 +306,11 @@ def get_alarms(zone=None):
             datetime.strptime(values['Duration'], "%H:%M:%S").time()
         instance.recurrence = values['Recurrence']
         instance.enabled = values['Enabled'] == '1'
-        instance.zone = [z for z in zone.all_zones
-                         if z.uid == values['RoomUUID']][0]
+        instance.zone = next((z for z in zone.all_zones
+                         if z.uid == values['RoomUUID']), None)
+        # some alarms are not associated to zones -> filter these out
+        if instance.zone is None:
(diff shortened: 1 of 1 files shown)


soco/alarms.py
```diff
@@ -306,8 +306,11 @@ def get_alarms(zone=None):
             datetime.strptime(values['Duration'], "%H:%M:%S").time()
         instance.recurrence = values['Recurrence']
         instance.enabled = values['Enabled'] == '1'
-        instance.zone = [z for z in zone.all_zones
-                         if z.uid == values['RoomUUID']][0]
+        instance.zone = next((z for z in zone.all_zones
+                         if z.uid == values['RoomUUID']), None)
```
ghcs27: PEP8 complains: This line should be indented justified with the first 'z' from the previous line.

ghcs27:
Do you have any information on what kinds of alarms don't have a room associated?
Because I'm not sure how justified it is to just skip them.

philippkeller:
this only happens when you have an alarm bound to a sonos player which you reset. Then the alarm is not attached to any sonos player. As this is pretty much an edge case I don't think it makes sense to pursue this PR any longer

ghcs27:
Sorry for irritating you with my question. We certainly don't want our methods to fail, even in edge cases. I just wanted to know whether the user might want to know about the alarm, which doesn't seem the case. Therefore, I'd say this PR is actually good to go.

philippkeller:
Alright, no offense taken!
I guess the user is ok to ignore them as the alarms are "sleeping" anyway and cannot fire as they are not attached to any alarm. If I find some time I will try to make linter happy.

ghcs27:
If you want, I can do the lint fix for you (as a maintainer I can edit this PR).

philippkeller:
@ghcs27 thanks for the patience. I now fixed the lint error. Is this PR now ready to go or is there another thing I should do?

stefankoegl:
Thanks!

PR #563, opened 2018-01-14
Make Subscription a context manager

ghcs27:
This adds `__enter__` and `__exit__` methods to `Subscription`, to make the following idiom possible:
```python
with player.AVTransport.subscribe() as subscription:
    # Dummy
    event = subscription.events.get()
```
The subscription is automatically unsubscribed when exiting the `with` block.


KennethNielsen:
Nice addition. Merged.

Remember the release notes.

PR #567, opened 2018-01-19
Delay object parsing exceptions, reraise on object use

### LINKED ISSUE #566: Make parsing errors in events less fatal
Recently, many issues have arisen because Sonos doesn't adhere to the Didl spec, causing exceptions from from_didl_string - the next one's already at #520. The problem is exacerbated by the fact that for events, non-compliant metadata causes an exception at parsing time, which kills the subscription thread.
This behavior is unacceptable, especially if we are headed for a 1.0 release. I propose that we delay the exception until the user touches the faulty object, so he can handle it. We could modify the 'variables' dict to reraise the exception when the faulty variable is retrieved, or make a DidlFault object which reraises if anything is called on it.

ghcs27:
This catches any exceptions thrown by `from_didl_string` during event handling. Instead a DidlFault is created, which rethrows the exception when the user tries to use it. This avoids an exception in the event thread, instead letting the user handle it when he needs the object. 
Compare with #568 for another implementation of the same proposal.
See #566 for information on why this is needed.
Closes #566, closes #568.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -22,7 +22,7 @@ from .compat import (
     Queue, BaseHTTPRequestHandler, URLError, socketserver, urlopen
 )
 from .data_structures_entry import from_didl_string
-from .exceptions import SoCoException
+from .exceptions import SoCoException, SoCoFault
 from .utils import camel_to_underscore
 from .xml import XML
 
@@ -136,7 +136,12 @@ def parse_event_xml(xml_event):
                     # If DIDL metadata is returned, convert it to a music
                     # library data structure
                     if value.startswith('<DIDL-Lite'):
-                        value = from_didl_string(value)[0]
+                        # Wrap any parsing exception in a SoCoFault, so the
+                        # user can handle it
+                        try:
+                            value = from_didl_string(value)[0]
+                        except SoCoException as ex:
+                            value = SoCoFault(ex, metadata=value)
                     channel = last_change_var.get('channel')
                     if channel is not None:
                         if result.get(tag) is None:
--- a/soco/events.py
+++ b/soco/events.py
@@ -22,7 +22,7 @@ from .compat import (
     Queue, BaseHTTPRequestHandler, URLError, socketserver, urlopen
 )
 from .data_structures_entry import from_didl_string
-from .exceptions import SoCoException, SoCoFault
+from .exceptions import SoCoException, SoCoFault, EventParseException
 from .utils import camel_to_underscore
 from .xml import XML
 
@@ -141,13 +141,16 @@ def parse_event_xml(xml_event):
                         # user can handle it
                         try:
                             value = from_didl_string(value)[0]
-                        except SoCoException as ex:
-                            log.error("Event contains illegal metadata"
-                                      "for '%s'.\n"
-                                      "Error message: '%s'\n"
-                                      "The result will be a SoCoFault.",
-                                      tag, str(ex))
-                            value = SoCoFault(ex, metadata=value)
+                        except SoCoException as original_exception:
+                            log.warning("Event contains illegal metadata"
+                                        "for '%s'.\n"
+                                        "Error message: '%s'\n"
+                                        "The result will be a SoCoFault.",
+                                        tag, str(original_exception))
+                            event_parse_exception = EventParseException(
+                                tag, value, original_exception
+                            )
+                            value = SoCoFault(event_parse_exception)
                     channel = last_change_var.get('channel')
                     if channel is not None:
                         if result.get(tag) is None:
--- a/soco/events.py
+++ b/soco/events.py
@@ -47,6 +47,7 @@ def parse_event_xml(xml_event):
               :code:`{'Volume': {'LF': '100', 'RF': '100', 'Master': '36'}}`)
             * an instance of a `DidlObject` subclass (eg if it represents
               track metadata).
+            * a `SoCoFault` (if a variable contains illegal metadata)
 
     Example:
 
@@ -173,7 +174,8 @@ class Event(object):
             `time.time` function).
         service (str): the service which is subscribed to the event.
         variables (dict, optional): contains the ``{names: values}`` of the
-            evented variables. Defaults to `None`.
+            evented variables. Defaults to `None`. The values may be
+            `SoCoFault` objects if the metadata could not be parsed.
 
     Raises:
         AttributeError:  Not all attributes are returned with each event. An
--- a/soco/events.py
+++ b/soco/events.py
@@ -141,6 +141,11 @@ def parse_event_xml(xml_event):
                         try:
                             value = from_didl_string(value)[0]
                         except SoCoException as ex:
+                            log.error("Event contains illegal metadata"
+                                      "for '%s'.\n"
+                                      "Error message: '%s'\n"
+                                      "The result will be a SoCoFault.",
+                                      tag, str(ex))
                             value = SoCoFault(ex, metadata=value)
                     channel = last_change_var.get('channel')
                     if channel is not None:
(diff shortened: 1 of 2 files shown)


KennethNielsen:
I think I tend to favor this implementation. Then end use result is similar and this implementation is simpler. What do you think about that @ghcs27 ?

soco/events.py
```diff
@@ -136,7 +137,17 @@ def parse_event_xml(xml_event):
                     # If DIDL metadata is returned, convert it to a music
                     # library data structure
                     if value.startswith('<DIDL-Lite'):
-                        value = from_didl_string(value)[0]
+                        # Wrap any parsing exception in a SoCoFault, so the
+                        # user can handle it
+                        try:
+                            value = from_didl_string(value)[0]
+                        except SoCoException as ex:
```
KennethNielsen: Nitpick: Suggest writing exception all out

soco/events.py
```diff
@@ -136,7 +137,17 @@ def parse_event_xml(xml_event):
...
+                            log.error("Event contains illegal metadata"
+                                      "for '%s'.\n"
+                                      "Error message: '%s'\n"
+                                      "The result will be a SoCoFault.",
+                                      tag, str(ex))
```
KennethNielsen: I think the proper way to create a traceback as a string from a exceptions is to use the traceback module: https://stackoverflow.com/questions/4564559/get-exception-description-and-stack-trace-which-caused-an-exception-all-as-a-st

soco/exceptions.py
```diff
@@ -80,3 +80,37 @@ class SoCoSlaveException(SoCoException):
...
+
+    def __getitem__(self, item):
+        raise self.exception
+
+    def __repr__(self):
+        return '<{0}: {1} at {2}>'.format(self.__class__.__name__,
+                                          repr(self.exception),
+                                          hex(id(self)))
+
+    def __str__(self):
```
KennethNielsen: When inheriting from object, __str__ seems to default to calling __repr__, so this methods seems unnecessary.

soco/exceptions.py
```diff
@@ -80,3 +80,37 @@ class SoCoSlaveException(SoCoException):
 
 class NotSupportedException(SoCoException):
     """Raised when something is not supported by the device"""
+
+
+class SoCoFault(object):
+    """Class to represent a failed object instantiation.
+
+    It rethrows the exception on common use.
```
KennethNielsen: If we don't do aything else, I think we should document the exception property explicitely, I think the name of the Sphinx/Napoleon section is simply attribute.

soco/exceptions.py
```diff
@@ -80,3 +80,37 @@ class SoCoSlaveException(SoCoException):
...
+        self.__dict__ = kwargs
+        self.__dict__['exception'] = exception
+
+    def __getattr__(self, name):
+        raise self.exception
+
+    def __setattr__(self, name, value):
+        raise self.exception
+
+    def __getitem__(self, item):
```
KennethNielsen: Why not a __setitem__  ?

ghcs27:
I just found out that this does not really work as intended. It will take some time to repair it.

soco/events.py
ghcs27: I didn't want to log the traceback here, since the exception will be reraised. Instead, I wanted to give extra information about the tag and alert the user that a following exception might be caused by a SoCoFault.

soco/exceptions.py
ghcs27: That was actually a stub, it has a different implementation now.

soco/exceptions.py
ghcs27: Good idea.

soco/exceptions.py
ghcs27: I left it out since no SoCo data structure implements `__setitem__`, but I'll put it in for completeness' sake.

ghcs27:
I did a rework of this PR: Now the original exception is not directly rethrown, but instead wrapped in an EventParseException, together with information on the relevant tag and the faulty metadata.

KennethNielsen:
I'll do a second review

soco/exceptions.py
```diff
@@ -80,3 +80,60 @@ class SoCoSlaveException(SoCoException):
 
 class NotSupportedException(SoCoException):
     """Raised when something is not supported by the device"""
+
+
+class EventParseException(SoCoException):
+    """Raised when a parsing exception occurs during event handling."""
+
```
KennethNielsen: As with the other one, I think i would be a good idea to document the instance attributes. From the docs: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html it seems that the way to do it is to add this section to the class docstring:

```
Attributes:
    tag (str): The tag for which the exception occured
    metadata (str): The metadata which failed to parse
    __cause__ (Exception): The original exception
```

Why is cause marked with double underscores btw?

soco/exceptions.py
KennethNielsen: As above

KennethNielsen:
> I did a rework of this PR: Now the original exception is not directly rethrown, but instead wrapped in an EventParseException, together with information on the relevant tag and the faulty metadata.

Good idea.

soco/exceptions.py
ghcs27: `__cause__` is a special Exception attribute for chaining. The output looks like this:

```python
  File "C:/Users/Christian/.PyCharmCE2017.3/config/scratches/scratch_2.py", line 5, in <module>
    raise SoCoException('my_message')
soco.exceptions.SoCoException: my_message

The above exception was the direct cause of the following exception:

  File "C:/Users/Christian/.PyCharmCE2017.3/config/scratches/scratch_2.py", line 13, in <module>
    value.xyz
  File "C:\Users\Christian\PycharmProjects\SoCo\soco\exceptions.py", line 121, in __getattr__
    raise self.exception
soco.exceptions.EventParseException: Invalid metadata for 'my_tag'
```

ghcs27: Should I include `__cause__` in the attributes section anyways?

soco/events.py
KennethNielsen: Ok

soco/exceptions.py
KennethNielsen: Oh wow. I tried googling that and came up empty. Must have been weak google-foo. I'm unsure if that works in Python 2.7 also, but I guess if it doesn't, then it just wont do anything.

KennethNielsen:
Great. All good. I'll merge. Remember a comment in the release notes issue.

amelchio:
> This catches any exceptions thrown by from_didl_string during event handling.

It seems that it actually does not catch all exceptions, like [this one](https://github.com/SoCo/SoCo/blob/v0.14/soco/data_structures.py#L172) (playing Pandora, I think):

```
  File "event.py", line 19, in <module>
    x = soco.events.parse_event_xml(event)
  File "/private/tmp/SoCo/soco/events.py", line 143, in parse_event_xml
  File "/private/tmp/SoCo/soco/data_structures_entry.py", line 52, in from_didl_string
    item = cls.from_element(elt)
  File "/private/tmp/SoCo/soco/data_structures.py", line 492, in from_element
    DidlResource.from_element(res_elt))
  File "/private/tmp/SoCo/soco/data_structures.py", line 172, in from_element
    raise Exception('Could not create Resource from Element: '
Exception: Could not create Resource from Element: protocolInfo not found (required).
```

I think this is all you need but let me know if I should provide more information :)

ghcs27:
Thanks!
This problem arises because we only catch SoCoExceptions (to avoid catching things like TypeError), yet this exception is not a subclass even though it should be.

ghcs27:
PR is at #601.

PR #568, opened 2018-01-20
Delay object parsing exceptions, reraise on object retrieval

### LINKED ISSUE #566: Make parsing errors in events less fatal

ghcs27:
This catches any exceptions thrown by from_didl_string during event handling. Instead an ErrorDescriptor is created, which rethrows the exception when the user tries to retrieve it from the variable dict. This avoids an exception in the event thread, instead letting the user handle it when he needs the object. To make this possible, a DescriptorDict data structure is created, which handles descriptor items like descriptor attributes.
Compare with #567 for another implementation of the same proposal.
See #566 for information on why this is needed.
Closes #566, closes #567.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -28,6 +28,7 @@ helpful.
 
 from __future__ import unicode_literals
 
+import collections
 import sys
 import textwrap
 import warnings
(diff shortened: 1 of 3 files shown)


KennethNielsen:
I'll close this one in favor of #567

PR #569, opened 2018-01-23
Small addition to DidlFavorite docstring

ghcs27:
Added clarification on the roles of `DidlFavorite` and its `reference`.


KennethNielsen:
Merged. Thanks.

PR #573, opened 2018-02-02
Services extensions

ghcs27:
This extends the services module in two ways:

- Handle default value / allowed value range in iter_actions and add a `__str__` function to the resulting Actions, to make the output more readable

- Allow keyword arguments in send_command, so you can do 
  ```python
  renderingControl.SetRoomCalibrationStatus(RoomCalibrationEnabled=False)
  ```
  instead of the less pythonic
  ```python
  renderingControl.SetRoomCalibrationStatus([
      ('InstanceID', 0),
      ('RoomCalibrationEnabled', False)
  ])
  ```
  to switch off Trueplay (something we do not yet support directly).

Let me know what you think of these extensions.

### DIFF
--- a/soco/services.py
+++ b/soco/services.py
@@ -77,19 +77,21 @@ class Action(namedtuple('ActionBase', 'name, in_args, out_args')):
 class Argument(namedtuple('ArgumentBase', 'name, vartype')):
     """A UPnP Argument and its type."""
     def __str__(self):
-        argument, datatype = self.name, self.vartype.datatype
+        argument = self.name
         if self.vartype.default:
             argument = "{0}={1}".format(self.name, self.vartype.default)
-        if self.vartype.list:
-            datatype = '[{0}]'.format(', '.join(self.vartype.list))
-        if self.vartype.range:
-            datatype = '[{0}..{1}]'.format(self.vartype.range[0],
-                                           self.vartype.range[1])
-        return '{0}: {1}'.format(argument, datatype)
+        return '{0}: {1}'.format(argument, str(self.vartype))
 
 
-#: An argument type with default value and range.
-Vartype = namedtuple('Vartype', 'datatype, default, list, range')
+class Vartype(namedtuple('VartypeBase', 'datatype, default, list, range')):
+    """An argument type with default value and range."""
+    def __str__(self):
+        if self.list:
+            return '[{0}]'.format(', '.join(self.list))
+        if self.range:
+            return '[{0}..{1}]'.format(self.range[0], self.range[1])
+        return self.datatype
+
 
 # A shared cache for ZoneGroupState. Each zone has the same info, so when a
 # SoCo instance is asked for group info, we can cache it and return it when
@@ -154,6 +156,11 @@ class Service(object):
         #: a `TimedCache` with a default timeout=0.
         self.cache = Cache(default_timeout=0)
 
+        # Caching variables for actions and event_vars, will be filled when
+        # they are requested for the first time
+        self._actions = None
+        self._event_vars = None
+
         # From table 3.3 in
         # http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf
         # This list may not be complete, but should be good enough to be going
@@ -184,7 +191,7 @@ class Service(object):
             611: 'Invalid Control URL',
             612: 'No Such Session',
         }
-        self.DEFAULT_ARGS = OrderedDict()
+        self.DEFAULT_ARGS = {}
 
     def __getattr__(self, action):
         """Called when a method on the instance cannot be found.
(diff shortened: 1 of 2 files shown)


soco/services.py
```diff
@@ -283,6 +314,71 @@ def unwrap_arguments(xml_response):
...
+            list: a list of ``(name, value)`` tuples.
+
+        Raises:
+            `AttributeError`: If this service does not support the action.
+            `ValueError`: If the argument lists do not match the action
+                signature.
+        """
+        in_argdict2 = dict(in_arglist or ())
+
+        action = None
```
KennethNielsen: Variables from inside loops retain their value after the loop ends and the for loop has an else clause that gets executed if there was no `break`. Therefore:
```python
action = None
for act in self.actions:
    if act.name == action_name:
        action = act
        break
if action is None:
    raise AttributeError('Unknown Action: {0}'.format(action_name))
```
can be written as:
```python
# Find appropriate action
for action in self.actions:
    if action.name == action_name:
        break
else:
    raise AttributeError('Unknown Action: {0}'.format(action_name))
```
which gets you rid of that nastily abbreviated `act` placeholder, although the retain-value-after-loop-end is lesser known, so better attach a comment

soco/services.py
```diff
@@ -283,6 +314,71 @@ def unwrap_arguments(xml_response):
...
+
+        Returns:
+            list: a list of ``(name, value)`` tuples.
+
+        Raises:
+            `AttributeError`: If this service does not support the action.
+            `ValueError`: If the argument lists do not match the action
+                signature.
+        """
+        in_argdict2 = dict(in_arglist or ())
```
KennethNielsen: not a big fan of `foo_bar2`, how about `in_argdict_from_arglist`?

soco/services.py
```diff
@@ -336,14 +432,15 @@ def build_command(self, action, args=None):
         # is set over the network
         return (headers, body)
 
-    def send_command(self, action, args=None, cache=None, cache_timeout=None):
+    def send_command(self, action, args=None, cache=None, cache_timeout=None,
+                     **kwargs):
```
KennethNielsen: Well let's hope that none of the actions have a `cache` argument, or we will be in trouble ;)

soco/services.py
```diff
@@ -553,7 +690,15 @@ def iter_actions(self):
             statevars = srvStateTable.findall('{}stateVariable'.format(ns))
             for state in statevars:
                 name = state.findtext('{}name'.format(ns))
-                vartypes[name] = state.findtext('{}dataType'.format(ns))
+                datatype = state.findtext('{}dataType'.format(ns))
+                default = state.findtext('{}defaultValue'.format(ns))
+                vlist = [item.text for item in
```
KennethNielsen: although the rest of this method has lots of aggressively abbreviated variables, I'd probably go with `value_list` and `value_range`

KennethNielsen:
I have not yet tested, anyone else up for exercising this semi-large change in a sensitive place? @DPH ? Probably something about testing that the string formatting for the available actions looks good and trying to call a few actions directly on the service with the kwargs calling convention, preferably in both python 2 and 3.

DPH:
@KennethNielsen  hi - happy to test but won't be until next week. Would be good if others did too as this is a critical area.
Cheers David

KennethNielsen:
I will see about testing it myself also. @ghcs27 let me know as soon as you have addressed the comments and the failed checks.

soco/services.py
```diff
@@ -63,10 +65,33 @@
...
-Action = namedtuple('Action', 'name, in_args, out_args')
-#: A UPnP Argument and its type.
-Argument = namedtuple('Argument', 'name, vartype')
+
+class Action(namedtuple('ActionBase', 'name, in_args, out_args')):
+    """A UPnP Action and its arguments."""
+    def __str__(self):
+        args = ', '.join(str(arg) for arg in self.in_args)
+        returns = ', '.join(str(arg) for arg in self.out_args)
+        return '{0}({1}) -> {{{2}}}'.format(self.name, args, returns)
```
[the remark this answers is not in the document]
ghcs27: Yes, since the return value is a dict.

soco/services.py
```diff
@@ -161,6 +191,7 @@ def __init__(self, soco):
             611: 'Invalid Control URL',
             612: 'No Such Session',
         }
+        self.DEFAULT_ARGS = {}
```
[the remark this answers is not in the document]
ghcs27: It's very similar to UPNP_ERRORS. Should I change the capitalization?

soco/services.py
ghcs27: Thanks!

soco/services.py
ghcs27: The argument names are all capitalized, so no need to worry :)

soco/services.py
ghcs27: I'll go with `in_arglist_as_dict`.

soco/services.py
ghcs27: Okay, and I'll also decompress those list comprehensions a bit.

ghcs27:
Try out this script:
```python
from soco.services import *
service_classes = [AVTransport, ContentDirectory, DeviceProperties, RenderingControl, ZoneGroupTopology, AlarmClock, SystemProperties, MusicServices, Queue, GroupManagement, GroupRenderingControl]
for service_class in service_classes:
    service = service_class(player)
    print('== {} =='.format(service.service_type))
    for action in service.iter_actions():
        print(action)
    print()
```

Some example lines from the output:
```
BecomeCoordinatorOfStandaloneGroup(InstanceID: ui4) -> {DelegatedGroupCoordinatorID: string, NewGroupID: string}
GetMute(InstanceID: ui4, Channel: [Master, LF, RF, SpeakerOnly]) -> {CurrentMute: boolean}
SetPlayMode(InstanceID: ui4, NewPlayMode=NORMAL: [NORMAL, REPEAT_ALL, REPEAT_ONE, SHUFFLE_NOREPEAT, SHUFFLE, SHUFFLE_REPEAT_ONE]) -> {}
SetBass(InstanceID: ui4, DesiredBass: [-10..10]) -> {}
```

soco/services.py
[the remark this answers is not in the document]
KennethNielsen: Gotcha

soco/services.py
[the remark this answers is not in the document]
KennethNielsen: The allcaps are by naming convention used for constants. EPNU_ERRORS is a constant, but since this is initialized to an empty dict, I assume it is not

soco/services.py
KennethNielsen: Absolutely, "short, descriptive and to the point" to quote Linuxs Thorvalds on naming git branches, variables and functions

soco/services.py
KennethNielsen: :)

KennethNielsen:
@ghcs27 if you will comment on the DEFAULT_ARGS, then we are good to go review wise. I will test now.

soco/services.py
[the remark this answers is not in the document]
ghcs27: It _is_ constant as long as the object lives, but it's modified in the subclasses (e.g. [here](https://github.com/SoCo/SoCo/pull/573/files#diff-de75a829d8705a5dc2efa83021bffe7eR911)).
And we might want to add some defaults for convenience later.

KennethNielsen:
The text output description of the actions looks great. A though, the output for the return values looks like this:
```text
AddURIToQueue(InstanceID: ui4, EnqueuedURI: string, EnqueuedURIMetaData: string,
DesiredFirstTrackNumberEnqueued: ui4, EnqueueAsNext: boolean) ->
{FirstTrackNumberEnqueued: ui4, NumTracksAdded: ui4, NewQueueLength: ui4}
```
should we quote the names in the output, to make it look even close to the dict that it is? I.e:
```text
AddURIToQueue(InstanceID: ui4, EnqueuedURI: string, EnqueuedURIMetaData: string,
DesiredFirstTrackNumberEnqueued: ui4, EnqueueAsNext: boolean) ->
{"FirstTrackNumberEnqueued": ui4, "NumTracksAdded": ui4, "NewQueueLength": ui4}
```

soco/services.py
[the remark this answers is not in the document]
KennethNielsen: Ah. I missed that. Then I guess we are into the definition of what is considered a constant. My guess would be, that it is something that isn't changed after first defining it, since, in the languages that actually have constants, it isn't possible. Although they are only constants in the Python by naming convention, my guess would be that this definition applies and therefore DEFAULT_ARGS isn't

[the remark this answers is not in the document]
ghcs27: Well, UPNP_ERRORS is [modified by subclasses](https://github.com/SoCo/SoCo/pull/573/files#diff-de75a829d8705a5dc2efa83021bffe7eR888) as well.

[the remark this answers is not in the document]
KennethNielsen: Auch. Well, ok then. I stand by the definition above and think it is wrong for UPNP_ERRORS as well, but no reason to nit pick over that now. Let's go with it as is.

tests/test_services.py
```diff
@@ -147,6 +175,36 @@ def test_unwrap_invalid_char(service):
         "CurrentLEDState": "On",
         "Unicode": "AB"}
 
+
+def test_compose(service):
+    """Test argument composition."""
+    service.DEFAULT_ARGS = {}
+
```
KennethNielsen: Can you throw a few scattered comments is here about what it is that the different sections of test code tests. I mean, I can read it, if I scroll back and forth, between the test code and the test data, but I think it could use just a few small comments.

KennethNielsen:
@ghcs27 Awesome. Only remaining item is my comment above about "" in the return value for the __str__, if you can just comment (not necessarily change) then we are good to go.

ghcs27:
I agree that this would be closer to the real output, but I'd have to restructure the formatting code (since now, `Argument.__str__`, which formats the `argname: vartype` part, has no way to know if it's an `in_arg` or `out_arg`)
Should I do the change or just let it be?

KennethNielsen:
Ah. Just leave it then. Will merge later tonight.

KennethNielsen:
Ah. I can do it from the phone. Merged. Remember the release notes.

PR #576, opened 2018-02-18
Make build_album_art_full_uri public

amelchio:
I think this PR is wrong because obviously a bunch of breaking stuff wasn't just merged on the very day the next release was scheduled.

This makes things work again for my application but I did not have a lot of time to do testing.

CC: @ghcs27

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1214,8 +1214,8 @@ class SoCo(_SocoSingletonBase):
             album_art_url = metadata.findtext(
                 './/{urn:schemas-upnp-org:metadata-1-0/upnp/}albumArtURI')
             if album_art_url is not None:
-                track['album_art'] = self._build_album_art_full_uri(
-                    album_art_url)
+                track['album_art'] = \
+                    self.music_library.build_album_art_full_uri(album_art_url)
 
         return track
 
@@ -1580,7 +1580,7 @@ class SoCo(_SocoSingletonBase):
             item: The item to update the URI for
         """
         if getattr(item, 'album_art_uri', False):
-            item.album_art_uri = self._build_album_art_full_uri(
+            item.album_art_uri = self.music_library.build_album_art_full_uri(
                 item.album_art_uri)
 
     def create_sonos_playlist(self, title):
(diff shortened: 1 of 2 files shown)


ghcs27:
Oh, #573 should not have broken any code. What kinds of errors do you get?

ghcs27:
I see what the problem is: The current arguments to the calls don't match the signature given by iter_actions().
We should just revert #573 before the release, and then I'll look into how to make that PR backwards-compatible.

amelchio:
Yes, that was the error I got. Couldn't tell whether that was right or wrong.

Also note the other issue: `self._build_album_art_full_uri` is used by core, but that was removed in #542.

ghcs27:
Could you please remove the changes pertaining to #573? I'll make another PR for that problem.
But we should get the `self._build_album_art_full_uri` changes in.

amelchio:
Okay, I just pushed an updated branch with only the album art changes.

amelchio:
Is it okay to call an underscored method across modules?

KennethNielsen:
@amelchio looks good. Regarding the underscored method. In general no, I think we should promote that method to being public. Since underscored methods are considered internal implementational details, we wont have to worry about backwards compatibility.

KennethNielsen:
@amelchio do you want to add that change to this PR?

amelchio:
Well, I removed the underscores now, not sure if anything is needed documentation wise when making a method public?

Also, I rebased to exclude the service extension fixes that were reverted and I changed the title to make it fit the new content ... it may be better to just open a fresh PR?

Interestingly, `music_library` is calling `self.soco._update_album_art_to_full_uri` (i.e. with an underscore) ...

KennethNielsen:
@amelchio No need for changes documentation wise, if the doc string is clear.

You don't need to open a new PR. The history is clean enough and we can find our way around here.

> Interestingly, music_library is calling self.soco._update_album_art_to_full_uri (i.e. with an underscore) ...

You mean that music_library is calling _update_album_art_to_full_uri on a SoCo instance. That is a mistake and must have been missed during the migration. Probably the most important part of a change like this is grepping though the code to make sure to catch all calls to this methods, no-matter where it comes from.

amelchio:
This is included in #596 so I will close this PR now.

PR #577, opened 2018-02-18
Service extensions fix

ghcs27:
#573 unadvertently made some breaking changes: See the discussion for #576.
This ensures backward compatibility for old calls of `send_command`.
`compose_args` is now only called if `args` is None.


KennethNielsen:
Looks good. Merged.

PR #580, opened 2018-02-19
Update the feature list in the readme

ghcs27:
The development of SoCo has progressed a lot, but the readme page has not been kept up to date.
This PR adds some of the newer additions to README.rst in the 'Features' section.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -189,7 +189,8 @@ SoCo supports the following controls amongst others:
 -  Set the speaker’s Zone Name
 -  Start a music library update and determine if one is in progress
 
-SoCo also supports lower level access from Python to all Sonos services.
+SoCo also supports lower level access from Python to all Sonos services
+(e.g. ContentDirectory or RenderingControl).
 
 
 Related Projects
--- a/README.rst
+++ b/README.rst
@@ -134,13 +134,12 @@ Features
 
 SoCo supports the following controls amongst others:
 
--  Play
--  Pause
--  Stop
--  Next track
--  Previous track
--  Get current transport information(if speaker is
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -173,18 +187,9 @@ SoCo supports the following controls amongst others:
    -  MAC Address
 
 -  Set the speaker’s Zone Name
--  Find all the Sonos speakers in a network.
```
KennethNielsen: Did you miss this one when you re-factored, or is it on purpose, because of all the examples?

README.rst
```diff
@@ -173,18 +187,9 @@ SoCo supports the following controls amongst others:
...
--  Manage the Sonos queue (get the items in it, add to it, clear it,
-   play a specific song from it)
--  Get the saved favorite radio stations and shows (title and stream
-   URI)
--  Search for and play item from your music library
 -  Start a music library update and determine if one is in progress
 
-SoCo also supports lower level access from Python to all Sonos services (eg
-Alarms)
+SoCo also supports lower level access from Python to all Sonos services.
```
KennethNielsen: I actually think it was ok with an example here, but maybe better with something like: `... (e.g. ContentDirectory or RenderingControl)

ghcs27: Good idea.

README.rst
ghcs27: I left it out since all the other actions only work if you already have the SoCo instance, and the ability to discover speakers is already well-presented in the Examples section.

KennethNielsen: All right.

KennethNielsen:
Looks good. Merging. Remember the release notes.

Issue #397, opened 2016-05-23
set returned by `discover()` can mutate


leanto:
Recently, my program failed with the following error:

``` none
Traceback (most recent call last):
  File "./sonos-button.py", line 85, in <module>
    main()
  File "./sonos-button.py", line 52, in main
    coordinators = find_coordinators(zones)
  File "./sonos-button.py", line 79, in find_coordinators
    for zone in zones:
RuntimeError: Set changed size during iteration
```

This appears to have happened because `soco.discovery.discover()` does not return a static set, but rather [returns a reference to a set managed by a zone object](https://github.com/SoCo/SoCo/blob/22e787dad105ddb2814dfd4cd1d4d76e2349dc4a/soco/discovery.py#L177). This means that when the zone changes, the set's contents can change as well. And if the set's contents change in a separate thread due to an event, the above error can happen.

My suggestion would be to either change the documentation to make it clear that the returned set is dynamic and should be copied before iteration, or return a copy of the zone's set. (I have a slight preference for the documentation change since having a dynamic set is handy.)

KennethNielsen:
Thanks for reporting this and for the analysis and sorry about the delay getting back to you.

I think I tend to lean towards returning a copy, since I can't see a good use case for a changing zone set. If one needs to watch for changes in the topology, it is properly better to watch the events instead.

ghcs27:
I'm in favor of the copy as well.

KennethNielsen:
I've added the "Byte size" label, which indicates that it is a good bug for new contributors to start with. Spread the word.

KennethNielsen:
Closed by #582

PR #582, opened 2018-02-20
Fix mutating set from discover

ghcs27:
This is a bugfix for #397.
It makes all_groups, all_zones and visible_zones return copies of the current state, so the returned sets aren't modified later.


KennethNielsen:
Looks good. Closes #397. Merged. Thanks. Remember release notes.

PR #583, opened 2018-02-20
Initialize all DidlObject attributes to None

### LINKED ISSUE #356: Didl objects  - request for None valued attributes rather than missing attribute
With GIT nearly head 4b0926529a3ac139b183ff398e792f715e82d60b I was updating my playlist code to the latest changes and doing a speaker.browse() of a playlist I get back a list of `DidlMusicTrack` objects.

However, some of the tracks don't have albums which is a bit annoying:

    In [32]: r[0]
    Out[32]: <DidlMusicTrack 'Track 0' at 0x1045dddd0>
    
    In [33]: r[3]
    Out[33]: <DidlMusicTrack 'Track 3' at 0x1045f6350>
    
    In [34]: r[0].album
    Out[34]: u"Track 0 Title"
    
    In [35]: r[3].album
    ---------------------------------------------------------------------------
    AttributeError                            Traceback (most recent call last)
    <ipython-input-35-97099b9d313a> in <module>()
    ----> 1 r[3].album
    
    AttributeError: 'DidlMusicTrack' object has no attribute 'album'

Do I really have to do this for every attribute look up?

   album = item.album if hasattr(item, 'album') else None

or

   album = item.to_dict().get('album', None)

Seems to me the `_translation` dict could be used to form `None`-valued attributes inside the constructor for the `DidlObject` class near where `setattr(self, key, value)` is called.

ghcs27:
This is a PR for #356.
It initializes all allowed class-specific attributes for DidlObject to None.
Furthermore, it updates and restructures test_new_datastructures.py.
Closes #356.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -651,13 +651,15 @@ class DidlObject(with_metaclass(DidlMetaClass, object)):
 
         # Add the rest of the metadata attributes (i.e all those listed in
         # _translation) as sub-elements of the item element.
-        for key, value in self._translation.items():
-            if hasattr(self, key):
+        # pylint: disable=invalid-name
+        for key, (ns, tag) in self._translation.items():
+            value = getattr(self, key)
+            if value:
                 # Some attributes have a namespace of '', which means they
                 # are in the default namespace. We need to handle those
                 # carefully
-                tag = "%s:%s" % value if value[0] else "%s" % value[1]
-                XML.SubElement(elt, tag).text = ("%s" % getattr(self, key))
+                tag = "%s:%s" % (ns, tag) if ns else "%s" % tag
+                XML.SubElement(elt, tag).text = ("%s" % value)
         # Now add in the item class
         XML.SubElement(elt, 'upnp:class').text = self.item_class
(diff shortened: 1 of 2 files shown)


soco/data_structures.py
```diff
@@ -502,6 +506,8 @@ def from_element(cls, element):     # pylint: disable=R0914
             if result is not None:
                 # We store info as unicode internally.
                 content[key] = really_unicode(result)
+            else:
```
KennethNielsen: Isn't this redundant, since `content` turns into kwargs on `__init__` which is already being handled?

KennethNielsen (changes requested):
This seems like a good idea. A single comment.

EDIT. Oh, and the tests will need to be updated.

ghcs27:
Well, that escalated quickly.
This PR now includes a rework of test_new_datastructures.py, using fixtures and completely rebuilding assert_xml_equal.

pwt:
I took a run at bringing this up to date, but I don't understand the DIDL tests well enough to get them updated and working.

PR #587, opened 2018-02-21
Create music_source property for is_playing_...

### LINKED ISSUE #415: Universal attribute for is_playing_...
Having seperate methods for checking from which source music comes from seems redundant to me. In [this gist](https://gist.github.com/ghcs27/6096244f1eb6afae9beba157bb16f1be) I have written a possible alternative, which is one attribute that gives a string denoting the current music source.
Are there any possible inputs I am missing? Should I return None (of type NoneType) or "NONE" (of type String) if there is no music? Feedback is appreciated.
Another option would be incorporating this into the get_current_track_info() method - what do you think?

ghcs27:
This creates a music_source property, which checks the track URI for known prefixes and consequently returns a human-readable string like 'RADIO'.
Closes #415.


DPH:
Hi @ghcs27, great to see all your recent contributions.

I like the idea of this PR however I am trying to think how people would use it. 

There are a few more prefixes (full list I know of shown below) which is not always just the prefix - more the start of the uri.

URI prefix | Source
------------ | -------------
x-rincon-stream:  | Line In
x-sonos-htastream: | TV
x-file-cifs: | Local Library
x-rincon-mp3radio: | Radio
x-sonosapi-stream:| Radio
aac: | Radio
x-sonosapi-radio: | Radio
hls-radio: | Radio
http: | Track or Radio 
https: | Track or Radio
x-sonos-spotify: | Spotify
x-sonos-http: | Google (music library or mix)
x-sonos-http:library | Amazon (my music)
x-sonosapi-hls-static: | Amazon (tracks and radio stations)
x-sonos-http:ondemand_track | Napster tracks
x-sonosprog-http:station_track | Napster stations

On a Sonos controller music is shown with one of the following display formats and controls - I believe determined by the URI:

Controller format | Display
------------ | -------------
Radio format| Shows the name of the radio station and other available data. No seek, next, previous, or voting capability. Examples: TuneIn, radioPup
Smart Radio|  Shows track name, artist, and album. Limited seek, next  and sometimes voting capability depending on the Music Service.  Examples: Amazon Prime Stations, Pandora Radio Stations.
Track format| Shows track name, artist, and album the same as when playing from a queue. Full seek, next and previous capabilities. Examples: Spotify, Napster, Rhapsody.
Line in| with no controls
TV| with TV controls

If its purpose is to show what's displayed then we could restrict it to the 5 display types.

The other use is for stopping and starting or re-starting play - as in snapshot. Here the key things to know are is it being played from TV, Line_in, a Queue, a Cloud Queue (Alexa started queue!), or a Stream, or in fact a Slave to another player. You can't always get this information from the uri alone and have to use .avTransport.GetMediaInfo to get the URI of the music source. 

Sorry for the long ramble, I welcome thoughts on what we should be presenting as values, what these then mean, and how people can then use it. 

Cheers David

ghcs27:
Thanks for the URI prefix list, I'll add them to the `SOURCES` dict.
For the allowed actions, there's also the `AVTransport.GetCurrentTransportActions()`method (I'll make a PR for it soon).

> You can't always get this information from the uri alone and have to use .avTransport.GetMediaInfo to get the URI of the music source.
In which cases does that occur?

mountainsandcode:
Great work @ghcs27. Could you also add Airplay2 here?

Took a look and it looks as though it can be identified by URI in format
 `x-sonos-vli:RINCON_{identifier}:1,airplay:{another-identifier}`

Being to identify whether Sonos is playing from Airplay would be super valuable, for example because the volume levels are different

amelchio:
The prefix `x-sonosapi-hls:` was identified as `RADIO` in home-assistant/home-assistant#13397.

pwt:
I've resynced with master and fixed the conflicts and the minor wrinkles to pass the checks. A quick local test indicates that it works well.

@ghcs27: do you want to resurrect this feature?

soco/core.py
```diff
@@ -1217,29 +1215,43 @@ def switch_to_line_in(self, source=None):
...
+
+        *   ``'NONE'`` -- speaker has no music to play.
+        *   ``'LIBRARY'`` -- speaker is playing queued titles from the music
+            library.
+        *   ``'RADIO'`` -- speaker is playing radio.
+        *   ``'WEB_FILE'`` -- speaker is playing a music file via http/https.
+        *   ``'LINE_IN'`` -- speaker is playing music from the line-in at
+            another Sonos speaker.
+        *   ``'TV'`` -- speaker is playing input from TV.
+        *   ``'UNKNOWN'`` -- any other input.
```
amelchio, line 1244: This list is now missing `AIRPLAY`.

soco/core.py
```diff
@@ -2251,8 +2264,26 @@ def get_battery_info(self, timeout=3.0):
...
+    r"^x-rincon-mp3radio:": "RADIO",
+    r"^x-sonosapi-stream:": "RADIO",
+    r"^x-sonosapi-radio:": "RADIO",
+    r"^x-sonosapi-hls:": "RADIO",
+    r"^aac:": "RADIO",
+    r"^hls-radio:": "RADIO",
+    r"^https?:": "WEB_FILE",
+    r"^x-rincon-stream:": "LINE_IN",
+    r"^x-sonos-htastream:": "TV",
+    r"^x-sonos-vli:.*,airplay:": "AIRPLAY",
```
amelchio, line 2281: I think it would be good to define constants like `MUSIC_SOURCE_RADIO = "RADIO"` that clients could import to avoid typos.

soco/core.py
```diff
@@ -1217,29 +1215,43 @@ def switch_to_line_in(self, source=None):
...
-        return re.match(r"^x-rincon-stream:", track_uri) is not None
+        return self.music_source == "LINE_IN"
 
     @property
     def is_playing_tv(self):
         """bool: Is the playbar speaker input from TV?"""
+        return self.music_source == "TV"
+
+    @property
+    def music_source(self):
```
amelchio, line 1231: It would be useful to have a `uri` parameter that would override the `GetPositionInfo()` call. This would save a roundtrip when the information is already available, like when receiving an event.

That would also make it possible to decide the type of a favorite URI since one might want to handle each type differently.

amelchio (commented):
A few comments as I have been using something like this for a couple of years.

soco/core.py
pwt: Fixed.

soco/core.py
pwt: Done.

soco/core.py
amelchio: Well, if this one could be reworked it would not have to be additional.

I guess it comes down to the unanswered question from DPH above, how will it be used. I previously mentioned a case where the current implementation is not helpful: https://github.com/SoCo/SoCo/pull/532#issuecomment-370228251

pwt: OK, I've added a `music_source_from_uri()` method, and the `music_source` property implementation calls through to this.

amelchio: Awesome!

PR #559, opened 2018-01-06
Bug in from_didl_string causes incomplete information to be returned by soco.get_queue()

relevitt:
One line of code should be commented out until it has been adequately tested and is bug free.
Closes #552.

Issue #552, opened 2017-12-08
Pull request 455 breaks event object for Soundcloud


maru-sama:
With the new event handling functionality merged I updated my sonoscontroller to use the new version. Apparently this pull request changes the data returned in an avevent. I have the following code

`
metadata = event.current_track_meta_data 
metadata.album_art_uri
`
After the pull request the event is now returning a MSTrack instead of a DidlMusicTrack for metadata which only seems to contain the title. Is this intentional and should the MSTrack object contain the all the relevant information or is this a bug? Spotify on the other hand still returns a DidlMusicTrack.

maru-sama:
Ok, so apparently this is intentional. I had a closer look at it and attempt_data_structure update in data_structures_entry.py updates the item if it comes from it music service, for the moment it seems to only update the title attribute of the metadata and that's why album_art_uri is empty.

ghcs27:
Seems like the music service data structures need a lot of work.
I'd help out, but unfortunately I don't have any music service accounts, so I can't test it.

maru-sama:
SoundCloud is free of charge so you could use it for testing.

ghcs27:
Thanks for the tip!

DPH:
Hi,
I am happy to test on Spotify, Napster or Amazon
Cheers David

maru-sama:
Hmm, I have a brute force fix although I do not know if this will break other stuff
in data_structures_entry.py:131 where the metadata is passed to the cls, I replaced metadata with didl_item.__dict__. This should add all allowed metadata to the MSTrack object.
On the other hand the initial code went to great lenghts to specify the passed metadata so this might not be correct at all.

relevitt:
In soco.data_structures_entry.py, I have had to comment out the following line in the definition of from_didl_string:

line 53:             #item = attempt_datastructure_upgrade(item)

In my case, this line broke information for items from the "7Digital" musicservice. For me, there seems to be no downside in commenting out this line. Commenting out the line simply fixes what was broken.

As far as I can tell (with limited research), this line has caused bugs for a number of different soco users, with a variety of different use cases.

I haven't attempted to dig into the reasons why this line was added to the code (nor the related attempt_datastructure_upgrade function) nor where this is all going. But I am mystified why something that seemed to be working quite well before this change was broken by making a change which is half  baked and hasn't benefited from adequate (any?) testing.

The attempt_datastructure_upgrade function has the following comment:

        # Ignore other metadata for now, in future ask ms data
[...]

Issue #547, opened 2017-11-25
Not able to retrieve all values on streaming services


nlagaros:
I think think is particular to streaming services.  I've tested on Pandora and Spotify.  I can only retrieve the station name and track - not the creator/artist nor album. See the errors below in the debug log: "Update no possible, no resources".  I'm not sure this is the cause -- but it seems like all the content is in the raw event.

````
INFO:soco.events_base_twisted:Event 9 received for AVTransport service at 1511613207.81
[...]

PR #589, opened 2018-02-21
Copy metadata from DidlItem to MusicServiceItem in get_queue and events

### LINKED ISSUE #535: Queue items have changed
I recently updated to the latest version github version of SoCo and ran into this problem when trying to get information on items in the queue.

So using: `queue = master.get_queue()`
Previously the items in the the queue had properties that included album (=album title), title (=track title), and creator (=artist).  (note that version 0.11.1 supports this behavior.)

With the current version, there is a metadata property of the items in the queue but it appears to only contain the track title.

So my question is:  with whatever changes in data structures were made, is it possible to iterate through the items in the queue and get the album title, track title and artist?

Note that the items in the queue are coming from tracks stored in my amazon music account (not prime but tracks I own).

Thanks.

KennethNielsen:
This PR fixes an issue with a reduced amount of metadata for queue and
event items when upgrading them to their appropriate music service
type.

This should allow e.g. getting both title and album as properties and all remaining metadata from the DidlItem is available in the .metadata property of thr MusicService item.

Note. This PR is published as an alternative to #559. The reason for that is, that the music service data structures was added, since the music services use a completely different internal structure. In the long run, we want to be able to ask the music service to produce metadata for the MusicService items, including those coming from the queue. But until that has been implemented, we should preserve all the metadata we got.

Closes #535 and #559 and #547 and #552

### DIFF
--- a/soco/data_structures_entry.py
+++ b/soco/data_structures_entry.py
@@ -95,13 +95,12 @@ def attempt_datastructure_upgrade(didl_item):
         # matter what it is!
         item_id = '11111111{0}'.format(path)
 
-        # Ignore other metadata for now, in future ask ms data
-        # structure to upgrade metadata from the service
+        # Pass over all the available metadata in the metadata dict, in the
+        # future ask ms data structure to upgrade metadata from the service
         metadata = {}
-        try:
-            metadata['title'] = didl_item.title
-        except AttributeError:
-            pass
+        for key, value in didl_item.to_dict().items():
+            if key not in metadata:
+                metadata[key] = value
 
         # Get class
         try:


KennethNielsen:
Note: I seemed to remember that there were more issues pertaining to this, but I was unable to find them.

soco/data_structures_entry.py
```diff
@@ -23,7 +23,7 @@
 _LOG.debug('%s imported', __name__)
 
 
-def from_didl_string(string):
+def from_didl_string(string, music_service_medata_hack=False):
```
ghcs27: Is there a specific reason why you don't want to activate this by default?

soco/data_structures_entry.py
```diff
@@ -50,8 +50,20 @@ def from_didl_string(string):
...
-            item = attempt_datastructure_upgrade(item)
-            items.append(item)
+            upgraded_item = attempt_datastructure_upgrade(item)
+
+            # Until music service items are able to get their own metadata from
+            # the music services, make it possible to copy it over when we have
+            # some e.g. when retrieving the Queue
+            # THIS SHOULD BE REMOVE WHEN MUSIC SERVICE ITEMS AND MUSIC SERVICES
+            # ARE FIXED
+            if music_service_medata_hack and isinstance(upgraded_item,
```
ghcs27: Why not move this into attempt_datastructure_upgrade()?

KennethNielsen:
I will include a fix for #599 in this as well.

relevitt:
@KennethNielsen should I incorporate these changes into the work I'm doing on events?

maru-sama:
Hi,
Could this fix be merged please. I am currently using a monkey patched version but tend to forget about it only to be reminded when I work in a new virtualenv and wonder why there is no metadata in the events at all for certain music services.

KennethNielsen:
@relevitt I realize that I never answered you here. You never included this right, so I can continue with this PR without adding the functionality twice right?

relevitt:
@kennethnielsen, that's right, I didn't include it.

maru-sama:
@KennethNielsen Please do not forget to merge this for 0.18.

soco/data_structures_entry.py
KennethNielsen: I preparing a new version of this, so I wanted to answer the feedback. Yes, at the time I didn't know that our music services would continue to be broken and I really wanted the music services to provide the metadata. But since we don't know if and when the music services will re-appear I will now do this per default.

soco/data_structures_entry.py
KennethNielsen: Good idea.

KennethNielsen:
I re-implemented on top of current master an incorporated the feedback.

@maru-sama this is now ready for test.

maru-sama:
Hmm, 
when trying this I got another error, apparently this happens with master as well. When trying to subscribe to an event I get...

` File "/Users/maru/source/python/virtual/kivy3.7/lib/python3.7/site-packages/soco/services.py", line 602, in subscribe
     strict=strict)
   File "/Users/maru/source/python/virtual/kivy3.7/lib/python3.7/site-packages/soco/events.py", line 272, in subscribe
     return self._wrap(subscribe, strict, requested_timeout, auto_renew)
   File "/Users/maru/source/python/virtual/kivy3.7/lib/python3.7/site-packages/soco/events.py", line 399, in _wrap
     method(*args, **kwargs)
   File "/Users/maru/source/python/virtual/kivy3.7/lib/python3.7/site-packages/soco/events_base.py", line 428, in subscribe
     ip_address, port = self.event_listener.address
 ValueError: not enough values to unpack (expected 2, got 0)
`
This is happening with current master as well.....

relevitt:
Hi @maru-sama. I'm guessing the problem might be in this bit of code in events_base.py (starting at line 279):

        # Find our local network IP address which is accessible to the
        # Sonos net, see http://stackoverflow.com/q/166506
        with self._start_lock:
            if not self.is_running:
                # Use configured IP address if there is one, else detect
                # automatically.
                if config.EVENT_LISTENER_IP:
                    ip_address = config.EVENT_LISTENER_IP
                else:
                    temp_sock = socket.socket(socket.AF_INET,
                                              socket.SOCK_DGRAM)
                    try:
                        # doesn't have to be reachable
                        temp_sock.connect((any_zone.ip_address, 0))
                        ip_address = temp_sock.getsockname()[0]
                    except socket.error:
                        log.exception(
                            'Could not start Event Listener: check network.')
                        ip_address = None
                    finally:
                        temp_sock.close()
                if ip_address:  # Otherwise, no point trying to start server
                    # Check what port we actually got (twisted only)
                    port = self.listen(ip_address)
                    if port:
                        self.address = (ip_address, port)
                        self.is_running = True
                        log.info("Event Listener started")

Towards the end, you'll see the event server is started, and self.event_listener.address is set, only if ip_address could be determined. Any chance you could check to see if this is the problem? Failing that, could you send me the code where the service subscription takes place so I can see if it works at my end?

maru-sama:
Temp_sock.connect fails. Will try to see whats changed with the twisted merge in this place a little bit later on

maru-sama:
@relevitt The problem is
 `temp_sock.connect((any_zone.ip_address, 0))`
which does not work. 
Replacing this with 
`temp_sock.connect((any_zone.ip_address, config.EVENT_LISTENER_PORT))`
which is how the old code was doing it, makes it work.

relevitt:
@maru-sama Thank you - will do.

KennethNielsen:
@relevitt thanks for helping with the trouble shooting. I was just about to ask you for help just to find that you guys already solved :D

@maru-sama if you are able to test this PR with the hack in place, fell free to do so, otherwise I will rebase.

maru-sama:
@KennethNielsen Hi, I did a quick test and everything seems to work fine. During this I found another issue with Spotify direct, but this warrants a different issue.

KennethNielsen:
@maru-sama thanks. I will leave it open for a few more days to see if there are any objections or review and otherwise merge.

amelchio:
Does this fix #599 as well? An [earlier comment](https://github.com/SoCo/SoCo/pull/589#issuecomment-369196487) suggests that it will, but I see no related updates since then ...

KennethNielsen:
@amelchio auch, I think not. I will check up on it. Is #599 one of those things that you have a fix for in your fork or is it from scratch?

amelchio:
I still do not know how to reproduce it so I do not have a fix (I just downgraded the log severity).

PR #594, opened 2018-02-26
Create 'available_actions' SoCo property

ghcs27:
This PR introduces a get_available_actions() method to SoCo, which returns a list of strings like `['Set', 'Stop', 'Play']`. This can be used e.g. in a GUI application to disable the unavailable controls.
Internally, it simply uses AVTransport.GetCurrentTransportActions().


soco/core.py
```diff
@@ -1307,6 +1308,25 @@ def get_current_transport_info(self):
 
         return playstate
 
+    def get_available_actions(self):
```
KennethNielsen: This look great. I have just two quick questions related to the function definition.

1) Should this be a property. Python tends to favor them for trivial information items and I think this might be one of those.
2) Does this work on a slave or should it have the `only_on_master` decorator?

KennethNielsen:
Hi @ghcs27

I know that maintenance of SoCo is sporadic, so I can understand the delays. Anyway, this is a gentle nudge to hear if you are interested in, at some point, taking this to the point of merge.

pwt:
I've brought this PR up-to-date, resolved the review comments, fixed the docstrings, and added a unit test. Should be ready to merge.

PR #596, opened 2018-02-26
Fix what broke when the deprecated methods were removed

ghcs27:
I checked the changes I made in #542 and adjusted all calls to the now-removed methods. This is based on @amelchio's work on #576.
Closes #576.

PR #597, opened 2018-02-26
Create full_album_art_uri property for DidlItem

ghcs27:
Sonos uses a relative path for its album art uris. This means that the user has to call build_album_art_full_uri() (which is currently being moved) before invoking the URI. Some, but not all methods automatically upgrade the uri, creating an inconsistency.
This PR tries to solve this by adding a full_album_art_uri property to DidlItem. It upgrades the album_art_uri using a SoCo instance from discovery.any_soco(). It assumes that #576 or #596 get merged first.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -32,6 +32,7 @@ import sys
 import textwrap
 import warnings
 
+from .discovery import any_soco
 from .compat import with_metaclass
 from .exceptions import DIDLMetadataError
 from .utils import really_unicode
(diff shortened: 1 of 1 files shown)


ghcs27:
Should I go on and remove the automatic-upgrade methods? What kind of backwards-compatibility note should I leave (docstring / logginng)?

soco/data_structures.py
```diff
@@ -723,6 +724,16 @@ class DidlItem(DidlObject):
         }
     )
 
+    @property
+    def full_album_art_uri(self):
```
amelchio: It seems slightly inconsistent that this name and `library.build_album_art_full_uri` each have the "full" in different places.

I have no opinion about which one is better, though :)

PR #600, opened 2018-02-28
Add 'shuffle' and 'repeat' properties to the SoCo class

ghcs27:
To control repeat and shuffle, we currently have the `play_mode` property, which is a string combining both properties. To turn on shuffle, the a controller using SoCo has to interpret this string as a play mode, set the shuffle flag to True and reformat it to a string. To make this easier, I made `shuffle` and `repeat` properties which do this work. They use the `PLAY_MODES` dict, which maps each play mode string to a (shuffle, repeat) tuple.


soco/core.py
```diff
@@ -1573,16 +1603,6 @@ def __get_favorites(self, favorite_type, start=0, max_items=100):
 
         return result
 
-    def _update_album_art_to_full_uri(self, item):
```
amelchio: This cleanup seems unrelated.

ghcs27: Thanks, I mixed up my branches.

PR #601, opened 2018-03-03
Turn Exception in from_element and to_element into DIDLMetadataError

ghcs27:
This PR turns Exceptions and ValueErrors which are raised in from_didl_string into DIDLMetadataErrors, allowing them to be caught in the event handling code. See [#567](https://github.com/SoCo/SoCo/pull/567#issuecomment-370102539) for why this is necessary.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -160,7 +160,7 @@ class DidlResource(object):
                 try:
                     return int(result)
                 except ValueError:
-                    raise ValueError(
+                    raise DIDLMetadataError(
                         'Could not convert {0} to an integer'.format(name))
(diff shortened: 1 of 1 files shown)


amelchio:
Can we merge this for 0.16? Even if it is not the best solution, it is definitely better than the status quo.

(btw, not indicating that it isn't the best solution 😃)

ghcs27:
OK. If there are no objections, I will merge this next week.

amelchio:
@ghcs27 Just a friendly reminder about this one before the window closes.

ghcs27:
Thanks!

Issue #590, opened 2018-02-22
Stop the event listener if no subscriptions are active


ghcs27:
When the user subscribes to a service, the event server is started. When he unsubscribes again, the event server keeps running until the user calls `event_listener.stop()`, or as long as the interpreter runs.
If the user calls `event_listener.stop()`, there might still be other code which has a running subscription, which is then killed. This wouldn't be necessary if we'd just automatically stop the listener when the last subscription is cancelled (if `_subscriptions` is empty on unsubscribe).

relevitt:
I'll add this to the 0.15 version I'll be working on.

KennethNielsen:
@relevitt was this included in #602, so this can be closed?

PR #602, opened 2018-03-06
In events module: allow choice in how the event server is implemented (v2)

relevitt:
This is version 2, implementing suggestions by @KennethNielsen and @ghcs27. See #575 for discussion.

I propose to update this pull request to address #590 [DONE]. 

Let me know if there is anything else I should incorporate.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -2,225 +2,84 @@
 # pylint: disable=not-context-manager
 
 # NOTE: The pylint not-content-manager warning is disabled pending the fix of
-# a bug in pylint: https://github.com/PyCQA/pylint/issues/782
+# a bug in pylint. See https://github.com/PyCQA/pylint/issues/782
 
 # Disable while we have Python 2.x compatability
 # pylint: disable=useless-object-inheritance
 
 
-"""Classes to handle Sonos UPnP Events and Subscriptions."""
+"""Classes to handle Sonos UPnP Events and Subscriptions.
+
+The `Subscription` class from this module will be used in
+:py:mod:`soco.services` unless config.EVENTS_MODULE is set
+to point to :py:mod:`soco.events_twisted`, in which case
+:py:mod:`soco.events_twisted.Subscription` will be used.
+See the Example in :py:mod:`soco.events_twisted`.
+
+Example:
+
+    Run this code, and change your volume, tracks etc::
+
+        from __future__ import print_function
+        try:
+            from queue import Empty
+        except:  # Py2.7
+            from Queue import Empty
+
+        import logging
+        logging.basicConfig()
+        import soco
+        from pprint import pprint
+        from soco.events import event_listener
+        # pick a device at random and use it to get
+        # the group coordinator
+        device = soco.discover().pop().group.coordinator
+        print (device.player_name)
+        sub = device.renderingControl.subscribe()
+        sub2 = device.avTransport.subscribe()
+
+        while True:
+            try:
+                event = sub.events.get(timeout=0.5)
+                pprint (event.variables)
+            except Empty:
+                pass
+            try:
+                event = sub2.events.get(timeout=0.5)
+                pprint (event.variables)
+            except Empty:
+                pass
+
+            except KeyboardInterrupt:
+                sub.unsubscribe()
+                sub2.unsubscribe()
+                event_listener.stop()
+                break
+"""
 
 from __future__ import unicode_literals
 
-import atexit
 import logging
-import socket
 import threading
-import time
-import weakref
 
 import requests
 
-from . import config
 from .compat import (
-    Queue, BaseHTTPRequestHandler, URLError, socketserver, urlopen
+    BaseHTTPRequestHandler, URLError, socketserver, urlopen
 )
-from .data_structures_entry import from_didl_string
-from .exceptions import SoCoException, SoCoFault, EventParseException
-from .utils import camel_to_underscore
-from .xml import XML
+
+# Event is imported so that 'from events import Events' still works
+# pylint: disable=unused-import
+from .events_base import Event  # noqa: F401
+
+from .events_base import (
+    EventNotifyHandlerBase, EventListenerBase, SubscriptionBase,
+    SubscriptionsMap
+)
+
+from .exceptions import SoCoException
 
 log = logging.getLogger(__name__)  # pylint: disable=C0103
-
-
-# pylint: disable=too-many-branches
-def parse_event_xml(xml_event):
-    """Parse the body of a UPnP event.
-
-    Args:
-        xml_event (bytes): bytes containing the body of the event encoded
-            with utf-8.
-
-    Returns:
-        dict: A dict with keys representing the evented variables. The
-            relevant value will usually be a string representation of the
-            variable's value, but may on occasion be:
-
-            * a dict (eg when the volume changes, the value will itself be a
-              dict containing the volume for each channel:
-              :code:`{'Volume': {'LF': '100', 'RF': '100', 'Master': '36'}}`)
-            * an instance of a `DidlObject` subclass (eg if it represents
-              track metadata).
-            * a `SoCoFault` (if a variable contains illegal metadata)
-
-    Example:
-
-        Run this code, and change your volume, tracks etc::
-
-            from __future__ import print_function
-            try:
-                from queue import Empty
-            except:  # Py2.7
-                from Queue import Empty
-
-            import soco
-            from pprint import pprint
-            from soco.events import event_listener
-            # pick a device at random
-            device = soco.discover().pop()
-            print (device.player_name)
-            sub = device.renderingControl.subscribe()
-            sub2 = device.avTransport.subscribe()
-
-            while True:
-                try:
-                    event = sub.events.get(timeout=0.5)
-                    pprint (event.variables)
-                except Empty:
-                    pass
-                try:
-                    event = sub2.events.get(timeout=0.5)
-                    pprint (event.variables)
-                except Empty:
-                    pass
-
-                except KeyboardInterrupt:
-                    sub.unsubscribe()
-                    sub2.unsubscribe()
-                    event_listener.stop()
-                    break
-    """
-
-    result = {}
-    tree = XML.fromstring(xml_event)
-    # property values are just under the propertyset, which
-    # uses this namespace
-    properties = tree.findall(
-        '{urn:schemas-upnp-org:event-1-0}property')
-    for prop in properties:  # pylint: disable=too-many-nested-blocks
-        for variable in prop:
-            # Special handling for a LastChange event specially. For details on
-            # LastChange events, see
-            # http://upnp.org/specs/av/UPnP-av-RenderingControl-v1-Service.pdf
-            # and http://upnp.org/specs/av/UPnP-av-AVTransport-v1-Service.pdf
-            if variable.tag == "LastChange":
-                last_change_tree = XML.fromstring(
-                    variable.text.encode('utf-8'))
-                # We assume there is only one InstanceID tag. This is true for
-                # Sonos, as far as we know.
-                # InstanceID can be in one of two namespaces, depending on
-                # whether we are looking at an avTransport event, a
-                # renderingControl event, or a Queue event
-                # (there, it is named QueueID)
-                instance = last_change_tree.find(
-                    "{urn:schemas-upnp-org:metadata-1-0/AVT/}InstanceID")
-                if instance is None:
-                    instance = last_change_tree.find(
-                        "{urn:schemas-upnp-org:metadata-1-0/RCS/}InstanceID")
-                if instance is None:
-                    instance = last_change_tree.find(
-                        "{urn:schemas-sonos-com:metadata-1-0/Queue/}QueueID")
-                # Look at each variable within the LastChange event
-                for last_change_var in instance:
-                    tag = last_change_var.tag
-                    # Remove any namespaces from the tags
-                    if tag.startswith('{'):
-                        tag = tag.split('}', 1)[1]
-                    # Un-camel case it
-                    tag = camel_to_underscore(tag)
-                    # Now extract the relevant value for the variable.
-                    # The UPnP specs suggest that the value of any variable
-                    # evented via a LastChange Event will be in the 'val'
-                    # attribute, but audio related variables may also have a
-                    # 'channel' attribute. In addition, it seems that Sonos
-                    # sometimes uses a text value instead: see
-                    # http://forums.sonos.com/showthread.php?t=34663
-                    value = last_change_var.get('val')
-                    if value is None:
-                        value = last_change_var.text
-                    # If DIDL metadata is returned, convert it to a music
-                    # library data structure
-                    if value.startswith('<DIDL-Lite'):
-                        # Wrap any parsing exception in a SoCoFault, so the
-                        # user can handle it
-                        try:
-                            value = from_didl_string(value)[0]
-                        except SoCoException as original_exception:
-                            log.warning("Event contains illegal metadata"
-                                        "for '%s'.\n"
-                                        "Error message: '%s'\n"
-                                        "The result will be a SoCoFault.",
-                                        tag, str(original_exception))
-                            event_parse_exception = EventParseException(
-                                tag, value, original_exception
-                            )
-                            value = SoCoFault(event_parse_exception)
-                    channel = last_change_var.get('channel')
-                    if channel is not None:
-                        if result.get(tag) is None:
-                            result[tag] = {}
-                        result[tag][channel] = value
-                    else:
-                        result[tag] = value
-            else:
-                result[camel_to_underscore(variable.tag)] = variable.text
-    return result
-
-
-class Event(object):
-    """A read-only object representing a received event.
-
-    The values of the evented variables can be accessed via the ``variables``
-    dict, or as attributes on the instance itself. You should treat all
-    attributes as read-only.
-
-    Args:
-        sid (str): the subscription id.
-        seq (str): the event sequence number for that subscription.
-        timestamp (str): the time that the event was received (from Python's
-            `time.time` function).
-        service (str): the service which is subscribed to the event.
-        variables (dict, optional): contains the ``{names: values}`` of the
-            evented variables. Defaults to `None`. The values may be
-            `SoCoFault` objects if the metadata could not be parsed.
-
-    Raises:
-        AttributeError:  Not all attributes are returned with each event. An
-            `AttributeError` will be raised if you attempt to access as an
-            attribute a variable which was not returned in the event.
-
-    Example:
-
-        >>> print event.variables['transport_state']
-        'STOPPED'
-        >>> print event.transport_state
-        'STOPPED'
-
-    """
-    # pylint: disable=too-few-public-methods, too-many-arguments
-
-    def __init__(self, sid, seq, service, timestamp, variables=None):
-        # Initialisation has to be done like this, because __setattr__ is
-        # overridden, and will not allow direct setting of attributes
-        self.__dict__['sid'] = sid
-        self.__dict__['seq'] = seq
-        self.__dict__['timestamp'] = timestamp
-        self.__dict__['service'] = service
-        self.__dict__['variables'] = variables if variables is not None else {}
-
-    def __getattr__(self, name):
-        if name in self.variables:
-            return self.variables[name]
-        else:
-            raise AttributeError('No such attribute: %s' % name)
-
-    def __setattr__(self, name, value):
-        """Disable (most) attempts to set attributes.
-
-        This is not completely foolproof. It just acts as a warning! See
-        `object.__setattr__`.
-        """
-        raise TypeError('Event object does not support attribute assignment')
 
 
 class EventServer(socketserver.ThreadingMixIn, socketserver.TCPServer):
@@ -228,46 +87,34 @@ class EventServer(socketserver.ThreadingMixIn, socketserver.TCPServer):
     allow_reuse_address = True
 
 
-class EventNotifyHandler(BaseHTTPRequestHandler):
-    """Handles HTTP ``NOTIFY`` Verbs sent to the listener server."""
+class EventNotifyHandler(BaseHTTPRequestHandler, EventNotifyHandlerBase):
+    """Handles HTTP ``NOTIFY`` Verbs sent to the listener server.
+    Inherits from `soco.events_base.EventNotifyHandlerBase`.
+    """
+    def __init__(self, *args, **kwargs):
+        # The SubscriptionsMap instance created when this module is imported.
+        # This is referenced by soco.events_base.EventNotifyHandlerBase.
+        self.subscriptions_map = subscriptions_map
+        # super appears at the end of __init__, because
+        # BaseHTTPRequestHandler.__init__ does not return.
+        super(EventNotifyHandler, self).__init__(*args, **kwargs)
 
     def do_NOTIFY(self):  # pylint: disable=invalid-name
-        """Serve a ``NOTIFY`` request.
-
-        A ``NOTIFY`` request will be sent by a Sonos device when a state
-        variable changes. See the `UPnP Spec §4.3 [pdf]
-        <http://upnp.org/specs/arch/UPnP-arch
-        -DeviceArchitecture-v1.1.pdf>`_  for details.
+        """Serve a ``NOTIFY`` request by calling `handle_notification`
+        with the headers and content.
         """
-        timestamp = time.time()
         headers = requests.structures.CaseInsensitiveDict(self.headers)
-        seq = headers['seq']  # Event sequence number
-        sid = headers['sid']  # Event Subscription Identifier
         content_length = int(headers['content-length'])
         content = self.rfile.read(content_length)
-        # Find the relevant service and queue from the sid
-        with _subscriptions_lock:
-            subscription = _subscriptions.get(sid)
-        # It might have been removed by another thread
-        if subscription:
-            service = subscription.service
-            log.info(
-                "Event %s received for %s service on thread %s at %s", seq,
-                service.service_id, threading.current_thread(), timestamp)
-            log.debug("Event content: %s", content)
-            variables = parse_event_xml(content)
-            # Build the Event object
-            event = Event(sid, seq, service, timestamp, variables)
-            # pass the event details on to the service so it can update its
-            # cache.
-            # pylint: disable=protected-access
-            service._update_cache_on_event(event)
-            # Put the event on the queue
-            subscription.events.put(event)
-        else:
-            log.info("No service registered for %s", sid)
+        self.handle_notification(headers, content)
         self.send_response(200)
         self.end_headers()
+
+    # pylint: disable=no-self-use, missing-docstring
+    def log_event(self, seq, service_id, timestamp):
+        log.info(
+            "Event %s received for %s service on thread %s at %s", seq,
+            service_id, threading.current_thread(), timestamp)
 
     def log_message(self, fmt, *args):  # pylint: disable=arguments-differ
         # Divert standard webserver logging to the debug log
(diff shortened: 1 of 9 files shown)


soco/events_base.py
```diff
@@ -0,0 +1,642 @@
...
+        """
+        if not self.is_running:
+            return
+        self.is_running = False
+        self.stop_listening(self.address)
+        log.info("Event Listener stopped")
+
+    # pylint: disable=missing-docstring
+    def listen(self, ip_address):
+        pass
```
ghcs27: By convention, abstract methods should raise NotImplementedException.

soco/events_base.py
```diff
@@ -0,0 +1,642 @@
...
+        else:
+            callback = None
+        if callback and hasattr(callback, '__call__'):
+            callback(event)
+        else:
+            self.events.put(event)
+
+    # pylint: disable=missing-docstring
+    def auto_renew_start(self, interval):
+        pass
```
ghcs27: Couldn't this method be private (underscored)? It would help to know which methods get called from the outside and which are just subroutines.

relevitt:
@ghcs27 Thank you for the review! Much appreciated. Comments above.

soco/events_base.py
```diff
@@ -0,0 +1,642 @@
...
+                            result[tag] = {}
+                        result[tag][channel] = value
+                    else:
+                        result[tag] = value
+            else:
+                result[camel_to_underscore(variable.tag)] = variable.text
+    return result
+
+
+class Event(object):
```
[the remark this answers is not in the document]
relevitt: Will do.

soco/events_base.py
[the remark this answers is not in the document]
relevitt: Will do.

soco/events.py
```diff
@@ -299,87 +139,68 @@ def run(self):
...
 
-class EventListener(object):
+class EventListener(EventListenerBase):
     """The Event Listener.
 
     Runs an http server in a thread which is an endpoint for ``NOTIFY``
-    requests from Sonos devices.
-
+    requests from Sonos devices. Inherits from
+    `soco.events_base.EventNotifyHandlerBase`.
```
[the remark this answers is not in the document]
relevitt: Thank you. Well spotted.

soco/events_base.py
relevitt: Will do.

soco/events_base.py
relevitt: Will do.

soco/events_base.py
```diff
@@ -0,0 +1,642 @@
...
+        def success(*arg):
+            log.info(
+                "Unsubscribed from %s, sid: %s",
+                self.service.base_url + self.service.event_subscription_url,
+                self.sid)
+
+        # pylint: disable=missing-docstring
+        def failure():
+            log.warning(
+                "Error attempting to unsubscribe from %s, sid: %s",
```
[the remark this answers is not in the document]
relevitt: I'm trying to remember the different circumstances there have been unsubscribe failures. 

One was on attempts to unsubscribe when a subscription has timed out. This has been addressed, hopefully, by not attempting to unsubscribe if time_left is 0.

Another, I think, was (randomly?) when unsubscribe was invoked via atexit, but I'm not 100% sure.

I'll aim to do some more testing and see if I can replicate the failures. That should give us a better idea how best to deal with them.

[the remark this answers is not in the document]
relevitt: I've tested this further. The only scenarios in which I've been able to produce errors upon unsubscribe or auto_renew is upon network disconnects, now that the timeout problem has been addressed. 

Raising an exception on auto_renew seems pointless as there's no calling code to catch the error. Therefore I think it's appropriate to log a warning.

Raising an exception on subscribe does seem right, as the subscribe attempt will have failed and the calling code needs to deal with that.

On reflection, I think a warning is appropriate on unsubscribe for the following reasons:

- It's not quite true to say the unsubscribe attempt has failed (in the way a subscribe attempt would have failed). The subscription has been cancelled and deleted from the list of subscriptions. In that sense, it has been unsubscribed. It's just that the notification of 'UNSUBSCRIBE' to the sonos system failed, most likely because the network or the sonos system was down. It's hard to see what more the calling code could do about that, beyond what has already been done. It seems appropriate to warn about the problem through the logging system, rather than raise an exception.

- Unsubscribe is called automatically at exit / shutdown, where there is again no calling code to catch the exception. It also seems these exceptions at exit aren't reported by python. Query whether that means unsubscribe isn't being fully executed at exit.

My proposal therefore is to raise an exception on a subscribe failure, but to log a warning on an auto_renew or unsubscribe failure.

soco/events_base.py
```diff
@@ -0,0 +1,642 @@
...
+        Args:
+            requested_timeout (int, optional): The period for which a renewal
+                request should be made. If None (the default), use the timeout
+                requested on subscription.
+
+        """
+        # NB This code may be called from a separate thread when
+        # subscriptions are auto-renewed. Be careful to ensure thread-safety
+
+        log.info("Autorenewing subscription %s", self.sid)
```
ghcs27: The message '**Auto**renewing subscription %s' might not always be accurate. Just make it 'Renewing subscription %s'.

soco/events_base.py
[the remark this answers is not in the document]
ghcs27: Thanks for the research and clarification. Now I agree that subscribe should raise an error and unsubscribe should log a warning.
Originally, I misunderstood the code and thought that renew() would just log an error as well, but I see now that it throws an error. I think this is right because we can't take for granted that it will only ever be called by the auto renew thread.

While we're at it: What happens if the auto renewal fails? Currently, the exception remains uncaught, so the exception is written to stderr by the interpreter.

An alternative would be this (in `AutoRenewThread.run()`):
```python
def run(self):
    subscription = self.subscription
    stop_flag = self.stop_flag
    interval = self.interval
    try:
        while not stop_flag.wait(interval):
            log.info("Autorenewing subscription %s", sub.sid)
            subscription.renew()
    except SoCoException as exc:
        log.error(str(exc))

File: soco/events_base.py (hunk 2 above)
Comment (ghcs27) [reply soco/events_base.py]: Actually, I was just referring to the abstract Subscription methods `request`, `auto_renew_start` and `auto_renew_cancel`. These only get called by other methods within the same class, so there should be no problem making them private.

**Comment by relevitt** [comment]
@ghcs27 I think the commit I've just made addresses what we've discussed. Errors on autorenewal are handled with warnings, while those on other renewal are handled by raising an exception. I did this before I looked at your last alternative. We can go with that - I'd just need to look into how to achieve the same in twisted.

File: soco/events_base.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (ghcs27) [reply soco/events_base.py]: Your exception handling is good as well, the raise-catch-log pattern isn't everyone's taste.
But if you want to implement my version, here's how I would do it with twisted:
```python
    def _auto_renew_start(self, interval):
        """Starts the auto_renew loop."""
        def try_renew():
            log.info("Autorenewing subscription %s", self.sid)
            try:
                self.renew()
            except SoCoException as exc:
                log.error(str(exc))
                self._auto_renew_loop.stop()
        self._auto_renew_loop = task.LoopingCall(try_renew)
        # False means wait for the interval to elapse, rather than fire at once
        self._auto_renew_loop.start(interval, False)

File: soco/events.py [hunk 7]
```diff
@@ -299,87 +142,68 @@ def run(self):
             listener.handle_request()
```
ghcs27: I know you didn't touch this code, but could you please rename `listener` to `server` or someting, since it's an `EventServer` and not an `EventListener`.

soco/events.py
```diff
@@ -299,87 +142,68 @@ def run(self):
...
+        address = (ip_address, self.requested_port_number)
+        self._listener_thread = EventServerThread(address)
+        self._listener_thread.daemon = True
+        self._listener_thread.start()
+        return self.requested_port_number
+
+    def stop_listening(self, address):
+        """Stop the listener."""
         # Signal the thread to stop before handling the next request
         self._listener_thread.stop_flag.set()
```
ghcs27: Another suggestion: For the sake of encapsulation, could you create `stop()` methods for the threads, which just do this:
```python
def stop():
    self.stop_flag.set()
```

soco/events_base.py
```diff
@@ -0,0 +1,699 @@
...
+                "Error attempting to unsubscribe from %s, sid: %s",
+                self.service.base_url + self.service.event_subscription_url,
+                self.sid)
+
+        self._request(
+            'UNSUBSCRIBE',
+            self.service.base_url + self.service.event_subscription_url,
+            headers, success, failure)
+
+    def send_to_service(self, event):
```
ghcs27: Is there a reason why this method is called `send_to_service`, it doesn't use the service? I would have suggested `send_event` or something similiar.

soco/events_base.py
```diff
@@ -0,0 +1,699 @@
...
+        # pylint: disable=missing-docstring
+        def failure():
+            raise SoCoException(
+                "Could not subscribe to {}".format(
+                    service.base_url + service.event_subscription_url))
+
+        # Lock out EventNotifyHandler during registration.
+        # If events_twisted is used, this lock should always be
+        # available, since threading is not being used.
+        with self.subscriptions.subscriptions_lock:
```
ghcs27: Could you please add a comment like `This is to prevent the EventNotifyHandler from sending a notification before the subscription has been registered.`?

soco/events.py
```diff
@@ -388,261 +212,152 @@ def __init__(self, service, event_queue=None):
...
+            return
+
+        if failure:
+            failure()
 
-    def unsubscribe(self):
-        """Unsubscribe from the service's events.
 
-        Once unsubscribed, a Subscription instance should not be reused
+class Subscriptions(object):
```
ghcs27: `Subscriptions` doesn't seem like a good name to me, since it's easily misread as `Subscription` and doesn't make it clear that it works as a mapping. I'd propose `SubscriptionsMap'.

ghcs27: Couldn't you move `Subscriptions` to events_base.py? Using the lock wouldn't hurt even if there's only one thread, and the 'unsubscribe at exit' call could be moved to an abstract method (maybe on `Subscription`?)

soco/events_base.py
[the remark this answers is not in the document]
relevitt: Thank you! Twisted has a different approach to Exception handling. I think that must have been what drove my initial approach of logging errors rather than raising them. On reflection, as your comments show, that's not the right approach. I'm going to work through this next week to come up with something that handles errors better. 

In the Subscription class in events_twisted, the twisted method that sends HTTP requests (used by the subscribe, unsubscribe and renew methods) returns a [Deferred](https://twistedmatrix.com/documents/current/core/howto/defer-intro.html). If an Exception occurs inside a function that returns a Deferred, a [Failure](http://twistedmatrix.com/documents/current/api/twisted.python.failure.Failure.html) object is passed to the first errback that has been registered for that deferred with [Deferred.addErrback](https://twistedmatrix.com/documents/current/api/twisted.internet.defer.Deferred.addErrback.html).

I think the subscribe, unsubscribe and renew methods need to return a Deferred in the events_twisted version and to return None in the events version. Then, in the twisted version, the calling code can use addErrback to deal with failures.

Additionally, there should probably be a 'lenient' flag that, if set, will cause the subscribe, unsubscribe and renew methods to log Exceptions rather than raise them. The lenient flag will be used by internal calls. Autorenew is the example we have been discussing, but there's also automatic unsubscribe at exit.

I haven't yet thought through a solution to service.subscribe. This returns a Subscription object, after first calling its subscribe method. In the non-twisted version, if an Exception occurred during the HTTP request call in Subscription.subscribe, it could be caught and handled by the upstream code that called service.subscribe. 

In the twisted version, there would instead be an unhandled failure. The calling code receives a Subscription object, not a Deferred. Therefore, the calling code cannot register an errback to receive and handle the failure. One possibility would be to set a property of the Subscription object to refer to such a failure, so that the calling code could check that property after calling service.subscribe().

relevitt:
@ghcs27 Thank you! All comments look good. I'll incorporate in the next commits.

ghcs27:
I made a diagram of the events.py structure:
[EventsStructure.xml](https://www.draw.io/?lightbox=1&highlight=0000ff&edit=_blank&layers=1&nav=1&title=EventsStructure.xml#R7V3Zbts4FP0aP7bQvjwmbjodoNMpJgWmfTJki7bVKqIr0Vnm64eURFlcJDlabccp0NhX1HbP4eG95GU70%2BcPz3%2FE3m77F%2FRBONMU%2F3mmf5hpmuso%2BG9ieMkMlq5mhk0c%2BJmpZLgP%2FgO5MT9vsw98kDANEYQhCnascQWjCKwQY%2FPiGD6xzdYwZO%2B68zZAMNyvvFC0%2Fhv4aJtZbUM52D%2BBYLPN7%2Bzo%2BYGlt%2Fq1ieE%2Bym830%2FR1%2BpMdfvDopfL2ydbz4VPJpN%2FN9HkMIco%2BPTzPQUg8S72Wnfex4mjx2DGI0DEnGEZ2xqMX7gF95PTB0Av1BXmfXd4MxAg8yxDwlrS5Ij6DWrwZ5guADwDFL7hJfiHdzk%2FJuaJRLz8dPK8puW1b8rqb27wc7E1x6cMb4w%2F5S8sdQN%2BXcYAV4jvc%2BsEj4wjr957AcruEz%2B%2BS4L8g2sz0G4I5jH0Qv8Nm%2FAUfPzw5OaqSJzeLI2WCkMOYIMAgf4oWO8%2F3i2tru%2BfDjfGnTf47fcAlNdw9YlffgxjD820bA88vWsT8OdgfS4ktfVdq5fAnoAe4a9yEwSbCJgR3uJmXfwvBmpwFcat1mFJ5HRBv327RA%2F71QcUfExTDX2AOQxinl9Tn6Q8%2BQtpSewQjUNxfwrQyozLeipQahTKuQJl4Hwluw4%2BPGtzgpj9yN3D%2B5UF4CHyf3On2aRsgcL%2FzVuS2T1iMsS3lF%2FDpbfFBTKjP6YU%2BGNgSeksQ3qa05e66hhEqPaCS%2FpBrUJ0iXzahlyT55x2IA%2BxBEN9nd8nNXrzKRV1XumJKz2BVQhdFwpAArvUAOB27SoAnyIvRaUKuCJDrrSDvAzO7ETNVGwo0UdgTrFyLdeht3ifgCt5rwbPGBM8Q4AE%2BDtDyrzBGW7iBkRfeHayc6pVwBM8B%2Bp47l3z%2BQT6%2FN8m3CD%2FYd9qMfDkc%2BwkQeslFzNsjSAa54r6fIRkE09OqIACRf0MiUfK8OxB92wZRZv2Yjo8Z1JUMy96fvHQDfNhJcB%2BvACtVWJ42gALiymGOQeih4JG9QRfUbHda1DSTdfuKjFPBivG6%2BgpkBXQKbOu7XS1qqgiaJQHN6gpaeir2hPdSarCDQYSS0pW%2FEkMpFOcGWZNPHxraW7rJESZ7ggN9ilc5ilGmGGoJFBs7PTGcEdMTKrpnnZ7c4D72D4jA0xtJTsyKvjsKYcRY9Zqc9JCcVGE6eXJiyeLca3JyBGYTJieWLgHtmpy0B2%2FM5MS6JifNYS6N3Jg4V5PEuRXBV%2B%2FJiW5PH0qqOic5EtoasoGij8hAdy4glExnuj8HCcLhZHxhYWPG0NqwcTBymINomlqtaYqoaaeeuVMFKUsaxaMsaaY%2BkqQZYnJ4dqFfp55xdPA9WCxgyJYGcCAXphJFxO2MsBgokmtAT5hDGhE%2B07zKXhvZMyWRnDFWJEdvzvW5c%2BppA6jeqN1Glr7igecqe%2B1lb8xRS5OlAtJcSAoGi1xTiNt%2F2lT4ZYzIWBSbM82avkAUrF8%2BeRHuKfHFz8FnFJ8mmRLV0YeLL39%2F%2B%2FPjj8vWxezRacEkeZmKGfhOEJ6AgIpVk9u0Xy0i0skwAiiAJ7ro0iEgaQBkQjxMAY8QbhaAyN45oTBQJPLKjjRmJDlCxYjWWDLScjJ9Pj8%2BiaIRF1OrMxNzKDlQvadQFPFX%2Bb0x07WbUt2ujmaRb%2Bl2Q3S7OZLbdWt4urMYHL7%2BKMh%2F2tMNkvooqhJMUVtHyNqVR5ncTgVH4eDvWO5kD8GPmukoycriGfLDlCip3nU2qhVBTJcliKHU18%2Fx7XsnlDMtoc61IpNO6zOMmkRyLC44M%2FR6RvHtdaXfiky734XCKalRq4J0bH2vmAzbFZuy%2FytNbWflnPa1TFNpISBDtT6YJlLD4qbRLC6Azx4hP6uOY2bDhbIXFC7Ugm5umwWan%2FuHHW2R50%2BpKSORqrzRYdKWMM3qhWpdRc1uGCb59qbZ7zDptonL27GsLli%2FDJZpIssMZQqWqfy8hl3PMr49F7x1Z1mbofOqZXKWSUJ%2Ba5IAjQ%2Fhm1jGt%2Be22HRnWZsc8qplcpaZIsvMSVjmvm7A5NNQvWeO6ZMq2SnkDBzBXc2lhn7yAtsSqZfnCv0nBlw8bxnHJQZtmNNmQuKqTvJJCpEh7iS7RivSwcoRkNMyq19xKjLqypINuoR3qJxYYT%2BnRRFcJcUXsU4CiCYkmp5EExRNsWj6dWTNBTulIltsLC9wVm2Qq175FJfyJWuhHJ%2BLk6mDc7dmDWkPphWz9WuW9RVBwgyErCJIsmRp8jMVrRbP1GqCkRdlgKKMIgdIVRDI63YUWeHOrZeAEv7ZxSoIkC9jS2uxJFCV2JCCkcmYpgjg1CxpC7ByWzmrWNYD4KrB52YC3sUsVBlwfiRrh7dY5J3tZMav%2FrZhoSnyNLBYkm7IwREGpYpFqmYa6%2Fgj3VWhfzxkJT10FPJHhpC8Ym6AfkIzmCZA1F4AOWKL3VsDhO8howIiqUQUABl7z6OlSIZq2VitOj24wJJR8tzKd%2B%2F3y2QVB7usqPCiynTVqrWucdgh%2BZf8yL9TUjWq17nikOSNWzMoBurdHD99MacthtcLkm8vUmwW57bXZ6CiziYA%2BapOd0wENQHBJNOwJTjfvqUN1LfGrLe1xYiA7Vtntn34RDrXmBA6b32rvmQ6OPNI%2FQQqDYabFqvtsbbv2%2BKuhX101clKnRx1BBPnFrBE%2Ft6D5CqPr5fHMaNHRxUAelvy2FYLaTLEaKHVUQsb1pM6rCQ64lRG7zhrFwm0dNAbGGj5SmDR4WkYxdQqNzY3%2Bl04dNpUNVwUo%2FqMrCSlM6dBMndakvVUAPE2xyhHxqpJiks1vpi%2BYV%2BG0L7n4lJHTCYWKy9agXCRMNPH5xO89hKKcpGorFBguM3NbUKUNn3dbtyD5XvJthilph9jSpVbfYarTteNvK2koFiKoCNG%2FQDDN1e1fndo0TT5BFg3KdEYyvfFOlnsfBqssydmXZu0u3Ncc3Kk6zNcPg1ecYHK%2BMQS12gTjOYCwUUC4sdgdVYzpYwoidOm5PBXDyEQR2kbTRlqMlU1hwuE8NfD%2F2CZQX74T0L1u%2F8B)

relevitt:
@ghcs27 I have a few questions:

(1) There is nothing to stop a subscription from being subscribed multiple times. This in turn means multiple attempts to unsubscribe it automatically at exit will be registered and, if autorenew is set to true, multiple autorenew threads will be created. I was slightly surprised to discover this. However, perhaps it's generally OK. It seems the sonos system still sends only one copy of the relevant event to the event server, even if it receives a second redundant subscription request. Subscription.unsubscribe will fail silently if the Subscription has already been unsubscribed. Therefore, I propose to leave this behaviour mostly as it is, save that a subsequent call to Subscription.unsubscribe should stop the autorenew thread and if (a) autorenew was previously running or (b) autorenew is now requested, start autorenew  with the new timeout details. Does that sound right?

(2) What do you think should happen to a subscription if an HTTP 'SUBSCRIBE' request fails, whether on subscription or on autorenewal? Should it be cancelled (i.e. the _has_been_unsubscribed flag will be set to True)? If not, should autorenewal be stopped, if the failure occurred on an autorenew attempt, so that the Subscription doesn't continue attempting to autorenew indefinitely?

(3) A similar question arises where a SoCoException occurs on autorenewal. Presumably the  autorenew thread should be stopped?

ghcs27:
(1) Since we don't even allow a Subscription to be resubscribed after unsubscribe, I don't think we should allow subscribing multiple times. I would guess subscribing a second time will be a bug more often than a feature of the calling code, so it's better to let it fail.

(2) I will take a fail-fast stance on this as well: The (re)subscription has failed, which means that we might not receive events anymore. Therefore, I'd cancel the subscription (and raise an exception).

(3) Same as above, cancel the subscription and the autorenew thread with it.

soco/events_base.py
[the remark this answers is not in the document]
ghcs27: This is really interesting. The subscribe() and unsubscribe() methods return immediately, so afterwards 1) the subscription won't be finished yet and 2) you don't know if an exception will occur.
The twisted way is to return a Deferred and do 1) addCallback for when the subscription has finished and 2) addErrback for when it has thrown an exception.

Now the question is, what should service.subscribe do?

- Just return the Subscription (current behavior): User can't react to events. :x:
- Just return the Deferred: User doesn't get the Subscription. :x:
- Return a (Subscription, Deferred) tuple: Changes the method signature. :x:
- Return a Subscription with the Deferred as an attribute: Possible, but unintuitive (and probably cumbersome to implement) :white_check_mark:
- Return a Deferred with the Subscription as an attribute: Possible, but unintuitive :white_check_mark: 
- Return a Deferred which passes the Subscription to the callback: Intuitive, but user gets the Subscription only after subscribe call has returned :white_check_mark:
- Return a Deferred which passes the Subscription to the callback, and has the Subscription as an attribute: Like the above, but the user can get the Subscription directly if he wants to :heavy_check_mark:

So my favored option is the last one.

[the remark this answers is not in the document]
ghcs27: Some code:
```python
# events_base.SubscriptionBase
def subscribe(self, ...):
    ...

    def success(headers):
        ...

    with self.subscriptions.subscriptions_lock:
        return self._request('SUBSCRIBE', service.base_url + service.event_subscription_url,
            headers, success)
```

```python
# events.Subscription
def _request(self, method, url, headers, success=None, failure=None):
    try:
        response = requests.request(method, url, headers=headers)
        response.raise_for_status()
    except requests.exceptions.RequestException as exc:
        if failure:
            failure(exc)
    else:
        if success:
            success(response.headers)
        return self
```

```python
# events_twisted.Subscription
def _request(self, method, url, headers, success=None, failure=None):
    ...
    d = agent.request(*args)

    if success:
        def on_success(response):
            ...
            success(response_headers)
            return self
        d.addCallback(on_success)
    
    if failure:
        def on_failure(fail):
            failure(fail.value)
        d.addErrback(on_failure)

    d.subscription = self
    return d
```

```python
# services.Service
def subscribe(self, ...):
    subscription = config.EVENTS_MODULE.Subscription(
            self, event_queue)
    return subscription.subscribe(...)
```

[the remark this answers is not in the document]
relevitt: Looks good!

relevitt:
@ghcs27 I'm happy with (2) and (3). 

As to (1), having thought about it further, I think this may be too strict. The sonos system seems happy to receive subsequent 'SUBSCRIBE' requests and they have a purpose, because the timeout is updated each time a 'SUBSCRIBE' request is sent. Indeed, this is what happens upon autorenewal. A further 'SUBSCRIBE' request is sent to the sonos system when 15% of the requested duration is left, at which point it is reset to 100%.

ghcs27:
It is totally legitimate if you want to allow multiple calls to subscribe(). I'm a proponent of fail-fast systems, but that doen't mean you *have* to adjust your code for that.

relevitt:
@ghcs27 Thank you again for your help. You'll see multiple attempts to subscribe will fail, following your original suggestion. On reflection, I think that's the right outcome.

relevitt:
Please could anyone willing to do so help test this code? If you have an app using soco 0.14, you need only copy the following updated or new files from this PR into the directory with your soco python files and everything should work:

config.py
events.py
events_base.py
events_twisted.py
services.py

If you're willing to test events_twisted, all the better. There's an example showing how to do this at the top of the events_twisted file.

DPH:
Hi @relevitt  - well done on all the work in this. 
I can give it a test but not until next week if that helps
Cheers David

relevitt:
Hi @DPH - that would be great, yes please.

DPH:
Hi @relevitt - great work.
Have tested under both twisted and normal version of events on Python 2.7 and Python 3,6 on Windows and all works well. So looks ready to go.

A few observations - which may be to do with my understanding of Twisted:

-  installing twisted on python 2.7 is simple with pip. On 3.6 this doesn't work and I eventually installed it using the wheel (just followed on line advice but was not sure!) If this is the case we should include documentation (or a link to) how to install twisted on python 3. Or better still get Twisted to make pip work!

- Documentation would also be useful to guide people as to which version of events to use. What are the benefits / downsides with each?

- I also tried to use the twisted version in a thread, struggled passing parameters to the twisted callbacks, but even when I overcame this with a class, it appears twisted must be run in the main thread.
'builtins.ValueError: signal only works in main thread'  If this is true we should again document this so people don't get stuck. 

- finally it would be worth having a couple of examples in the examples directory - I know there are the basic examples in the code, but lets add them to the examples and maybe a few more to the examples to help people. I'm happy to do this if you want. 

Well done for your hard work and perseverance - I have been following the comments with great interest.
Cheers David

relevitt:
@DPH Great news that it worked and thank you for testing. I'm happy to do as you've suggested. What's the best place for documentation? Should I supplement docstrings in .py files or should this go somewhere else?

DPH:
@relevitt we have a doc directory so it may warrant a short document in there on events rather than overloading the code - I am not clear how documents in /doc get included in the online documentation  https://soco.readthedocs.io/en/latest/
@KennethNielsen can you advise on the right approach here please.
Cheers David

relevitt:
@DPH Thank you. I'v looked over the the documentation and what's in the soco/doc tree. There's already an "In depth" topics" section, with a subsection on "Events" that links to [events.rst](SoCo/doc/advanced/events.rst). Subject to other thoughts and comments, I propose to create two more rst files, one with additional examples and the other with more information on events_twisted, and link to them from [events.rst](SoCo/doc/advanced/events.rst).

relevitt:
@DPH I've updated the documentation as you suggested.

ghcs27:
Sorry for bothering you with this. I'm not sure using the _wrap methods is the right way to deal with exceptions and concurrency, since it makes the control flow harder to understand (especially when mixed with exceptions). The reason I'm bringing this up is that this code will have to be maintained, and you and I won't always be around to know how it works. I know this took quite a bit of work, but it doesn't seem right to me.
So what I'm asking: Is there another way to archieve a similiar result without the wrappers?

relevitt:
I'll have a look next week.

amelchio:
This seems a prime example of "perfect is the enemy of good".

Getting it merged would allow many more people to participate in moving it towards perfect.

stefankoegl:
We've been pushing this PR for two sprints now already. Can somebody summarize what needs to be done to get it merged?

relevitt:
@stefankoegl I think all that is needed is for me to update this PR to be compatible with release 0.17. I aim to do this over the next week or so, if I'm not out of time, and then I believe it can be included in 0.17. @ghcs27 has also raised a question, which is stylistic rather than functional. I'll have a look at that too, while updating the PR. If I'm too late for 0.17, please let me know, because then I'll skip 0.17 and go for 0.18.

relevitt:
@stefankoegl I've updated this PR to take account of commits since 0.15. It should be good to go now. Any questions, please let me know.

relevitt:
@stefankoegl The Travis build failed because of "W504 line break after binary operator", but some of these warnings relate to files I haven't changed, so I don't know what to do about it.

KennethNielsen:
Hi @relevitt 

Thanks for your work on the new version and for your incredible patience. 0.17 was just release, which means that is the time to merge large new features. I can see that @ghcs27 already reviewed and it seems all (or at least most) issues have been addressed, so now I think we should just get it merged and then we may refine it more later, when it is in the codebase and people start using.

This can still be merged cleanly, but the Travis build fails (I think due to some new lint version, which the rest of the codebase was updated against), so I would ask you if I can get you to rebase and fix any remaining lint issues in the code that is new to this PR and then I will merge immediately after. I know this is kind of short notice, but I will be working on SoCo tonight, so if you happen to have a little time we could get it merged by today.

If you would prefer, I can also do the rebase and remaining lint issue ;)

relevitt:
Hi @KennethNielsen, thanks you so much. I'll fix the lint issues as soon as I can this evening.

KennethNielsen:
@relevitt awesome, but remember, it should probably be done on top of a rebase, otherwise we are going to be in merge hell.

relevitt:
@KennethNielsen I think it's ready to go.

KennethNielsen:
Done. Thanks for you contribution and your patience. This change should already be mentioned in the release notes from earlier #665, but please check if the info there is correct, just in case.

PR #604, opened 2018-03-15
Event advertisement IP config

Juggels:
Recreated https://github.com/SoCo/SoCo/pull/450 to allow configuration of event advertisement IP.

### DIFF
--- a/soco/config.py
+++ b/soco/config.py
@@ -30,6 +30,15 @@ See also:
     The :mod:`soco.cache` module.
 """
 
++EVENT_ADVERTISE_IP = None
+"""The IP on which to advertise to Sonos.
+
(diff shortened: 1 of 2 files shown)


adamcstephens:
This is very useful for those running in docker (not using net=host).

stefankoegl:
Looks good :) If there are no objections, I'll merge in a few days.

adamcstephens:
Looking forward to this in fact. 

Without this I’m stuck with the home assistant polling fallback, which can take longer to trigger automations.

ghcs27:

ghcs27:

Juggels:
Thanks! I've added a post to the release notes thread.

PR #605, opened 2018-03-17
XML escape radio station titles in play_uri

amelchio:


amelchio:
I want to lose the "Needs info" label so I have added a description. However, that description does not actually contain information that was not already in the title. Please let me know if it is something else that you are looking for.

ghcs27:
Thank you, it helps to have the description split into problem and solution.

ghcs27:

ghcs27:

PR #606, opened 2018-03-23
Remove warning about missing track title

amelchio:
I think simply removing this warning is in line with the amount of logging generally used in soco.core but I can change it to merely downgrade the severity if you prefer?

Reported in home-assistant/home-assistant#13413: _Noticed when streaming from Amazon Music sources. On each track change, soco.core dumps warnings_


jjlawren:
This is a trivial change. Any way we can merge it in time for 0.15?

ghcs27:

ghcs27:

PR #639, opened 2018-09-24
Downgrade the severity of SoCoFault logging

amelchio:
It seems reasonable to keep the log quiet when users cannot do anything about the situation.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -146,11 +146,11 @@ def parse_event_xml(xml_event):
                         try:
                             value = from_didl_string(value)[0]
                         except SoCoException as original_exception:
-                            log.warning("Event contains illegal metadata"
-                                        "for '%s'.\n"
-                                        "Error message: '%s'\n"
-                                        "The result will be a SoCoFault.",
-                                        tag, str(original_exception))
+                            log.debug("Event contains illegal metadata"
+                                      "for '%s'.\n"
+                                      "Error message: '%s'\n"
+                                      "The result will be a SoCoFault.",
+                                      tag, str(original_exception))
                             event_parse_exception = EventParseException(
                                 tag, value, original_exception
                             )


ghcs27:
I believe the logging level for this was set in the hope that users might contact us with information about the warning, so we can adjust the parser.

amelchio:
Ah, I didn't realize that. Is this enough information for such an adjustment?

```
Event contains illegal metadatafor 'enqueued_transport_uri_meta_data'.
Error message: 'Unknown UPnP class: object.container.playlistContainer.tracklist'
The result will be a SoCoFault.
```
This is when playing via Spotify Direct.

ghcs27:
Well, it isn't enough to make an adjustment, but enough to know roughly what we're looking at. 
It seems like Spotify has created a `PlaylistContainer.tracklist` subclass, which isn't included in the Didl-Lite specification. It should probably be treated as a `PlaylistContainer`, like a `.#` unofficial subclass.
To be sure, could you please log the `value` variable in events.py line 154 and post the result here?

amelchio:
Sure:
```
<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/">
 <item id="" parentID="-1" restricted="true">
  <dc:title></dc:title>
  <upnp:class>object.container.playlistContainer.tracklist</upnp:class>
  <desc id="cdudn" nameSpace="urn:schemas-rinconnetworks-com:metadata-1-0/">SA_RINCON2311_X_#Svc2311-0-Token</desc>
  <upnp:albumArtURI></upnp:albumArtURI>
  <r:contentService id="9" name="Spotify"/>
 </item>
</DIDL-Lite>
```

ghcs27:
Well, it isn't enough to make an adjustment, but enough to know roughly what we're looking at. 
It seems like Spotify has created a `PlaylistContainer.tracklist` subclass, which isn't included in the Didl-Lite specification. It should probably be treated as a `PlaylistContainer`, like a `.#` unofficial subclass.
To be sure, could you please log the `value` variable in events.py line 154 and post the result here?

ghcs27:
Wow, that's some really peculiar metadata. It has an empty item id, negative parent id, empty title and a `r:contentService` subelement with only attributes (which we always ignore).

It should probably not be treated like a `playlistContainer`. Instead, we could create a `playlistContainerTracklist` class for this, or put the `playlistContainer.tracklist` Didl class on a list of classes to ignore.
What do you think?

amelchio:
Since we do not know how it should be used, it seems okay to ignore it.

This event always follows two `object.item.audioItem.musicTrack` events that contain the current/next track information.

amelchio:
Closing this in preference of #645.

PR #640, opened 2018-09-24
Remove logging of UPnP failures

amelchio:
This unconditional logging created noise for reasonable situations such as trying to skip the last track. Since we already raise an exception, just let the client decide how it should be handled.


ghcs27:
Agreed, we don't need to log anything if the client receives an exception anyway.

KennethNielsen:
Sorry about the wait. This looks good. Merged. Can I persuade you to add a note about the feature in the release notes issue #665.

KennethNielsen:
Sorry forgot to mention @amelchio. See above.

PR #644, opened 2018-10-31
Added household_id as optional parameter. 

twosky2000:
Use byte code b'Sonos_7O********************R7eU' from X-RINCON-HOUSEHOLD: Sonos_7O********************R7eU.

### DIFF
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -19,7 +19,11 @@ _LOG = logging.getLogger(__name__)
 # pylint: disable=too-many-locals, too-many-branches
 
 
-def discover(timeout=5, include_invisible=False, interface_addr=None):
+def discover(
+        timeout=5,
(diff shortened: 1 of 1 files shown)


twosky2000:
Sorry for breaking lint. I'm new and would like tips for making my programming style better.

pwt:
Note that #733 covers the underlying requirement here. It allows for multi-household discovery, and scanning the network for Sonos devices is the only way I know of finding all households.

Should we close this PR?

pwt:
I'm going to close this PR, as multi-households are quite well supported in SoCo 0.21 (to be released shortly).

Use the `network_scan` variants of the discover commands (with `multi_household=True`) to ensure that all households are discovered.

pwt:
Note: I've resurrected this feature in PR #939.

Issue #660, opened 2019-04-03
Group broken by Sonos version 10.1


rleeden:
I updated my Sonos to version 10.1 today and noticed that some of my scripts that rely on SoCo seem to have broken.

Seems to be a problem with groups. When trying to find coordinator of a group with

**device.group.coordinator.player_name**

I get error:

`Traceback (most recent call last):
  File "./test", line 7, in <module>
    print(device.group.coordinator.player_name)
AttributeError: 'NoneType' object has no attribute 'coordinator'`

and **print(device.group)** just gives output of:

`None`

I'm assuming this is the 10.1 upgrade. Does anybody else see this, and know if it is fixable?

pmatos:
Same here. `soco.discover()` returns the empty set.

rleeden:
It looks like the information is still there. Doing a tcpdump whilst running soco.discover() on my 4 speaker system shows this:

```
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
  <s:Body>
    <u:GetZoneGroupStateResponse xmlns:u="urn:schemas-upnp-org:service:ZoneGroupTopology:1">
      <ZoneGroupState>
        <ZoneGroupState>
          <ZoneGroups>
            <ZoneGroup Coordinator='RINCON_347E5C19199001400' ID='RINCON_B8E93799BB7A01400:3441'>
              <ZoneGroupMember UUID='RINCON_B8E93799BB7A01400' Location='http://192.168.0.66:1400/xml/device_description.xml' ZoneName='Living Room' Icon='x-rincon-roomicon:living' Configuration='1' SoftwareVersion='49.2-63270' MinCompatibleVersion='48.0-00000' LegacyCom
patibleVersion='36.0-00000' BootSeq='152' TVConfigurationError='0' HdmiCecAvailable='0' WirelessMode='1' WirelessLeafOnly='0' HasConfiguredSSID='1' ChannelFreq='2452' BehindWifiExtender='0' WifiEnabled='1' Orientation='0' RoomCalibrationState='4' SecureRegState='3' Voice
ConfigState='0' MicEnabled='0' AirPlayEnabled='0' IdleState='0' MoreInfo=''/>
[...]

PR #662, opened 2019-04-24
Fix for Issue 660 - Properly parse discover

EliSchleifer:
Properly parse discover return messages. Addresses 
https://github.com/SoCo/SoCo/issues/660

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -949,7 +949,7 @@ class SoCo(_SocoSingletonBase):
         self._all_zones.clear()
         self._visible_zones.clear()
         # Loop over each ZoneGroup Element
-        for group_element in tree.findall('ZoneGroup'):
+        for group_element in tree.find('ZoneGroups').findall('ZoneGroup'):
             coordinator_uid = group_element.attrib['Coordinator']
             group_uid = group_element.attrib['ID']
             group_coordinator = None


KennethNielsen:
Thanks @EliSchleifer for the patch and @rousseldenis for testing

Streetwalrus:
This change breaks discovery for me, maybe because my device isn't updated. I only use use the line input on my Connect:Amp and rely exclusively on SoCo to control and display the volume on the desktop computer it's connected to. I usually block the device from accessing the internet as well. I'll see if updating resolves the issue, but it's still something you might want to consider fixing.

```
>>> soco.discovery.by_name("La Grotte")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/site-packages/soco/discovery.py", line 226, in by_name
    for device in discover():
  File "/usr/lib/python3.7/site-packages/soco/discovery.py", line 185, in discover
    return zone.visible_zones
  File "/usr/lib/python3.7/site-packages/soco/core.py", line 1026, in visible_zones
    self._parse_zone_group_state()
  File "/usr/lib/python3.7/site-packages/soco/core.py", line 950, in _parse_zone_group_state
    for group_element in tree.find('ZoneGroups').findall('ZoneGroup'):
AttributeError: 'NoneType' object has no attribute 'findall'
```

Similar issue with socos's `list` command.

DPH:
Hi, good observation.

Typically with Sonos people update to the latest version, which is recommended.
The latest version of SoCo works with the latest Sonos. 
If you don't want to update Sonos then you could keep the previous version of SoCo. 

I don't think it is realistic to make the new version backward compatible.
Maybe we should warn people that this is not backward compatible.

Cheers David

Streetwalrus:
Yeah that's perfectly reasonable. I've already updated my device and that fixed it for me so I'll be fine. I know I'm not really using it in the intended way, but it's the only stereo amp I have around so I'll take it until I can replace it.

KennethNielsen:
OK, so a little more info. Sonos 10.1 chaned things in the API, which meant that SoCo stopped working. However, so far only two things **have been found** that were problems. One of the, around favorites, has a fix that **is** backwards compatible. The other, around discovery, has a fix that isn't **but it could be**. I only have limited time to work on SoCo, so I can only commit to try and fix things moving forwards, but it actually should be that difficult to make the discovery backwards compatible. So provided someone wants to put in the work, and attach the old code path with a proper deprecation warning, that I would merge such a PR. Basically ask git for a diff for the discovery function and there then there should be both the old and the new codepath to look at.

Streetwalrus:
Yeah, I understand where you're coming from. Preserving backwards compatibility takes effort and adds cruft to the code base. That said, I would argue that discovery is the bare minimum to keep working to an extent, so that the user would at least get the chance to know why parts of soco might break with their system after an update. I would gladly contribute, but this isn't important enough to me considering I've already updated my system.
At any rate, I don't really have a reason to stay on old firmware, it's just that I _never_ use the sonos app, it's not even installed on my phone and I had to download it to perform the upgrade this time.

PR #654, opened 2019-01-16
Travis fixes

amelchio:
This makes the build pass again.

Handles #647 by ignoring W504 until a decision is made.

PR #670, opened 2019-07-05
Remove pytest-cov upper bound

Infinisil:
Works just fine with 2.6.* and 2.7.* versions

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,5 +6,5 @@ graphviz
 flake8
 pylint >= 1.4
 coveralls
-pytest-cov>=2.4.0,<2.6
+pytest-cov>=2.4.0
 wheel


Infinisil:
Ping?

KennethNielsen:
Hi @Infinisil 

Looks good. Right now I have a little bit of a problem with a release, which means that I would rather not merge. Hopefully it will be fixed this weekend and I will pull it next week.

KennethNielsen:
Hmm. Ok, so I did a little digging. This upper restriction was introduced with commit 2b17ca0 in #654 to make sure that continuous integration passes. As we can see, the CI passes for this PR but I would like to ensure that the CI environment actually reflects this change and isn't just running in the old environment.

PR #679, opened 2019-10-20
Update CI environments after release of Python 3.8

KennethNielsen:
This is a simple PR to update the continuous integration environments after Python 3.8 was released. I also restricted the number of dev environments for stable Python versions to 2, to not over-use the ressource.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,14 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
-  - "3.5-dev"
-  - "3.6-dev"
+  - "3.8"
   - "3.7-dev"
   - "3.8-dev"
   - "nightly"
 
 matrix:
   allow_failures:
-    - python: "3.8-dev"
+    - python: "3.9-dev"
     - python: "nightly"
 
 cache:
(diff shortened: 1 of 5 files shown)


KennethNielsen:
Hmm. In the way XML is parsed and the written back out has changed, in a way that does seem to keep tag attribute order. Of course relying on tag attribute order is a bad idea to begin with, but as I remember it from way back, there really wasn't any good way in element tree to compare XML trees.

KennethNielsen:
I have updated the data_structure tests and made some changes in the old music service data structures. Should be ready for merge.

KennethNielsen:
@DPH don't know if you have time to review, otherwise don't sweat it.

DPH:
@KennethNielsen Hi.
Have done a quick test that the track info (title, artist, etc.) is still correct for Music Lib, Tune in radio, Amazon, Napster and Spotify tracks. All is as before. Not sure if this is what you need.
Sorry but I would not know how to test you tests!
Cheers David

KennethNielsen:
@DPH ahh sorry. I meant if you would look at the code and see if your could spot any obvious errors.

KennethNielsen:
Ok. I think this should be fine for merging.

Issue #674, opened 2019-10-17
Release notes 0.19


KennethNielsen:
Release notes for 0.19 (#673)

[Changes since 0.18](https://github.com/SoCo/SoCo/compare/v0.18...master)

KennethNielsen:
* Fixed broken link in loudness docstring #671

relevitt:
In events, fixed bug affecting some users in code to determine system's own IP address. Some systems require a valid port to be used (not port 0), so we use config.EVENT_LISTENER_PORT. #680

psjoholm:
Added class `DidlRecentShow` to implement service `object.item.audioItem.musicTrack.recentShow`. Used for podcasts etc. #677

KennethNielsen:
Change #680 mentioned by @relevitt has now also been released as a bugfix release named 0.18.1

KennethNielsen:
Add support for Python 3.8 #679

KennethNielsen:
* Copy metadata from DidlItem to MusicServiceItem in get_queue and events #589

pwt:
- Add methods list_library_shares() and delete_library_share() to MusicLibrary. Partially addresses #678.

pwt:
- Add a `balance` property the SoCo class, allowing get/set of speaker balance. Addresses #692. Credit to @tephlon for the idea and the majority of the implementation.

pwt:
- Add the `set_relative_volume()` method to the `SoCo` class.

pwt:
- Add unit test for `SoCo.MusicLibrary.delete_music_share()` (#694)

KennethNielsen:
* Add deprecation warning concerning the removal of Python 2.7 support #697

pwt:
- Add group volume operations (#688)

freand76:
Fixed a bug (avoid trying to iterate a None) in discovery.py c8e4a24

KennethNielsen:
Released

PR #680, opened 2019-10-20
Update events_base.py

relevitt:
Use config.EVENT_LISTENER_PORT in temp_sock.connect(). Fixes bug found in #589

### DIFF
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -289,7 +289,8 @@ class EventListenerBase(object):
                                               socket.SOCK_DGRAM)
                     try:
                         # doesn't have to be reachable
-                        temp_sock.connect((any_zone.ip_address, 0))
+                        temp_sock.connect((any_zone.ip_address,
+                                           config.EVENT_LISTENER_PORT))
                         ip_address = temp_sock.getsockname()[0]
                     except socket.error:
                         log.exception(


KennethNielsen:
Thanks @relevitt for a quick fix. Merged. Please add a comment to release notes thread #674 

Quick follow up question, as I didn't read into the problem. Does this only affect twisted events or also events in general?

relevitt:
@KennethNielsen, I'll add a comment to the release notes thread. The problem affected events in general, but I may have introduced the problem myself, tho' I can't remember for sure. There's an ancient hack, which SoCo uses, to determine the IP address of the system you're running on. I may have tried to simplify it slightly, but it was a long time ago and memory fades.

KennethNielsen:
Ok. Just as long as the code for that now is as it was. I think this is probably serious enough to do a bugfix release for. I will see if I can figure out how to do that.

relevitt:
@KennethNielsen, it doesn't affect everyone. For example, it doesn't affect me nor presumably anyone who tested the code. It may be just one or two. Not sure, though.

KennethNielsen:
@relevitt yeah, it didn't affect me neither. It will probably depend on how much trouble it is to make the extra release.

PR #683, opened 2019-11-09
Travis CI: Drop the EOL Python 3.4, add 3.9-dev

cclauss:
Also, flake8 the _whole_ repo including tests and setup.py

Reformat code with __psf/black__ on dev_tools, examples, tests to improve flake8 compatibility.


KennethNielsen:
@cclauss Wow, that's a huge chunk of clean-up. @stefankoegl  and I had been very recently discussing about using black, so it's great to see some interest in it. I would however prefer to have that change in a PR of its own to increase visibility of that change and to allow for additional changes along side it like adding a CI check. Is the black change constrained to a single commit here, so that it could easily be broken of into its own PR?

KennethNielsen:
@cclauss #685 merged. If the only thing left in this PR are black changes, I suggest on of two things; rewrite the history of this branch to omit the commits already present in #685 or (preferred) open a separate PR concerned only with black.

If there are more things left in this PR than black. I suggest to make a separate branch with black only and rewrite the history of this one to contain only the last minor changes changes except those already merged or black changes.

If you need help with the git let me know.

PR #686, opened 2019-11-11
WIP: Add method to create a SoCo object from a known Sonos device IP address

pwt:
This method allows instantiation of a SoCo object using a known Sonos device IP address. This is a convenience for cases where the normal discovery method is failing, for example when there is a network problem with multicast traffic.


pwt:
Ah, I see that one can already do this with the normal SoCo constructor.

PR #687, opened 2019-11-15
Adds the SoCo.set_relative_volume() method.

pwt:
This PR adds the `set_relative_volume()` method to the `SoCo` class.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -488,6 +488,12 @@ class SoCo(_SocoSingletonBase):
         volume to undershoot the minimum value of 0, the volume will be set
         to 0.
 
+        This method is an alternative to using increment and decrement
+        assignment operators (+=, -=) on the `volume` property of a `SoCo`
+        instance. These operators perform the same function as
+        `set_relative_volume()` but require two network calls per operation
+        instead of one.
+
         Args:
             relative_volume (int): The relative volume adjustment. Can be
                 positive or negative.
@@ -495,8 +501,6 @@ class SoCo(_SocoSingletonBase):
         Returns:
             int: The new volume setting.
         """
-        # Coerce to within the range -100 to +100
-        relative_volume = max(-100, min(relative_volume, 100))
         response = self.renderingControl.SetRelativeVolume([
             ('InstanceID', 0),
             ('Channel', 'Master'),
--- a/soco/core.py
+++ b/soco/core.py
@@ -133,6 +133,7 @@ class SoCo(_SocoSingletonBase):
         play_mode
         cross_fade
         ramp_to_volume
+        set_relative_volume
         get_current_track_info
         get_speaker_info
         get_current_transport_info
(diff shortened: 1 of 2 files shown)


pwt:
Note: I couldn't devise a meaningful unit test for this method. Suggestions welcome.

soco/core.py
```diff
@@ -479,6 +480,30 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
         return int(response['RampTime'])
 
+    def set_relative_volume(self, relative_volume):
+        """Adjust the volume up or down by a relative amount.
+
+        If the adjustment causes the volume to overshoot the maximum value
+        of 100, the volume will be set to 100. If the adjustment causes the
+        volume to undershoot the minimum value of 0, the volume will be set
+        to 0.
+
```
KennethNielsen: The SoCo instance has a `volume` property, which means that for relative changes you can do something like `device.volume += 10`, which has the same result as this BUT which requires 2 network calls. I think the documentation here should mention that option and the difference. You should probably check the other doc strings to see by what name a SoCo instance is usually referred to.

pwt: Thanks for the comments. I've added documentation to refer to the increment/decrement assignment operators.

pwt:
Note that I've removed the bounds check in the method, partly because this is not meaningful without knowing the current volume, and partly because Sonos handles it  at the receiving end and does the right thing.

Anyway, I've added a unit test patterned on other tests. However, since it won't actually catch any errors, I'm not sure of its value. Perhaps I'm missing something.

pwt:
I'm hoping this is ready to merge now, once a committer gets a chance.

soco/core.py
```diff
@@ -479,6 +480,36 @@ def ramp_to_volume(self, volume, ramp_type='SLEEP_TIMER_RAMP_TYPE'):
...
+        instead of one.
+
+        Args:
+            relative_volume (int): The relative volume adjustment. Can be
+                positive or negative.
+
+        Returns:
+            int: The new volume setting.
+        """
+        relative_volume = int(relative_volume)
```
KennethNielsen: Ok. So although this is hardly the standard for all code in soco (or even in Python in general), we might as well look to improve a little on the existing standard. In the Sphinx documentation-from-docstrings mechanics there is also the support for documenting exceptions that might be raised. I think this is a prime candidate, since the int call will fail with a `ValueError` if whatever you pass in cannot be converted to an int. So, as a last change I will suggest adding to the docstring, between the Args and the Returns, two lines like this:

```python
Raises:
    ValueError: If `relative_volume` cannot be cast as an integer
```

pwt:
I've made the docstring addition. Thanks for the comments on the unit tests. I'm starting to get a handle on how they're used in SoCo.

KennethNielsen:
Merged. Thanks for the contribution. Please add a comment to the release notes thread #674 about the change.

PR #688, opened 2019-11-18
Add group volume operations to SoCo

pwt:
This PR adds group `volume` and `mute` properties, and the group `set_relative_volume()` method, to the ZoneGroup class.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1070,6 +1070,73 @@ class SoCo(_SocoSingletonBase):
         zone_group_state_shared_cache.clear()
         self._parse_zone_group_state()
 
+    @property
+    def group_volume(self):
+        """The volume of the group of which this speaker is a member.
+
+        An integer between 0 and 100.
+        """
+        response = self.groupRenderingControl.GetGroupVolume([
+            ('InstanceID', 0)
+        ])
+        return int(response['CurrentVolume'])
+
+    @group_volume.setter
+    def group_volume(self, group_volume):
+        group_volume = int(group_volume)
+        group_volume = max(0, min(group_volume, 100))  # Coerce in range
+        self.groupRenderingControl.SetGroupVolume([
+            ('InstanceID', 0),
+            ('DesiredVolume', group_volume)
+        ])
+
+    @property
+    def group_mute(self):
+        """The group mute state for the group of which this speaker
+        is a member.
+
+        True or False.
+        """
+        response = self.groupRenderingControl.GetGroupMute([
+            ('InstanceID', 0)
+        ])
+        return bool(response['CurrentMute'])
+
+    @group_mute.setter
+    def group_mute(self, group_mute):
+        group_mute = bool(group_mute)
+        self.groupRenderingControl.SetGroupMute([
+            ('InstanceID', 0),
+            ('DesiredMute', group_mute)
+        ])
+
+    @only_on_master
+    def set_relative_group_volume(self, relative_group_volume):
+        """Adjust the group volume up or down by a relative amount.
+
+        Makes a relative adjustment for the group of which this speaker is
+        the coordinator.
+
+        If the adjustment causes the volume to overshoot the maximum value
+        of 100, the volume will be set to 100. If the adjustment causes the
+        volume to undershoot the minimum value of 0, the volume will be set
+        to 0.
+
+        Args:
+            relative_group_volume (int): The relative volume adjustment. Can be
+                positive or negative.
+
+        Returns:
+            int: The new group volume setting.
+        """
+        # Coerce to within the range -100 to +100
+        relative_group_volume = max(-100, min(relative_group_volume, 100))
+        response = self.groupRenderingControl.SetRelativeGroupVolume([
+            ('InstanceID', 0),
+            ('Adjustment', relative_group_volume)
+        ])
+        return int(response['NewVolume'])
+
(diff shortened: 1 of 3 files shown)


KennethNielsen:
Hi @pwt 

Looks interesting. A couple of questions before I do the actual review.

1) Have your tested how many of these work/doesn't work on a master vs slaver in a group. I can see that you added the `only_on_master` decorator on one of them, but is it accurate that the rest of the group functions can be called no problem on slaves?
2) Possibly a question for another PR. Do you have an overview over how many group methods there are? Since we didn't release any of your new methods yet, we still have option to change the API. What I was thinking is that the soco instance has a group property, which points to a ZoneGroup object, which has so far mainly been used for topology discovery. BUT that ZoneGroup object already has a reference to the master of the group, meaning that it would be real easy to implement the group methods there. That would then change the API from `device.group_volume` and `device.set_relative_group_volume` to `device.group.volume` and `device.group.set_relative_volume` which IMO is better, since it uses the notion that the device is part of a group represented by an object. This also completely removes the questions about `only_on_master` since the call would then always be redirected to the master of the group. I think I tend to think that we should pull it in that direction, but I'm fully open to the idea of merging your open PR's as are and then make the API change in a separate PR. What do you think?

pwt:
> Have your tested how many of these work/doesn't work on a master vs slaver in a group. I can see that you added the only_on_master decorator on one of them, but is it accurate that the rest of the group functions can be called no problem on slaves?

Yes, I've tested this.

> Do you have an overview over how many group methods there are?

That's all of them in the `GroupRenderingControl` category, except for `SnapshotGroupVolume` (which I don't understand).

> What I was thinking is that the soco instance has a group property, which points to a ZoneGroup object, which has so far mainly been used for topology discovery. 

I did think about using the `ZoneGroup` class for these methods, and I'm open to changing over to this approach. I just observed that the `SoCo` class already has a number of group management methods (while `ZoneGroup` doesn't do much), so these seemed to fit.

I'll have a play with dropping the methods into `ZoneGroup` instead.

soco/core.py
```diff
@@ -1122,15 +1122,19 @@ def set_relative_group_volume(self, relative_group_volume):
         volume to undershoot the minimum value of 0, the volume will be set
         to 0.
 
+        This method is an alternative to using increment and decrement
+        assignment operators (+=, -=) on the `group_volume` property of a
```
KennethNielsen: += and -= are in place add and subtract, not increment and decrement

pwt: True!

pwt:
The current commit moves the operations into the `ZoneGroup` class. Unit tests are still pending.

pwt:
This PR is now complete and ready for review.

tests/test_core.py
```diff
@@ -1149,23 +1149,23 @@ def test_group(self, moco_zgs):
         assert isinstance(moco_zgs.group, ZoneGroup)
         assert moco_zgs in moco_zgs.group
 
-    def test_all_zones(selfself, moco_zgs):
+    def test_all_zones(self, moco_zgs):
```
KennethNielsen: I think these other methods were as they were, with the `selfself` first argument to prevent some linting issue. It may be possible to handle that in another way, but let's do that separately. So please remove this and the next three single line changes that are not associated with tests for the new feature.

tests/test_core.py
```diff
@@ -1185,6 +1185,48 @@ def test_group_short_label(selfself, moco_zgs):
...
+        g.volume = 75
+        c.groupRenderingControl.SetGroupVolume.assert_called_once_with(
+            [('InstanceID', 0), ('DesiredVolume', 75)]
+        )
+
+    def test_group_mute(self, moco_zgs):
+        g = moco_zgs.group
+        c = moco_zgs.group.coordinator
+        c.groupRenderingControl.GetGroupMute.return_value = {
+            'CurrentMute': 0
```
KennethNielsen: I thought the bool endpoints returned the zeroes and ones as strings?

KennethNielsen:
@pwt I'm pushing for a feature freeze of 0.19 soon, preferably today and I would like to get this in. Please let me know if you will have any time to do the final polishing today.

pwt:
@KennethNielsen: Yes, I'll make the changes today.

KennethNielsen:
@pwt sorry about pushing for same day fix, when the feature freeze ended up being delayed anyway due to https://github.com/SoCo/SoCo/pull/698.

I will merge this now, thanks a lot for your contribution. Please add a comment about the change to the release notes thread #674.

One thing that dawned on me since the review is a structure thing concerning the tests. We always keep one test file per source code file, so at some point the tests you added should be moved into their own file, that concerns itself only with tests for groups. But since it is pure structure, I think we should just merge this now and then I can move the tests later.

pwt:
> @pwt sorry about pushing for same day fix, when the feature freeze ended up being delayed anyway due to #698.

That's fine @KennethNielsen. Thanks for the merge.

> One thing that dawned on me since the review is a structure thing concerning the tests. We always keep one test file per source code file, so at some point the tests you added should be moved into their own file, that concerns itself only with tests for groups. But since it is pure structure, I think we should just merge this now and then I can move the tests later.

I noticed this, but added my tests into `test_core.py` because that's where the other group tests currently reside. Should be easy to migrate them all into their own file.

PR #691, opened 2019-11-24
Avoid trying to iterate a None

freand76:
Fix iteration bug i discover.py


KennethNielsen:
@freand76 looks solid, thanks. Please add a note to the release notes thread #674 about the change

PR #695, opened 2019-12-27
Fixing error in index.py so it will run.

sho-luv:
There is an error when trying to run this code: TypeError: Unicode-objects must be encoded before hashing. I added the code to fix that issue.


KennethNielsen:
@sho-luv thanks for the fix. Please add a comment to the release notes thread #674 about the change.

PR #698, opened 2019-12-29
Beginning of data structure quirks, fix for missing protocolInfo

KennethNielsen:
For a long time it has annoyed me, that I couldn't find a compromise I was happy with, between keeping the implementation of DIDL-Lite in data_structures.py specification compliant and at the same time allowing service specific fixes for non spec-compliant behavior. Now I have a proposal for one.

As is usually the case, I didn't make it up, but simply borrowed a solution for cleverer people. I think it was in the realm of graphics card drivers or 3D library implementations for games that I first heard of the idea of quirks, as a way of (separately) implementing exceptions for specific hardware or games that behaves differently from what they are supposed to and I thought that would be a good solutions for us.

In the long run, the idea is to pull the DIDL-Lite implementation in data_structures.py back towards spec compliance and then specifying fixes for all non spec-compliant behavior in data_structure_quirks by manipulating the XML before it is processed. This has the added advantage of allowing us to easily track all the exception we have implemented, instead of them being scattered in the main implementation.

At current the module consist only of the `apply_resource_quirks` function with a crude fix for missing protocolInfo, but there will in the future also be a `apply_object_quirks` function to take care of that.

Let me know what you think of the idea?

### DIFF
--- a/soco/data_structure_quirks.py
+++ b/soco/data_structure_quirks.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+"""This module implements 'quirks' for the DIDL-Lite data structures
+
+A quirk, in this context, means that a specific music service does not follow
+a specific part of the DIDL-Lite specification. In order not to clutter the
+primary implementation of DIDL-Lite for SoCo (in :mod:`soco.data_structures`)
+up with all these service specific exception, they are implemented separately
+in this module. Besides from keeping the main implementation clean and
+following the specification, this has the added advantage of making it easier
+to track how many quiks are out there.
+
+The implementation of the quirks at this point is just a single function which
+applies quirks to the DIDL-Lite resources, with the options of adding one that
+applies them to DIDL-Lite objects.
+
+"""
+
+import logging
+
+
+_LOG = logging.getLogger(__name__)
+
+
+def apply_resource_quirks(resource):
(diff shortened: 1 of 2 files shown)


KennethNielsen:
I consider this a bug fix, so I will merge before 0.19 feature freeze, but I would love some testing on it concerning the missing protocolInfo

KennethNielsen:
Ok. So I would like to start progressing on a better way to solve broken data structures, so I will merge this as a bug fix, but I still welcome comments in this thread on the structure.

maru-sama:
Hi, I did a quick test and this commit partly fixes the issue I am having with spotify direct. To completely work the following change is required as well.

`        content['uri'] = element.text if element.text else ""
`
Apparently in the Sonos case the URI is empty as well, but is of course expected by the code. I tried adding this to the quirks section so it gets added but it did not work for some reason, so I did the change above.

maru-sama:
Well it works in quirks as well.

`if not resource.text:
            resource.text = ""`

and you workaround this one as well. What's apparently happening is that spotify direct apparently send totally bogus elements. Before the quirks they got ignored since protocol info was not set. With the quirks we work around this but since the text is emtpy we run into issue later on. So if we "fix" it in quirks be setting a dummy protocol info we also have to make sure that the text is set to at least an empty string.

PR #699, opened 2020-01-12
Fix quirks implementation

maru-sama:
If protocol_info gets set manually we also have to make sure that
resource.text is defined otherwise we run into issues later on.


KennethNielsen:
Thansk @maru-sama, merged. Fell free to add a comment in the release notes issue #709 about the change

KennethNielsen:
Ah wait, this was maybe a little fast. Can you comment a little about the context (which music service etc.) where this was observed to be an issue, then I will add a comment.

PR #706, opened 2020-03-13
Paint soco black

KennethNielsen:
Yeah. There is finally a PR to paint SoCo black.

black is "the uncomprimising code formatter" (https://github.com/psf/black) and its sole purpose is to format **all** code into a consistent style. black **is** uncompromising, in that it allows very little customization, but personally I find that I much prefer to have a codestyle tool that can't be modified, to debating which style is best. From my perspective, codestyle consistency is a hell of a lot more important that the itty bitty details of the style.

But perhaps the biggest advantage of black is that it almost removes code style discussions from reviews. I set it up so that there is a travisCI check that checks that the branch is black formatted. As those of that have gone through a review with me will know, I do care about style, but implementing it this way will remove the need to comment on codestyle to the point where I can just refer to the tool.

I painted both soco, the examples and the tests black. As an added advantage to this, black formatting the test code brought it within an hours of work to make it pass linting as well, which means that we now have linted test code as well :man_cartwheeling: 

Since this is code formatting, it touches just about all lines of code, which means that all other PRs will break, but I will help with that.

Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ script:
   - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
   - pylint soco
   - py.test --cov-config .coveragerc --cov soco .
-  - black --check soco tests
+  - if [ `python -c "from __future__ import print_function;import sys;print(int(sys.version_info.major==3 and sys.version_info.minor>=6 and sys.implementation.name == 'cpython'))"` -gt 0 ];then black --check soco tests;fi
 
 after_script:
   - coveralls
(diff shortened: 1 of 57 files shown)


KennethNielsen:
Ok. I need to wrestly with TravisCI a little more to get that to work (it is hard to test on my own computer). But the core of the PR is the same.

KennethNielsen:
Well, it only took 7 attempts ... but now the black checks on TravisCI are restricted to the Python versions that can run it. But **now** there is a functioning black check that will be executed when the pull request is made to check whether code was run through black.

Comments?

pwt:
LGTM.

KennethNielsen:
I will start with merging this, thus probably breaking all other PR's in the progress. I will help with the rebasing as needed.

PR #713, opened 2020-04-13
Autogenerate vendor extended DIDL-Lite classes

KennethNielsen:
Ok. So this has been brewing. We see quite a lot of vendor extended DIDL-Lite classes on a Sonos network. Every time a music service supplier created a new one, this would cause an error in SoCo. This PR adds the ability to autogenerate unknown extended classes.

I also managed to find a way to name the new classes that is mostly compatible with the way we have done so far, which actually means that the vendor extended classes:

DidlRecentShow
DidlComposer
DidlAlbumList
DidlSameArtist
DidlPlaylistContainerFavorite
DidlRadioShow

in data_structures.py are now redundant. I have however chosen not to delete them, to make sure not to cause a regression if a user imports one of these classes directly.

The autogenerated classes are simple extensions of their base class which means the only item that gets changes is the ``item_class``. If the new class add attributes or has a non-standard content tag, then it will still be necessary to create the class by hand. But in all other cases, this should now work and that also means that we will not add anymore vendor extended classes to data_structures.py unless it is necessary for one of the two reasons mentioned above.

Comments and reviews welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: python
 python:
-  - "pypy"  # Run slowest first so as not to hold up the build
   - "pypy3"
-  - "2.7"
   - "3.5"
   - "3.6"
   - "3.7"
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -37,7 +37,7 @@ import warnings
 
 from .compat import with_metaclass
 from .exceptions import DIDLMetadataError
-from .utils import really_unicode
+from .utils import really_unicode, first_cap
 from .xml import XML, ns_tag
 from .data_structure_quirks import apply_resource_quirks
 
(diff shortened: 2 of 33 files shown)


KennethNielsen:
Decided to remove the Python 2.7 checks from TravisCI as well, since the support will be disappearing from SoCo any second now.

KennethNielsen:
Ok. So given that there has been no objections to the implementation and that this will fix lots of issues I will merge it. Please test as much as possible.

amelchio:
I have a favorite like this:

```
<DIDL-Lite xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/">
  <item id="10092020urn%3adr%3amu%3achannel%3a4f3b8911860d9a33ccfdaf41" parentID="10fe2064live%3adistrict%3ap4" restricted="true">
    <dc:title>DR P4 Fyn</dc:title>
    <upnp:class>object.item.audioItem.audioBroadcast.#channel</upnp:class>
    <desc id="cdudn" nameSpace="urn:schemas-rinconnetworks-com:metadata-1-0/">SA_RINCON74247_</desc>
  </item>
</DIDL-Lite>
```

When I try to use it like this:
```
python -c 'import soco; print(soco.SoCo("10.0.2.14").music_library.get_sonos_favorites()[9].reference.resources)
```

I get this error:
```
soco.exceptions.DIDLMetadataError: UPnP class is incorrect. Expected 'object.item.audioItem.audioBroadcast.channel', got 'object.item.audioItem.audioBroadcast'
```

I have found this to be due to different handling of `.#`:

https://github.com/SoCo/SoCo/blob/508cb1e7c1a6adf58a7bb96e0af0107e78e64b2a/soco/data_structures.py#L84-L86

https://github.com/SoCo/SoCo/blob/508cb1e7c1a6adf58a7bb96e0af0107e78e64b2a/soco/data_structures.py#L568-L571

However, I am not sure what exactly is happening or what the right path forward would be?

Issue #709, opened 2020-03-22
Release notes 0.20


KennethNielsen:
Release notes for 0.20 (#708)

[Changes since 0.19](https://github.com/SoCo/SoCo/compare/v0.19...master)

IMPORTANT: When you write your release note addition and you mention an issue or pull request with the # syntax, please write before whether it is an issue or a pull request. It help me later writing the release notes.

KennethNielsen:
Format all soco main, test and example code with the black code formatter and make it mandatory going forward including a TravisCI check. Pull request #706.

pwt:
Add support for creating and separating stereo pairs of speakers. Note: works with dissimilar Sonos speakers if required. PR #704.

KennethNielsen:
Fix bug where data_structures_upgrade would fail on items that has no uri #702

KennethNielsen:
* Add support for autogenerating vendor extended DIDL-Lite classes (PR #713). This should fix all the problems we have with SoCo crashing if some vendor spedific data type is unkonwn.

pwt:
- Process share browsing correctly #717.

iainp999:
Categorise Sonos Amp as a playbar in order to provide support for 'night sound' and 'speech enhancement' https://github.com/SoCo/SoCo/pull/721

pwt:
Improve test_remove_playlist_bad_id() to handle the case of no existing playlists.
PR #726, fixes issue #725.

bdraco:
If port 1400 is in use, the next available 100 ports will be tried.  #724

ahummel25:
Is the pairing functionality from #704 not merged in yet? I did an install of the latest version (0.19) and those two functions aren't in `core.py`.

pwt:
It's merged for inclusion in 0.20, but the 0.20 release seems to be stuck.

ahummel25:
Any idea what the hold up is?

pwt:
> Any idea what the hold up is?

No, sorry. I asked in Gitter (https://gitter.im/SoCo/SoCo) about three weeks ago, but no responses so far.

KennethNielsen:
0.20 is finally out. Thanks to everyone who contributed.

PR #719, opened 2020-05-12
Fixes #717

### LINKED ISSUE #717: Music Library get_music_library_information processes 'share' browsing incorrectly
Using
`.get_music_library_information("share", search_term='//nas/my folder')`
sends
`<ObjectID>S::%2F%2Fnas%2Fmy%20folder</ObjectID>`
This fails on 2 counts:
1. The double colon
2. Escaping the /

What should be sent is either
`<ObjectID>S://nas/my%20folder</ObjectID>`
or
`<ObjectID>S://nas/my folder</ObjectID>`
Both work

My working code for get_music_library_information in music_library.py  is
```
from .compat import quote_url
...        
        search = self.SEARCH_TRANSLATION[search_type]
        share_translation = self.SEARCH_TRANSLATION['share']
        # Add sub categories NB sub categories not allowed when searching shares
        if subcategories is not None and search_term != share_translation:
            for category in subcategories:
                search += '/' + url_escape_path(really_unicode(category))
        # Add fuzzy search
        if search_term is not None:
            if search == share_translation:
                # Don't insert ":" and don't escape "/" (so can't use url_escape_path)
                search += quote_url(really_unicode(search_term).encode('utf-8'))
            else:
                search += ':' + url_escape_path(really_unicode(search_term))

```
FWIW both of the following work but are messy
```
.browse(ml_item=type('obj', (object,), {'item_id' : "S://nas/my folder"}))
.browse(ml_item=type('obj', (object,), {'item_id' : "S://nas/my%20folder"}))

```
I didn't have the confidence to do a pull so I was advised to post here

Thanks
Mike
[edit]
FWIW a Wireshark of the Android app shows it does not escape anything in the ObjectID when accessing shares (e.g. <ObjectID>S://nas/my folder</ObjectID>)

pwt:
This address issue #717 using the code supplied by @Sonosy (modified for Black compliance). I also had to fix a pylint 'disable' directive to pass pylint tests.

Integration tests run successfully.

### DIFF
--- a/soco/music_library.py
+++ b/soco/music_library.py
@@ -19,6 +19,7 @@ from .data_structures import SearchResult, DidlResource, DidlObject, DidlMusicAl
 from .data_structures_entry import from_didl_string
 from .exceptions import SoCoUPnPException
 from .utils import url_escape_path, really_unicode, camel_to_underscore
+from .compat import quote_url
 
 _LOG = logging.getLogger(__name__)
 
@@ -179,8 +180,7 @@ class MusicLibrary(object):
         args = tuple(["radio_shows"] + list(args))
         return self.get_music_library_information(*args, **kwargs)
 
-        # pylint: disable=too-many-locals, too-many-arguments,
-        # too-many-branches
+        # pylint: disable=too-many-locals, too-many-arguments, too-many-branches
 
     def get_music_library_information(
         self,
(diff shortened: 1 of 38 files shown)


soco/music_library.py
```diff
@@ -288,14 +288,20 @@ def get_music_library_information(
              `SoCoException` upon errors.
         """
         search = self.SEARCH_TRANSLATION[search_type]
+        share_translation = self.SEARCH_TRANSLATION["share"]
 
         # Add sub categories
-        if subcategories is not None:
+        # sub categories are not allowed when searching shares
+        if subcategories is not None and search_term != share_translation:
```
KennethNielsen, line 295: I **think** something is wrong here. `search_term` is a word or network path or similar, so I don't think that should be compared to `share_translation`. I think the intention must have been (I'm aware that you are implementing it from the suggestions in the issue) it should have read `search != share_translation`.

But actually, I would prefer doing these tests differently. It's a little bit the long way around to introduce a new variable and compare to the search, when really what determines the changes in behavior is the `search_type`. So I would prefer it if you would remove the extra variable and rewrite the test here and below against `search_type`.

Sonosy: I also think you're right with suggesting
`if subcategories is not None and search != share_translation:`
My original proposal
`if subcategories is not None and search_term != share_translation:`
was wrong
FWIW I introduced the share_translation variable to make it readable (for me!).
It could easily be
`if subcategories is not None and search != "S:":`
or, if you want to use search_type
`if subcategories is not None and search_type != "share":`
From your comment I presume you favour the last option
I'm happy with that

pwt: Modded as per comments.

KennethNielsen: @Sonosy great that we got that cleared up. Yes, I favor the latter, but that is because it is already established behavior. The `search_type` exactly describes which search type we are doing, which is what you are looking for, and it's option is described in the doc string. In any case, thanks a lot for the information and the fix. We will get it merged in no time.

KennethNielsen:
Thanks @Sonosy and @pwt for typing it up.

KennethNielsen:
Please drop a line describing the change in the release notes thread #709

pwt:
#709 Updated.

PR #720, opened 2020-05-12
Fixes required to pass updated pylint checks

pwt:
Pylint has been updated and is picking up a couple of new issues, hence all test builds will fail. This PR fixes the issues.


KennethNielsen:
Thanks, merged

Issue #708, opened 2020-03-22
Release 0.20


KennethNielsen:
Summary

Previous release: 0.19 (#673)
Release notes: #709
Milestone: [0.20](https://github.com/SoCo/SoCo/milestone/16)
Feature Freeze: evening of June 18th, 2020 (see #581)
Release date: July 2nd, 2020

As something new this release I would like to ask for nominations for PR's or bugs to be in this release. There is simply to many PR's to do all of them, so I want to make sure I start with those that have active interest. Write your nominations below, **one PR/issue nomination per comment** and upvote the ones that you care about if they have already been nominated.

I will then add them to the milestone.

pwt:
I'd like to see #704 (stereo pairing and unpairing methods) in 0.20. The PR is now up to date with, and compliant with, the Black reformatting changes.

tagdara:
I vote for #713 of course as a fix to all of the downstream DIDL issues.  The crashes from these simple metadata issues also tend to stop subscriptions and greatly reduce reliability in long-running applications.

KennethNielsen:
I updated the feature freeze and release dates to reflect reality :|

KennethNielsen:
Last PR merged. Version 0.20 is now feature frozen. I will push the release date slightly to make room for testing. As always I'd like to get:

1) a couple of people to sign of on the unit and functional tests passing on their machine (and if we can we should get preferably 2 operation systems involved).
2) Confirmation from as many people that we can that it works for their application

Happy testing

KennethNielsen:
I might end the testing period as much as a week before if we get appropriate testing.

pwt:
Unit and integration tests all pass on:
- macOS (10.14.6), Python 3.7.7
- Raspbian (Buster), Python 3.7.3

Tests were conducted against Sonos S1 (11.2) and Sonos S2 (12.0) devices.

KennethNielsen:
Would anyone care to test another Python version or check in that their SoCo using application works?

KennethNielsen:
0.20 is out. Thanks to everyone who contributed.

PR #724, opened 2020-06-05
Make events match events_twisted behavior when port 1400 is unavailable.

bdraco:
Merged downstream as https://github.com/amelchio/pysonos/pull/60

Resolves https://github.com/home-assistant/core/issues/36183

If port 1400 is unavailable, try the next 100 ports
in order to match the behavior in events_twisted.

When multiple Home Assistant instances are running on
the same server, sonos was falling back to polling.  As
support is already available for trying the next port
in events_twisted, this functionality is now extended
to events.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -59,6 +59,7 @@ Example:
 
 from __future__ import unicode_literals
 
+import errno
 import logging
 import threading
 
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@bdraco hi, this looks great. We have some other problems with our Travis builds, but it seems like this PR is triggering an error in one of the checks that does work, the black check.

We use black to format all code consistently and check it via Travis before merging. Can I persuade you run black on all the files you changed and push a commit with the changes?

KennethNielsen:
The TravisCI problems mentioned earlier has been fixed which means that now the checks that fail are show problems correctly.

bdraco:
Repushed with `black`

KennethNielsen:
@bdraco thanks! Merged. Please add a comment about to the change to the release notes thread #708

PR #730, opened 2020-06-15
Update core.py to recognize "arc" as a soundbar

KurtSanders:
Add the value "arc" to SOUNDBARS for "soundbar product names"


KennethNielsen:
Thanks. Please add a comment to the release notes thread #709 about the change

PR #656, opened 2019-02-14
Make zone-group-state cache per-household

amelchio:
I am trying to enable multi-household control. A blocker for this is that we currently maintain a global state cache that is invalid when a speaker from a different household is accessed.

This PR removes the global variable and makes the cache per-household.

I moved 10.0.2.14 ("Test") to a separate household. Before this PR:
```
$ python -c 'import soco; a = soco.SoCo("10.0.2.18"); b = soco.SoCo("10.0.2.14"); print(a.player_name); print(b.player_name);'
Kitchen
None
```

After this PR:
```
$ python -c 'import soco; a = soco.SoCo("10.0.2.18"); b = soco.SoCo("10.0.2.14"); print(a.player_name); print(b.player_name);'
Kitchen
Test
```

PR #733, opened 2020-06-28
Network Scan Discovery

pwt:
This is a proof-of-concept to add a new discovery function `scan_network()` that detects Sonos devices by scanning all IPs on the networks to which the SoCo host is attached. I've added the function to the existing `discovery.py` file, but it is standalone and not integrated with the other discovery functions.

This is to circumvent situations where the network is not correctly passing multicast traffic, meaning that the normal discovery functions do not work.

It appears to work on Windows, macOS, and Linux hosts, although my testing has necessarily been restricted to a small number of networks. I'd welcome other people trying it out.

**Edit: The following no longer applies**: _(Like the existing discovery functions, this function is vulnerable to the situation where there are multiple Sonos households/systems operating on the same network, e.g., in the case of an S1/S2 split system. It will return data on the first household that it discovers, ignoring any others.)_

### DIFF
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -54,7 +54,7 @@ def discover(
             devices.
         **network_scan_kwargs: Arguments for the
             :function:`scan_network` function. See its docstring for
-            documentation.
+            details.
     Returns:
         set: a set of `SoCo` instances, one for each zone found, or else
             `None`.
@@ -276,6 +276,64 @@ def by_name(name, allow_network_scan=False, **network_scan_kwargs):
     return None
 
 
+def _is_ipv4_address(ip_address):
+    """Helper function to test for an IPv4 address.
+
+    Args:
+        ip_address (str): The IP address to be tested, e.g.,
+            "192.168.1.35".
+
+    Returns:
+        bool: True if this is a well-formed IP address.
+    """
+    try:
+        ipaddress.IPv4Network(ip_address)
+        return True
+    except ValueError:
+        return False
+
+
+def _find_ipv4_networks(min_netmask):
+    """Discover attached IP networks.
+
+    Helper function to return a set of IPv4 networks to which
+    this node is attached. Excludes non-private and loopback
+    networks.
+
+    Args:
+        min_netmask(int): The minimum netmask to be used.
+
+    Returns:
+        set: A set of `ipaddress.ip_network` instances.
+    """
+
+    ipv4_net_list = set()
+    adapters = ifaddr.get_adapters()
+    for adapter in adapters:
+        for ifaddr_network in adapter.ips:
+            if _is_ipv4_address(ifaddr_network.ip):
+                ipv4_network = ipaddress.ip_network(ifaddr_network.ip)
+                # Restrict to private networks and exclude loopback
+                if ipv4_network.is_private and not ipv4_network.is_loopback:
+                    # Constrain the size of network that will be searched
+                    netmask = ifaddr_network.network_prefix
+                    if netmask < min_netmask:
+                        _LOG.debug(
+                            "%s: Constraining netmask from %d to %d",
+                            ifaddr_network.ip,
+                            ifaddr_network.network_prefix,
+                            min_netmask,
+                        )
+                        netmask = min_netmask
+                    network = ipaddress.ip_network(
+                        ifaddr_network.ip + "/" + str(netmask),
+                        False,
+                    )
+                    ipv4_net_list.add(network)
+    _LOG.info("Set of networks to search: %s", str(ipv4_net_list))
+    return ipv4_net_list
+
+
 # pylint: disable=too-many-statements
 def scan_network(
     include_invisible=False, max_threads=256, scan_timeout=0.1, min_netmask=24
@@ -298,7 +356,7 @@ def scan_network(
         max_threads (int, optional): The maximum number of threads to use when
             scanning the network.
         scan_timeout (float, optional): The network timeout in seconds to use when
-            checking each IP address for a Sonos device
+            checking each IP address for a Sonos device.
         min_netmask (int, optional): The minimum number of netmask bits. Used to
                 constrain the network search space.
 
@@ -306,54 +364,16 @@ def scan_network(
         set: A set of `SoCo` instances, one for each zone found, or else `None`.
     """
 
-    def is_ipv4_address(ip_address):
-        """Helper function to test for an IPv4 address."""
-        try:
-            ipaddress.IPv4Network(ip_address)
-            return True
-        except ValueError:
-            return False
-
-    def find_ipv4_networks(min_netmask):
-        """Helper function to return a set of IPv4 networks to which
-        this node is attached.
-
-        Args:
-            min_netmask (int): The minimum number of netmask bits. Used to
-                constrain the network search space.
-
-        Returns:
-            set: A set of attached networks.
-        """
-        ipv4_net_list = set()
-        adapters = ifaddr.get_adapters()
-        for adapter in adapters:
-            for ifaddr_network in adapter.ips:
-                if is_ipv4_address(ifaddr_network.ip):
-                    ipv4_network = ipaddress.ip_network(ifaddr_network.ip)
-                    # Restrict to private networks and exclude loopback
-                    if ipv4_network.is_private and not ipv4_network.is_loopback:
-                        # Constrain the size of network that will be searched
-                        if ifaddr_network.network_prefix < min_netmask:
-                            ifaddr_network.network_prefix = min_netmask
-                        network = ipaddress.ip_network(
-                            ifaddr_network.ip
-                            + "/"
-                            + str(ifaddr_network.network_prefix),
-                            False,
-                        )
-                        ipv4_net_list.add(network)
-        _LOG.info("List of networks to search: %s", str(ipv4_net_list))
-        return ipv4_net_list
-
     def check_ip_and_port(ip_address, port, timeout):
         """Helper function to check if a port is open"""
+
         _socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         _socket.settimeout(timeout)
         return not bool(_socket.connect_ex((ip_address, port)))
 
     def is_sonos(ip_address):
         """Helper function to check if this is a Sonos device"""
+
         try:
             # Try getting a device property
             _ = config.SOCO_CLASS(ip_address).is_visible
@@ -370,6 +390,7 @@ def scan_network(
         Once a there is a hit, the list is cleared to prevent any further
         checking of addresses by any thread.
         """
+
         while True:
             try:
                 ip_address = str(ip_list.pop())
@@ -386,8 +407,8 @@ def scan_network(
 
     # Generate the set of IPs to check
     ip_set = set()
-    for network in find_ipv4_networks(min_netmask):
-        ip_set.update(network)
+    for network in _find_ipv4_networks(min_netmask):
+        ip_set.update(set(network))
 
     # Find Sonos devices on the list of IPs
     # Use threading to scan the list efficiently
@@ -396,18 +417,22 @@ def scan_network(
     if max_threads > len(ip_set):
         max_threads = len(ip_set)
     for _ in range(max_threads):
+        thread = threading.Thread(
+            target=sonos_scan_worker_thread,
+            args=(ip_set, scan_timeout, sonos_ip_addresses),
+        )
         try:
-            thread = threading.Thread(
-                target=sonos_scan_worker_thread,
-                args=(ip_set, scan_timeout, sonos_ip_addresses),
-            )
-            thread_list.append(thread)
             thread.start()
         except RuntimeError:
-            # We probably can't create any more threads
+            # We probably can't start any more threads
             # Cease thread creation and continue
-            _LOG.info("Runtime error creating threads ... continue")
+            _LOG.info(
+                "Runtime error starting thread number %d ... continue",
+                len(thread_list) + 1,
+            )
             break
+        thread_list.append(thread)
+    _LOG.info("Created %d scanner threads", len(thread_list))
 
     # Wait for all threads to finish
     for thread in thread_list:
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -415,17 +415,17 @@ def scan_network(
 
         while True:
             try:
-                ip = ip_set.pop()
+                ip_addr = ip_set.pop()
             except KeyError:
                 break
 
-            ip_address = str(ip)
+            ip_address = str(ip_addr)
             try:
                 check = _check_ip_and_port(ip_address, 1400, socket_timeout)
             except OSError:
                 # With large numbers of threads, we can exceed the file handle limit.
                 # Put the address back on the list and drop out of this thread.
-                ip_set.add(ip)
+                ip_set.add(ip_addr)
                 break
 
             if check:
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -445,7 +445,7 @@ def scan_network(
         )
         try:
             thread.start()
-        except RuntimeError:
+        except (RuntimeError, OSError):
             # We probably can't start any more threads
             # Cease thread creation and continue
             _LOG.info(
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -244,7 +244,7 @@ def by_name(name):
     return None
 
 
-def scan_network_for_devices(max_threads=256, timeout=1.0, include_invisible=False):
+def scan_network(max_threads=256, timeout=1.0, include_invisible=False):
     """Scan all attached networks for Sonos devices.
 
     Scans the IPv4 network attached to each interface to check for network devices
@@ -292,11 +292,11 @@ def scan_network_for_devices(max_threads=256, timeout=1.0, include_invisible=Fal
                             ipv4_net_list.append(_network)
         return ipv4_net_list
 
-    def check_ip_and_port(_ip_address, port, timeout):
+    def check_ip_and_port(ip_address, port, timeout):
         """Helper function to check if a port is open"""
         _socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         _socket.settimeout(timeout)
-        return bool(_socket.connect_ex((_ip_address, port)))
+        return not bool(_socket.connect_ex((ip_address, port)))
 
     def sonos_scan_worker_thread(ip_list, socket_timeout, sonos_ip_addresses):
         """Helper function worker thread to take IP addresses off a list and
@@ -314,8 +314,8 @@ def scan_network_for_devices(max_threads=256, timeout=1.0, include_invisible=Fal
     # Generate the list of IPs to check
     ip_check_list = []
     for network in find_ipv4_networks():
-        for ip_addr in network:
-            ip_check_list.append(ip_addr)
+        for ip_address in network:
+            ip_check_list.append(ip_address)
     # Find IP addresses with open port 1400
     # Use threading to scan the IP range efficiently
     sonos_ip_addresses = []
@@ -332,7 +332,7 @@ def scan_network_for_devices(max_threads=256, timeout=1.0, include_invisible=Fal
     # Wait for all threads to finish
     for thread in thread_list:
         thread.join()
-    # Pick the first IP address to create a SoCo instance
+    # Pick the first IP address in the list to create a SoCo instance
     for ip_address in sonos_ip_addresses:
         try:
             zone = config.SOCO_CLASS(ip_address)
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -382,8 +382,7 @@ def scan_network(max_threads=256, timeout=3.0, min_netmask=24, include_invisible
     # Generate the set of IPs to check
     ip_set = set()
     for network in find_ipv4_networks(min_netmask):
-        for ip_address in network:
-            ip_set.add(ip_address)
+        ip_set.update(network)
 
     # Find Sonos devices on the list of IPs
     # Use threading to scan the list efficiently
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -332,12 +332,14 @@ def _check_ip_and_port(ip_address, port, timeout):
         bool: True if a connection can be made.
     """
 
+    _socket = None
     try:
         _socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         _socket.settimeout(timeout)
         return not bool(_socket.connect_ex((ip_address, port)))
     finally:
-        _socket.close()
+        if _socket:
+            _socket.close()
 
 
 def _is_sonos(ip_address):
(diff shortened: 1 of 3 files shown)


soco/discovery.py
```diff
@@ -239,3 +242,125 @@ def by_name(name):
...
+        this node is attached.
+        """
+        ipv4_net_list = []
+        adapters = ifaddr.get_adapters()
+        for adapter in adapters:
+            for ip_address in adapter.ips:
+                if is_ipv4_address(ip_address.ip):
+                    # Restrict to common domestic private IP ranges and sensible
+                    # netmasks. Experimental ... assumptions need to be tested.
+                    if (
```
KennethNielsen, line 291: I'm wondering. Is this really necessary. On my system, I get only two ip addresses with network_prefix >= 24, which are localhost and the real local network. I'm just wondering because unless there is a reason to restrict it I don't think we should. SoCo might also be used e.g. for small office networks, which might have very different networks setups.

I think I would suggest to change it to just the network_prefix check AND that the ip address is not 127.0.0.1. On top of that add the option to disregard networks as a keyword argument for cases where it is a major performance hassle.

KennethNielsen (changes requested):
Wow. I think this looks great and it will be a great addition for those cases you mention. I just went a head and gave it a small review and there are only a few minor comments.

We usually add requirements sparsely, but this seems to generate enough bugs that I think it is worth it. I think we can also use the ifaddr in events, where I think there right now is some home cooked code to try and find all interfaces.

I will give it a quick test and make a general comment on a small suggested addition.

KennethNielsen:
It works just fine on my network.

I think this is solid to move forward with. In addition to the review I would add one interface suggestion.

I think there is something elegant around the idea of having the just one function that is the main interface to discovery. So I would suggest that you add this network scan as an optional fallback in the `discover` function, hidden behind a `allow_network_scan=False` keyword argument, and then use it as fallback in cases where the (faster?) broadcast approach fails. I think it should be disabled by default (but easy to reach), because brute force network scans are often considered naughty. **But** leave the `scan_network` function as a _public_ toplevel function, so that user that know they have problems can reach for that directly.

pwt:
Great! Thanks for the comments on this. I'll incorporate your suggestions in the next revision.

I've tested this on about 10 separate domestic networks, and on a variety of host platforms, and it seems fairly solid. My primary concern was the one you mention around network scans, but I guess if one is explicit that it's being done, it should be OK in most network environments.

soco/discovery.py
```diff
@@ -239,3 +242,125 @@ def by_name(name):
...
+                        ip_address.ip.startswith("192.168")
+                        or ip_address.ip.startswith("10.")
+                    ) and ip_address.network_prefix <= 24:
```
rytilahti, line 294: This check would currently ignore the 172.16/12 network. Instead of hardcoded checks, you could use `ip_network()` and the `is_private` property (and `is_loopback` for 127/8):
```
network = ip_network("address")
if not network.is_private or network.is_loopback:
    continue
```

`supernet(new_prefix=24)` can be used to force the network prefix to a specific length, with the caveat of ignoring the fact that some may have larger home networks :-)

pwt: Great -- thanks. This is much better. I'll try it out.

soco/discovery.py
```diff
@@ -239,3 +242,125 @@ def by_name(name):
...
+                        network = ipaddress.ip_network(
+                            ip_address.ip + "/" + str(ip_address.network_prefix), False
+                        )
+                        # Avoid duplicate subnets
+                        if network not in ipv4_net_list:
```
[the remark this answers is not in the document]
pwt: Done.

soco/discovery.py
pwt: I've changed the logic of this based on @rytilahti's comment below. I've also provided an optional `min_netmask` kwarg to constrain the network search space.

soco/discovery.py
pwt: Done. Thanks for the help.

pwt:
> I think there is something elegant around the idea of having the just one function that is the main interface to discovery. So I would suggest that you add this network scan as an optional fallback in the `discover` function, hidden behind a `allow_network_scan=False` keyword argument, and then use it as fallback in cases where the (faster?) broadcast approach fails. I think it should be disabled by default (but easy to reach), because brute force network scans are often considered naughty. **But** leave the `scan_network` function as a _public_ toplevel function, so that user that know they have problems can reach for that directly.

I've refactored the implementation to follow the proposal above. Each of the `discover()`, `by_name()` and `any_soco()` functions now take two optional kwargs: `allow_network_scan` and `min_netmask`. (It's arguable whether the second of these is required, because one can use the `scan_network()` function directly and access all of its options.)

pwt:
There's a failing test in the checks, although not on my machine with the same versions of the relevant tools. Will investigate.

```
=================================== FAILURES ===================================
608_________________________________ test_by_name _________________________________
609
610    def test_by_name():
611        """Test the by_name method"""
612        devices = set()
613        for name in ("fake", "non", "Kitchen"):
614            mymock = Mock(player_name=name)
615            devices.add(mymock)
616    
617        # The mock we want to find is the last one
618        mock_to_be_found = mymock
619    
620        # Patch out discover and test
621        with patch("soco.discovery.discover") as discover_:
622            discover_.return_value = devices
623    
624            # Test not found
625            device = by_name("Living Room")
626            assert device is None
627>           discover_.assert_called_once_with()
628
629tests/test_discovery.py:82: 
630_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
631../../../virtualenv/python3.8.0/lib/python3.8/site-packages/mock/mock.py:957: in assert_called_once_with
632    return self.assert_called_with(*args, **kwargs)
633../../../virtualenv/python3.8.0/lib/python3.8/site-packages/mock/mock.py:944: in assert_called_with
634    six.raise_from(AssertionError(_error_message(cause)), cause)
635_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
636
637value = None, from_value = None
638
639>   ???
640E   AssertionError: expected call not found.
641E   Expected: discover()
642E   Actual: discover(allow_network_scan=False, min_netmask=24)
643
644<string>:3: AssertionError
```

pwt:
Just to note that commit https://github.com/SoCo/SoCo/pull/733/commits/198316f5b68e7242162beaf12fc13cbfd28b85ba corrects a corner case problem, where a device can be found with port 1400 open, but which is not a Sonos device, **and** no other IP addresses are collected.

In my case I had a machine on the network running soco-cli and subscribed to events on Port 1400.

Now I explicitly check for a Sonos device before adding an IP to the sonos list, and shutting down the threads.

soco/discovery.py
```diff
@@ -215,27 +243,180 @@ def any_soco():
...
+                    _LOG.info("Confirmed Sonos device at IP '%s'", ip_address)
+                    sonos_ip_addresses.append(ip_address)
+                    # Clear the list to eliminate further searching by
+                    # all threads
+                    ip_list.clear()
+
+    # Generate the set of IPs to check
+    ip_set = set()
+    for network in find_ipv4_networks(min_netmask):
+        for ip_address in network:
```
KennethNielsen, line 382: If network is iterable, it might be possible to get rid of the inner loop with something like:

```python
for network in find_ipv4_networks(min_netmaks):
    ip_set.update(set(network))
```

soco/discovery.py
```diff
@@ -215,27 +243,180 @@ def any_soco():
...
     if devices is None:
         return None
 
     for device in devices:
         if device.player_name == name:
             return device
     return None
+
+
+def scan_network(max_threads=256, timeout=3.0, min_netmask=24, include_invisible=False):
```
KennethNielsen, line 280: I was thinking a little about what a sensible default value of timeout would be here. Since that timeout applies to every individual network wait-for-reply, the max time the function takes is only capped at: `num_ip_addresses_in_network / max_threads * timeout`, so if the network is large, that time could be large and the real time it takes is at least proportional to the timeout. So the thing is that if we do get a reply, I would expect it to be fairly quick. On my own network this function return the correct result with timeout as low as 1E-5:
```
>>> scan_network(timeout=0.001)
{SoCo("192.168.1.173"), SoCo("192.168.1.197"), SoCo("192.168.1.31"), SoCo("192.168.1.18"), SoCo("192.168.1.215")}
>>> scan_network(timeout=1E-5)
{SoCo("192.168.1.173"), SoCo("192.168.1.197"), SoCo("192.168.1.31"), SoCo("192.168.1.18"), SoCo("192.168.1.215")}
>>> scan_network(timeout=1E-6)
>>>
```

One should of course always be careful with extrapolating from just one case, but setting it to say 0.1 would still be 1E4 more than my lowest observed success case. I know that you tested this on multiple different networks, so I would be interested to hear what your thoughts are.

soco/discovery.py
```diff
@@ -19,7 +22,13 @@
 # pylint: disable=too-many-locals, too-many-branches
 
 
-def discover(timeout=5, include_invisible=False, interface_addr=None):
+def discover(
+    timeout=5,
+    include_invisible=False,
+    interface_addr=None,
+    allow_network_scan=False,
+    min_netmask=24,
```
KennethNielsen, line 30: Ok, so I wish I had remembered to mention this when I first suggested this implementation approach. But since we are adding 2 extra arguments to all of them, then we might as well make the second one a `network_scan_kwargs` dict argument and allow full configurability of the network_scan with the same number of arguments. It might also come in handy in the future, since I don't imagine these functions are done growing in size. E.g. we might in the future see them changed to allow not to stop after the first result (to allow to discover S1&S2 systems) and maybe a caching option will be added. What do you think?

KennethNielsen (changes requested):
I gave this s second review after all the changes and I think it looks great. There a few slightly more in-depth questions and a few nitpicks left.

soco/discovery.py
```diff
@@ -215,27 +243,180 @@ def any_soco():
...
+            return False
+
+    def sonos_scan_worker_thread(ip_list, socket_timeout, sonos_ip_addresses):
+        """Helper function worker thread to take IP addresses off a list and
+        test whether there is (1) a device with port 1400 open at that IP
+        address, then (2) check the device is a Sonos device.
+        Once a there is a hit, the list is cleared to prevent any further
+        checking of addresses by any thread.
+        """
+        while len(ip_list) > 0:
```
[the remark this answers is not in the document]
pwt: I think you're right. Will be fixed in https://github.com/SoCo/SoCo/commit/7f5e87e2860dbf4999bca768db77795498b66a9c.

soco/discovery.py
pwt: Very neat. The inner `set()` cast is not required as `network` is already a set. Implemented in https://github.com/SoCo/SoCo/commit/5e4465ffa3791d61f51c56326c71a0fc96342afe.

soco/discovery.py
pwt: You're right. Changed in https://github.com/SoCo/SoCo/commit/1b81d6fa1dfcd8018f2736230546443a3f785721.

There's a bit of history behind this, back from when I wanted an exhaustive list of Sonos devices on the network for SoCo-CLI, and there were some outlier responses on my networks (ARP delays?). The shorter timeout should be OK in the majority of cases, although I sometimes saw timeouts in excess of 3s.

soco/discovery.py
pwt: I'm currently taking  a look at switching args ever to a dict.

Scanning all households (such as the split S1/S2 case) would be a natural enhancement (though we'd probably need to look holistically at SoCo to make sure it operates consistently). Incidentally, that's what I do with the equivalent implementation in soco-cli [1], which also caches the results in a file. I need it myself because I have an S1/S2 system.

[1] https://github.com/avantrec/soco-cli/blob/master/soco_cli/speakers.py

soco/discovery.py
KennethNielsen: Oh yeah, but a set that you create or one that is provided by the library. The type of return values from the library could change but "being iterable" is much less likely to change.

soco/discovery.py
pwt: This is now implemented in https://github.com/SoCo/SoCo/pull/733/commits/9964b2bc9946e90cacc84483fe3840021b17726c, and tested. I need to check whether the docstrings are correctly formatted, since this is not a pattern I've used much.

KennethNielsen: Hi @pwt.

Ah, I was a little in doubt about how much to write about it. It's, I think, almost the way that it is mostly done.

It is fairly common to collect arguments for a sub-function via the **scan_network, just as you did. I would however not document them there, but simply refer to the function they apply to. Another thing is that `allow_network_scan` really is an argument for the e.g. `discover` function, not for `network_scan`, so I would leave that as a normal argument. Also, this whole thing, with unpacking arguments into a function call from a dict, works just fine even though network scan has normal arguments, so I would definitely keep those, since they carry more information. So to sum up, something like this is what I would suggest, and the way I think that pattern is usually implemented:

```python

def discover(
        timeout=5, include_invisible=False, interface_addr=None,
        allow_network_scan=False, **network_scan_kwargs
):
    """Discover Sonos zones on the local network.

    Return a set of `SoCo` instances for each zone found.
    Include invisible zones (bridges and slave zones in stereo pairs if
    ``include_invisible`` is `True`. Will block for up to ``timeout`` seconds,
     after which return `None` if no zones found.

    Args:
        timeout (int, optional): block for this many seconds, at most.
            Defaults to 5.
        include_invisible (bool, optional): include invisible zones in the
            return set. Defaults to `False`.
        interface_addr (str or None): Discovery operates by sending UDP
            multicast datagrams. ``interface_addr`` is a string (dotted
            quad) representation of the network interface address to use as
            the source of the datagrams (i.e. it is a value for
            `socket.IP_MULTICAST_IF <socket>`). If `None` or not specified,
            the system default interface for UDP multicast messages will be
            used. This is probably what you want to happen. Defaults to
            `None`.
        allow_network_scan (bool, optional): If normal discovery fails, fall
            back to a scan of the attached network(s) to detect Sonos
            devices.
        **network_scan_kwargs: Arguments for the
            :function:`scan_network` function. See its docstring for
            documentation

    Returns: set: a set of `SoCo` instances, one for each zone
        found, or else `None`.

    ...

    """
    ...
    
        elif allow_network_scan:
            _LOG.info("Falling back to network scan discovery")
            return scan_network(
                include_invisible=include_invisible,
                **network_scan_kwargs,
            )

# any_soco similar to by_name below

def by_name(name, allow_network_scan=False, **network_scan_kwargs):
    """Return a device by name.

    Args:
        name (str): The name of the device to return.
        allow_network_scan (bool, optional): If normal discovery fails, fall
            back to a scan of the attached network(s) to detect Sonos
            devices.
        **network_scan_kwargs: Arguments for the
            :function:`scan_network` function. See its docstring for
            documentation Returns: set: a set of `SoCo` instances, one
            for each zone found, or else `None`.

    Returns:
        :class:`~.SoCo`: The first device encountered among all zone with the
            given player name. If none are found `None` is returned.
    """
    devices = discover(
        allow_network_scan=allow_network_scan, **network_scan_kwargs
    )
    if devices is None:
        return None

    for device in devices:
        if device.player_name == name:
            return device
    return None


# Function def args here are as before |
#                         v

# pylint: disable=too-many-statements
def scan_network(max_threads=256, timeout=3.0, min_netmask=24, include_invisible=False):
    """Scan all attached networks for Sonos devices.

    Scans the IPv4 network attached to each interface to check for network devices
    with port 1400 open. Check IPs in parallel threads for efficiency. Once the first
    Sonos device is found, stop checking and use that device to find the other
    devices. Returns a set of `SoCo` instances, or `None` if no Sonos devices are
    discovered.

    This function is intended for use when the usual discovery functions aren't
    working, perhaps due to multicast problems on the network to which the SoCo
    host is attached.

    Args:
        max_threads (int, optional): The maximum number of threads to use when
            scanning the network.
        scan_timeout (float, optional): The network timeout in seconds to use when
            checking each IP address for a Sonos device
        min_netmask (int, optional): The minimum number of netmask bits. Used to
                constrain the network search space.
        include_invisible (bool, optional): Whether to include invisible Sonos devices
            in the set of devices returned.

    Returns:
        set: A set of `SoCo` instances, one for each zone found, or else `None`.
    """
    ...

```

I hope the arguments makes sense, otherwise let me know.

pwt: Got it. Thanks for the explanation. Amended in https://github.com/SoCo/SoCo/pull/733/commits/449d1668e2abd7ab985707c7af3c09aa76a36192.

soco/discovery.py
```diff
@@ -215,27 +242,240 @@ def any_soco():
...
+    Args:
+        min_netmask(int): The minimum netmask to be used.
+
+    Returns:
+        set: A set of `ipaddress.ip_network` instances.
+    """
+
+    ipv4_net_list = set()
+    adapters = ifaddr.get_adapters()
+    for adapter in adapters:
```
rytilahti, line 309: Here's a simpler variant (to avoid multi-level nesting), if you wish:
```
# Get list of IPv4 networks
nets = [ip_network(f"{ip.ip}/{ip.network_prefix}", strict=False)   # strict=False allows host bits in the address
        for adapter in ifaddr.get_adapters()
        for ip in adapter.ips 
        if not isinstance(ip.ip, tuple)  # IPv6 addresses are reported as tuples by ifaddr
]
# Filter out loopback and public networks
nets = [net for net in nets if net.is_private and not net.is_loopback]

# unfortunately there is no easy way to get a network using given prefixlen without touching internals..
def _minimize_netmask(net, min):
    if net.prefixlen < min:
        net._prefixlen = min
    return net

# Get networks with the minimum netmask
nets = [_minimize_netmask(net, min_netmask) for net in nets]
```

soco/discovery.py
```diff
@@ -215,27 +242,250 @@ def any_soco():
...
+        port(int): The port to be checked.
+        timeout(float): The timeout to use.
+
+    Returns:
+        bool: True if a connection can be made.
+    """
+
+    _socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    _socket.settimeout(timeout)
+    return not bool(_socket.connect_ex((ip_address, port)))
```
rytilahti, line 348: I'm not definitely sure, but wouldn't this leak sockets without doing something like this?
```suggestion
   try:
        _socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        _socket.settimeout(timeout)
       return not bool(_socket.connect_ex((ip_address, port)))
   finally:
       _socket.close()
```

pwt: Thanks. I wondered about that, but I think you're right ... they don't get GC'd until the process ends.

soco/discovery.py
pwt: I've kept the nested form for now.

pwt:
I've now added a `multi_household` option, which will exhaustively find all speakers on the attached networks. This means that multiple household Sonos systems (e.g., S1/S2 split systems) can be discovered, which is a new SoCo feature.

This is not guaranteed to work well with the rest of SoCo. In particular, it interacts poorly with caching, so I recommend that `soco.core.zone_group_state_shared_cache.enabled` is set to `False`, as is done in the `scan_network()` code when `multi_household` is enabled. There may be other wrinkles.

Note that speakers from different households can be controlled independently by SoCo, but they are in different systems and cannot be grouped or paired, and have separate playlists, subscriptions, libraries, etc.

amelchio:
The bad cache interaction is fixed by #656. Home Assistant has been using this for about two years without issue (via my pysonos fork), though I guess multi-household systems are rare.

pwt:
> The bad cache interaction is fixed by #656. Home Assistant has been using this for about two years without issue (via my pysonos fork), though I guess multi-household systems are rare.

Good. We should get that fix into SoCo. I'll look into that. Multi-household systems are probably more common now, following the S1/S2 thing. That's why I'm running a 'split' system.

(I've just been disabling caching when I do discovery in [SoCo-CLI](https://github.com/avantrec/soco-cli), and that's been working OK. Similar code to this PR.)

pwt:
This has stabilised now, and is ready for final review. If/when we do merge it, let's aim to use squash-merge to avoid including all of its evolution in the master branch history.

KennethNielsen:
@pwt @amelchio I have made #656 part of this milestone. Let's have a look at that next.

soco/discovery.py
```diff
@@ -215,27 +243,378 @@ def any_soco():
...
+            See its docstring for details. (Note that the argument
+            'multi_household' is forced to `True` when this function is
+            called.)
+
+    Returns:
+        set: A set of `SoCo` instances, one for each zone found, or else `None`.
+    """
+
+    # Take a copy to avoid creating side effects
+    new_kwargs = network_scan_kwargs.copy()
```
KennethNielsen, line 425: This isn't necessary, since the dict is only created when the function is called.

soco/discovery.py
```diff
@@ -215,27 +243,378 @@ def any_soco():
...
+            'multi_household' is forced to `True` when this function is
+            called.)
+
+    Returns:
+        list: A list of Sonos Household IDs, each in the form of a `str`
+        like 'Sonos_XXXXXXXXXXXXXXXXXXXXXXXXXX'.
+    """
+
+    # Take a copy to avoid creating side effects
+    new_kwargs = network_scan_kwargs.copy()
```
KennethNielsen, line 451: Same here

soco/discovery.py
```diff
@@ -215,27 +243,378 @@ def any_soco():
...
+    # multi_household must be set to True
+    new_kwargs["multi_household"] = True
+    zones = scan_network(include_invisible=True, **new_kwargs)
+    household_ids = set()
+    if zones:
+        for zone in zones:
+            household_ids.add(zone.household_id)
+
+    _LOG.info("Returning household IDs: %s", household_ids)
+    return list(household_ids)
```
KennethNielsen, line 461: Or just return the set, since it is one anyway and the content clearly is so.

soco/discovery.py
```diff
@@ -215,27 +243,378 @@ def any_soco():
...
+            'multi_household' is forced to `True` when this function is
+            called.)
+
+    Returns:
+        SoCo: A `SoCo` instance representing the zone, or `None` if no
+        matching zone is found. Only returns visible zones.
+    """
+
+    # Take a copy to avoid creating side effects
+    new_kwargs = network_scan_kwargs.copy()
```
KennethNielsen, line 485: and here

KennethNielsen (dismissed):
There are just a few remaining comments about a non-necessary dict copy and then we are good to go. I noticed it grew quite a few extra convenience functions all of which I think are a good idea.

soco/discovery.py
pwt: Didn't know that. Done.

KennethNielsen (approved):
:+1:

PR #738, opened 2020-08-19
Add DIDL Lite support for AirPlay 2 streams

pwt:
An AirPlay 2 stream will cause exceptions to be thrown in `attempt_datastructure_upgrade`. This PR adds the required DIDL classes.


pwt:
This fix is longer required in v0.20.

Issue #740, opened 2020-09-10
Add to_dict() to MSTrack


declension:
First, thanks for a great project!
### Feature Request
A new `MSTrack.to_dict()`, to behave like `DidlMusicTrack.to_dict()`
 
### Motivating Example
```python
>>> print([(type(track).__name__, track.to_dict())
...              for track in get_playlist_tracks(device, playlist)])
#
# lots of serialised DidlMusicTrack output, then
# 
AttributeError: Class MSTrack has no attribute "to_dict"
```

Related: I was expecting Music Service Tracks to share some of the same Didl hierarchy, but I guess there are good reasons why that's not the case, but a common superclass / mixin with just `to_dict()` might be nice too...

JetDeveloping:
Hey, @declension! I created a pull request that added an 'MSTrack.to_dict()' function. However, until this gets approved (or it doesn't), MSTrack has a `content` attribute, so in the meantime you can basically just do:
`MSTrack.content.` Obviously, since the values are dynamic, you'll have to initialize the class beforehand trying to get this.

Hope I could help!

KennethNielsen:
@declension hi. Thanks you for the requests. Is there a usecase behind the request other thans printing? Also, which MSTrack do you refer to. For historical reasons there are 2 (don't ask). One (in soco.ms_data_structures) is used for music service plugins, of which there was a few back in the day, and the other (in soco.music_services.data_structures) is used by the dynamically generated music services (which currently don't work due to Sonos having changed authentication on us). If you item you are referring to is coming from the queue, it is most likely the latter, but please confirm.

declension:
Thanks - IIRC it was getting playlists items out of a playlist with mostly local files and occasional Music Services files, if that helps

PR #742, opened 2020-09-11
Added a "to_dict" to MSTrack

JetDeveloping:
MSTrack already has a content attribute, I just created a "to_dict" that would return MSTrack.content!


KennethNielsen:
This PR might be modding the wrong class. Waiting for clarification in #740.

JetDeveloping:
Sounds good, @KennethNielsen!

PR #744, opened 2020-10-04
Pin pylint to v2.5 and Black to v19, to avoid new warnings until after 0.20 release. 

pwt:
All checks are currently failing due to the pylint update to v2.6. This PR suppresses two additional pylint warnings, allowing the checks to pass.

Also, the update to Black v20 flags multiple files for reformatting, hence failing the Black check. Pin Black to v19 for the time being.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -7,7 +7,7 @@
 # progress
 disable=too-many-lines,locally-disabled,duplicate-code,too-few-public-methods,
     bad-option-value,no-else-return,cyclic-import,too-many-public-methods,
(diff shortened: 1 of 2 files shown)


pwt:
I've changed the PR to defer the new warnings from pylint and Black by pinning their versions, allowing checks to pass. This should be reviewed after the v0.20 SoCo release.

KennethNielsen:
I will merge now to cerate the release and then we can see about more forward pointing fixes aftewards,

PR #749, opened 2020-10-19
Unpin Pylint version and eliminate new warnings

pwt:
This PR removes the Pylint version pin used in the checks, and eliminates the multiple instances of two new warnings that are reported under Pylint 2.6:

1. Removes 'super-with-arguments' warnings (R1725) by changing all `super()` invocations to Python 3 style.
2. Removes 'raise-missing-from' warnings (W0707) by adding `from <initial exception>` where required.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -7,7 +7,7 @@
 # progress
 disable=too-many-lines,locally-disabled,duplicate-code,too-few-public-methods,
     bad-option-value,no-else-return,cyclic-import,too-many-public-methods,
-    bad-continuation, raise-missing-from
+    bad-continuation
 
 
 [REPORTS]
--- a/pylintrc
+++ b/pylintrc
@@ -7,7 +7,7 @@
 # progress
 disable=too-many-lines,locally-disabled,duplicate-code,too-few-public-methods,
     bad-option-value,no-else-return,cyclic-import,too-many-public-methods,
-    bad-continuation
+    bad-continuation, raise-missing-from
 
 
 [REPORTS]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,7 +4,7 @@ pytest >= 2.5
 mock >= 1.0.1
 graphviz
 flake8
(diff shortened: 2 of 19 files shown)


pwt:
I decided to address the `W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)` warnings rather than suppress them. It was straightforward.

(The definitive discussion on this new warning is at: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with)

pylintrc
```diff
@@ -7,7 +7,7 @@
 # progress
 disable=too-many-lines,locally-disabled,duplicate-code,too-few-public-methods,
     bad-option-value,no-else-return,cyclic-import,too-many-public-methods,
-    bad-continuation
+    bad-continuation, raise-missing-from
```
KennethNielsen: This one is a little special. The `raise from` form actually adds information to the traceback, which is useful and I therefore think this is one of the additions to the pylint checker that I fully agree with. Se here for an explanation of the error: https://stackoverflow.com/a/63852262/11640721 and here for a little more explanation of the `raise from` form: https://stackoverflow.com/questions/24752395/python-raise-from-usage

I don't know how many try-excepts will need to be fixed, but I can lend a hand if it is many.

KennethNielsen (changes requested):
Ok. All the `super` changes look good. As per the review comment I actually think we should fix the `raise from` stuff up. Let me know what you think.

pylintrc
pwt: Yep. I've actually already gone ahead and fixed up the `raise from` warnings. Decided it was better once I understood it.

Might also want to specify `pylint >= 2.6` in `requirements-dev.txt`; what do you think?

KennethNielsen (approved):
Looks good. Ready for merge.

pylintrc
KennethNielsen: Yeah, that is probably a good idea.

PR #751, opened 2020-10-22
Replace 'isAlive()' with 'is_alive()' for Python 3.9 compatibility

pwt:
Python 3.9 checks are failing because `Thread.isAlive()` was deprecated in Python 3.8 and removed in Python 3.9. Replace with `Thread.is_alive()`.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -232,7 +232,7 @@ class EventListener(EventListenerBase):
         # to ensure the main thread does not hang
         self._listener_thread.join(1)
         # check if join timed out and issue a warning if it did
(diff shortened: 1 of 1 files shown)


pwt:
The Python '3.9-dev' check job is now passing. Yey.

KennethNielsen:
Great. Merged.

pwt:
Note that if you want SoCo < v0.21 events to work with Python 3.9, add the following somewhere in your code to patch in the now missing `isAlive()` method:
```
from threading import Thread
Thread.isAlive = Thread.is_alive
```

PR #755, opened 2020-11-13
Network scan discovery: add ability to specify the network(s) to scan

pwt:
This PR just adds the option to specify exactly which network(s) to scan, instead of depending on the default mechanism of interrogating the network adapters.


KennethNielsen (approved):
Looks great. Let's get it in. Don't forget to add a comment to the release notes thread after merging.

Issue #723, opened 2020-06-01
Add Sonos Move Battery Level to SoCo API


KurtSanders:
Add SoCo API functionality to obtain Sonos Move battery level.

![Sonos Move Battery Level](https://user-images.githubusercontent.com/12133562/83397743-1fc6ce80-a3cc-11ea-8aad-f499115328e1.jpg)

jjlawren:
A charging indicator would also be useful.

KennethNielsen:
Looks like a great idea. Is either of you interested in investigating whether it is possible and doing the implementation?

jjlawren:
I would, but I’m starting from scratch on my knowledge of the protocol. Any suggestions on where to start looking for the battery/charger payloads?

pwt:
I used Wireshark to take a superficial look at the interactions between a Sonos controller app and a Move, but the battery level and charging state didn't jump out at me.

jjlawren:
I've looked through the various XML endpoints and nothing seemed to indicate a battery level or charging indicator. Is it possible it's only provided via events?

pwt:
One place battery info can be found is in the speaker support page. Go to:
`http://192.168.0.41:1400/support/review` (substituting the correct IP address for your Move), then click on the name of the Move speaker, then `Battery Status`:

![screenshot_0838](https://user-images.githubusercontent.com/1089749/84040487-9babb280-a99a-11ea-8f5a-6861424e3f56.png)

Note that my Move is running the just released **S2** Sonos software, so it's possibly different under S1.

(The good news is that my preliminary testing indicates that SoCo works with S2.)

tksunw:
I have an S1 Move, and it does not have the "Battery Status" link in the 'support/review' page.

KennethNielsen:
@tksunw you might try and have a look at the address. It might very well have the status page even if the app doesn't expose it.

KennethNielsen:
@pwt can you try and have a look if it has an xml version as well in some variant of xml right after the port and as an extension, like the device page: http://192.168.1.100:1400/xml/device_description.xml

pwt:
> @pwt can you try and have a look if it has an xml version as well in some variant of xml right after the port and as an extension, like the device page: http://192.168.1.100:1400/xml/device_description.xml

Nothing obvious in the device desc. XML. There is, however, a chunk of XML in the page source of the support/review URL I pointed to above:

```
...
<LocalBatteryStatus>
<Data name="Health">GREEN</Data>
[...]

PR #756, opened 2020-11-14
Add 'get_battery_info' method to the SoCo class

pwt:
This PR addresses #723.

I've only tested against a Move on my S2 system, and there are indications that this approach doesn't work with S1 systems. It would be useful if someone could test this.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -12,9 +12,13 @@ import re
 import socket
 from functools import wraps
 from xml.sax.saxutils import escape
+from xml.parsers.expat import ExpatError
 import warnings
+import xmltodict
 
 import requests
+from requests.exceptions import ConnectionError as RequestsConnectionError
+from requests.exceptions import ConnectTimeout, ReadTimeout
 
 from . import config
 from .compat import UnicodeType
@@ -224,6 +228,7 @@ class SoCo(_SocoSingletonBase):
         get_sleep_timer
         create_stereo_pair
         separate_stereo_pair
+        get_battery_info
 
     .. warning::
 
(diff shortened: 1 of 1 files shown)


KennethNielsen:
I think I have a mixed S1 and S2, but I don't have any devices with battery :(In any case.

pwt:
> I think I have a mixed S1 and S2, but I don't have any devices with battery :(

I factory reset my Move and attempted to add it to my S1 system for testing. However, it looks like there's a bug in the current Sonos S1 iOS/iPadOS controller apps, meaning the process gets stuck when adding WiFi creds. Fortunately, I was able to add it back to my S2 system without any issues.

> In any case. Since the code reaches fairly wide, when the feature is not supported, I would be ok to merge it even without testing on S1 and then improve if anyone opens up an issue saying it doesn't work on S1 but should. There isn't much more we can do without more data or testing. What do you think @pwt ?

Yes, I agree. It's not as if it's a mission-critical feature :)

KennethNielsen:
Do you want to add a software test for it or should I review as is?

pwt:
> Do you want to add a software test for it or should I review as is?

I'm not planning to add a test for this one.

pwt:
Note that I could test explicitly for whether the target speaker is a Move, and avoid the network call if it isn't.

I chose not to do this because if additional battery-powered speakers arrive, there's a good chance that this call will work without change for these new speakers.

soco/core.py
```diff
@@ -2153,6 +2158,70 @@ def get_sonos_playlist_by_attr(self, attr_name, match):
...
+            timeout (float, optional): The timeout to use when making the
+                HTTP request.
+
+        Returns:
+            dict: A `dict` containing battery status data.
+
+            Example return value::
+
+                {'Health': 'GREEN',
+                 'Level': '100',
```
KennethNielsen, line 2183: Should we maybe parse the Level and make it an int?

KennethNielsen (approved):
I have one minor suggestion, but it is approved as is if you don't agree.

soco/core.py
pwt: I did consider reaching into the battery data, but concluded that it's less brittle just to gather what's there and pass it back without making any assumptions.

pwt: OK, if "Level" is present in the dict, it now gets converted to an int. Will still work if the fields change.

Issue #764, opened 2020-12-01
discovery() only enumerates first ipv4 address for host name


gfischershaw:
Running VirtualBox with a Host-Only Ethernet Adapter. The **discovery()** method picks up that address first and so cannot enumerate the devices on the local network. Switching to **gethostbyname_ex()** seems to work.

```
`        addresses = set()
        try:
            _,_,addressList = socket.gethostbyname_ex(socket.gethostname())
            addresses.update(addressList)
        except socket.error:
            pass
        try:
            _,_,addressList = socket.gethostbyname_ex(socket.getfqdn())
            addresses.update(addressList)
        except socket.error:
            pass
````

pwt:
Note that you can supply an `interface_addr` as an optional parameter to `soco.discovery.discover()` [1]. Perhaps you could give that a try. Also, there's a new discovery technique arriving in SoCo 0.21 that can search all interfaces for Sonos devices, which will circumvent problems with selecting the right network interface(s).

[1] http://docs.python-soco.com/en/v0.20/api/soco.discovery.html
[2] http://docs.python-soco.com/en/latest/api/soco.discovery.html#soco.discovery.scan_network

gfischershaw:
When I first ran into this problem I did indeed hardcode the correct IP, but it did not seem like a good solution to do that for each and every device on which I ran a script. The function is also used in discovery.any_soco(), discovery.by_name(), and get_alarms(). It seemed better to just fix the problem, which is to get all the interface IPs. The new code looks like it just uses the current methodology before then calling discovery.scan_network(). Seems like first portion could be switched to from gethostbyname() to using gethostbyname_ex(), and therefore better deal with most cases and never need to call scan_network().

pwt:
> When I first ran into this problem I did indeed hardcode the correct IP, but it did not seem like a good solution to do that for each and every device on which I ran a script. And the function is also used in discovery.any_soco(), discovery.by_name(), and get_alarms(). It seemed better to just fix the problem, which is to get all the interface IPs.

Your proposed change looks like a good improvement for single interface hosts with multiple IP addresses. Would you like to create a PR for review?

> The new code looks like it just uses the current methodology before then calling discovery.scan_network(). Seems like first portion could be switched from gethostbyname() to using gethostbyname_ex(), and therefore better deal with most cases and never need to call scan_network().

[...]

PR #767, opened 2020-12-02
Improve discover(): perform discovery on all NICs/IPs

pwt:
This PR changes the way that `discover()` determines the IP addresses to use for multicast discovery. It replaces the use of `gethostbyname()` with a comprehensive list of IPv4 addresses obtained using the `ifaddr` library. This should cover cases where the host has multiple NICs and/or multiple IP addresses associated with those NICs, and prevent some of the common problems encountered when using discovery. 

Addresses #764.

### DIFF
--- a/soco/discovery.py
+++ b/soco/discovery.py
@@ -60,7 +60,7 @@ def discover(
     """
 
     def create_socket(interface_addr=None):
-        """A helper function for creating a socket for discover purposes.
+        """A helper function for creating a socket for discovery purposes.
 
         Create and return a socket with appropriate options set for multicast.
         """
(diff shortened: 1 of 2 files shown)


gfischershaw:
Works fine on my setup with multiple addresses on an interface. Probably don't need the `None `default on `create_socket `any longer.

`def create_socket(interface_addr=None):`

KennethNielsen:
My PC just have a single interface, so I don't how much it is worth, but I can confirm that discovery on the S1 system still works with this PR.

soco/discovery.py
```diff
@@ -541,6 +506,34 @@ def _find_ipv4_networks(min_netmask):
     return ipv4_net_list
 
 
+def _find_ipv4_addresses():
+    """Discover all the host's IP addresses.
```
KennethNielsen, line 510: Discover and return ... 
maybe

KennethNielsen (approved):
Looks great. Approved as is, although I should say that this sort of network code is not my strongest suit, so if you can poke someone from one of the other PRs do a second review that might be a good idea.

pwt:
> My PC just have a single interface, so I don't how much it is worth, but I can confirm that discovery on the S1 system still works with this PR.

Thanks for testing, @KennethNielsen. It's now been tested on multihomed and aliased systems, and on Windows, Linux, and macOS. Not seeing any issues (and it will certainly address some existing limitations).

soco/discovery.py
pwt: Done.

PR #772, opened 2020-12-07
Do not try to restore bass/treble/loudness with fixed volume

amelchio:
The fixed volume setting (on e.g. the Connect) also fixes the equalization settings so we should not try to restore those.


pwt:
LGTM.

PR #773, opened 2020-12-08
Add a 'fixed_volume' property to the SoCo class

pwt:
This PR adds a gettable/settable `fixed_volume` property to the SoCo class, with accompanying tests. It also adds a `supports_fixed_volume` property to determine if a device supports this feature.

This (currently) only applies to Sonos Connect and Sonos Port devices. All non-applicable devices raise a `NotSupportedException` on set attempts, and will always return `False` on gets.

Tested with both Sonos Connect (S1) and Sonos Port (S2), as well as with devices for which `fixed_volume` is non-applicable.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -197,6 +197,7 @@ class SoCo(_SocoSingletonBase):
         balance
         night_mode
         dialog_mode
+        fixed_volume
         status_light
 
     ..  rubric:: Playlists and Favorites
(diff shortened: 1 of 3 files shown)


KennethNielsen (approved):
Looks great. Approved as is.But I would like to see a positive test on one other house hold.

KennethNielsen:
I tested it and I see the volume slider disappear in the Android app so that's a 👍  to merge

PR #774, opened 2020-12-08
Add a 'buttons_enabled' property to the SoCo class

pwt:
This PR adds a gettable/settable `buttons_enabled` property to the SoCo class, including accompanying tests.

Note that this inverts the function of the underlying `deviceProperties.ButtonLockState*` operations. I think it's more meaningful to deal with buttons that are enabled or disabled, rather than locked or unlocked.

Can only be set on visible speakers, or an exception (`SoCoNotVisibleException`) is raised. Applying to non-visible speakers does 'interesting' things, and probably not what's desired.

Tested on S1 and S2 devices, with and without control buttons. For devices without buttons (e.g., Sonos Port) the operations still succeed and don't appear to have any effect nor to do any harm.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -35,6 +35,7 @@ from .exceptions import (
     SoCoSlaveException,
     SoCoUPnPException,
     NotSupportedException,
+    SoCoNotVisibleException,
 )
 from .groups import ZoneGroup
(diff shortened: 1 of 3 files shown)


KennethNielsen (approved):
Looks good. I will give it a quick test run.

PR #775, opened 2020-12-09
Add 'trueplay' property to SoCo

pwt:
This PR adds a gettable/settable property `trueplay`, with accompanying tests. The property allows the Trueplay state of a speaker (`True` or `False`) to be determined, and to be changed.

Devices that do not support Trueplay, or which do not have  a current Trueplay calibration, will return `None` on getting the property, and  raise a `NotSupportedException` when setting the property. Can only be set on visible devices. Attempting to set on non-visible devices will raise a `SoCoNotVisibleException`.

Tested on S1 and S2 devices, devices that don't support Trueplay, and devices that do support Trueplay, with and without a current calibration profile.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -968,7 +968,10 @@ class SoCo(_SocoSingletonBase):
 
         Devices that do not support Trueplay, or which do not have
         a current Trueplay calibration, will raise a `NotSupportedException`
-        both when getting and setting the property..
+        when both getting and setting the property.
+
+        Can only be set on visible devices. Attempting to set on non-visible
+        devices will raise a `SoCoNotVisibleException`.
         """
         response = self.renderingControl.GetRoomCalibrationStatus([("InstanceID", 0)])
         if response["RoomCalibrationAvailable"] == "0":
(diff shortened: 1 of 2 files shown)


rytilahti:
Works fine on trueplay tuned beam, changing the value reflects immediately to the settings.

soco/core.py
```diff
@@ -960,6 +961,48 @@ def dialog_mode(self, dialog_mode):
...
+        a current Trueplay calibration, will raise a `NotSupportedException`
+        when both getting and setting the property.
+
+        Can only be set on visible devices. Attempting to set on non-visible
+        devices will raise a `SoCoNotVisibleException`.
+        """
+        response = self.renderingControl.GetRoomCalibrationStatus([("InstanceID", 0)])
+        if response["RoomCalibrationAvailable"] == "0":
+            raise NotSupportedException
+        return response["RoomCalibrationEnabled"] == "1"
```
rytilahti, line 979:
```suggestion
        if response["RoomCalibrationAvailable"] == "0":
            raise NotSupportedException
            
        return response["RoomCalibrationEnabled"] == "1"
```
Separating the error handling makes the code cleaner.

A question, should this rather return `None` when the trueplay calibration is not done? This way it would be consistent with the handling of some other features (e.g., night_mode, dialog_mode which only raise an exception in setters).

pwt: Thanks. Yes, I think this is better, and I've made the change.

pwt:
Passes all check jobs locally, and has been independently tested, so I'm going to merge this.

PR #646, opened 2018-11-07
Add playbar/beam surround mode support

manugorre:
Allow turning the surround mode on/off for soundbar speakers.
Closes #461.

PR #780, opened 2020-12-27
Add properties for surround speakers in TV, music and ambient modes

### LINKED ISSUE #461: Surround Sound Mode
I have a playbar with a connect:amp configured as surround rear speakers.

[surround setup: https://sonos.custhelp.com/app/answers/detail/a_id/2248]
[as you can see from the link, it's quite tedious to configure them as surround rear speakers, and when I'm not watching any movies, to set them back as normal separate speakers.

I was wondering if it is possible to toggle surround sound mode?

rytilahti:
Adds following (setable) properties (see https://support.sonos.com/s/article/4804?language=en_US):
* ~~`surround_mode`: surround enabled/disabled~~ (already done in #870)
* `surround_ambient_enabled`: ambient/full volume in the music mode (bool)
* `surround_volume_tv`: surround volume adjustment [-15,15] for the tv mode (int)
* `surround_volume_music`: surround volume adjustment [15,15] for the music mode (int)

Currently missing reseting to defaults, and adjusting the surround speaker balance (which can only be done when trueplay is disabled, see https://github.com/SoCo/SoCo/pull/646#pullrequestreview-541049818).

Obsoletes #646. Fixes #461.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -14,6 +14,7 @@ from xml.sax.saxutils import escape
 from xml.parsers.expat import ExpatError
 import warnings
 import xmltodict
+from typing import Optional
 
 import requests
 from requests.exceptions import ConnectionError as RequestsConnectionError
@@ -156,6 +157,21 @@ def only_on_master(function):
     return inner_function
 
 
+def only_on_soundbars(function):
+    """Decorator that raises NotSupportedException when trying to use soundbar methods on unsupported devices."""
+
+    @wraps(function)
+    def inner_function(self, *args, **kwargs):
+        if not self.is_soundbar:
+            raise NotSupportedException(
+                f"The device is not a soundbar and doesn't support {function.__name__}."
+            )
+
+        return function(self, *args, **kwargs)
+
+    return inner_function
+
+
 # pylint: disable=R0904,too-many-instance-attributes
 class SoCo(_SocoSingletonBase):
 
@@ -238,6 +254,11 @@ class SoCo(_SocoSingletonBase):
         audio_delay
         night_mode
         dialog_mode
+        surround_mode
+        surround_mode_music_playback
+        surround_volume_tv
+        surround_volume_music
+        soundbar_audio_input_format
         supports_fixed_volume
         fixed_volume
         soundbar_audio_input_format
@@ -1190,12 +1211,10 @@ class SoCo(_SocoSingletonBase):
         if not self.is_soundbar:
             return None
 
-        response = self.renderingControl.GetEQ(
-            [("InstanceID", 0), ("EQType", "NightMode")]
-        )
-        return bool(int(response["CurrentValue"]))
+        return self.renderingControl.get_eq_variable("NightMode")
 
     @night_mode.setter
+    @only_on_soundbars
     def night_mode(self, night_mode):
         """Switch on/off the speaker's night mode.
 
@@ -1204,17 +1223,7 @@ class SoCo(_SocoSingletonBase):
         :raises NotSupportedException: If the device does not support
         night mode.
         """
-        if not self.is_soundbar:
-            message = "This device does not support night mode"
-            raise NotSupportedException(message)
-
-        self.renderingControl.SetEQ(
-            [
-                ("InstanceID", 0),
-                ("EQType", "NightMode"),
-                ("DesiredValue", int(night_mode)),
-            ]
-        )
+        self.renderingControl.set_eq_variable("NightMode", int(night_mode))
 
     @property
     def dialog_mode(self):
@@ -1225,12 +1234,10 @@ class SoCo(_SocoSingletonBase):
         if not self.is_soundbar:
             return None
 
-        response = self.renderingControl.GetEQ(
-            [("InstanceID", 0), ("EQType", "DialogLevel")]
-        )
-        return bool(int(response["CurrentValue"]))
+        return bool(self.renderingControl.get_eq_variable("DialogLevel"))
 
     @dialog_mode.setter
+    @only_on_soundbars
     def dialog_mode(self, dialog_mode):
         """Switch on/off the speaker's dialog mode.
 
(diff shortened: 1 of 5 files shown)


rytilahti:
Just a side note, after doing some testing (and XML checking), it looks like that changes to the audio input format is not sending events -> requires polling to get this information :-(

rytilahti:
@pwt do you mind giving this one a quick review? What do you think should be done to get this closer to get merged? Something besides fixing the docstrings & separating the `soundbar_audio_input_format`?

soco/core.py
```diff
@@ -129,6 +130,21 @@ def inner_function(self, *args, **kwargs):
...
 
 
+def only_on_soundbars(function):
+    """Decorator that raises NotSupportedException when trying to use soundbar methods on unsupported devices."""
+
+    @wraps(function)
+    def inner_function(self, *args, **kwargs):
+        if not self.is_soundbar:
+            raise NotSupportedException(
+                f"The device is not a soundbar and doesn't support {function.__name__}."
```
pwt, line 140: Using f-strings will break Python 3.5 compatibility. Could you replace with %, please?

soco/core.py
```diff
@@ -949,17 +956,88 @@ def dialog_mode(self, dialog_mode):
         :raises NotSupportedException: If the device does not support
         dialog mode.
         """
+        self.renderingControl.set_eq_variable("DialogLevel", int(dialog_mode))
+
+    @property
+    def surround_mode(self) -> Optional[bool]:
```
pwt, line 962: General comment: while it's great to see type hints, this is obviously not something found elsewhere in the SoCo code. However, I'm going to propose leaving them in, trading off their benefit for the inconsistency.

Note that there aren't type checks in the check jobs at present.

pwt:
> @pwt do you mind giving this one a quick review? What do you think should be done to get this closer to get merged? Something besides fixing the docstrings & separating the `soundbar_audio_input_format`?

This is a great. I've made a couple of comments: if you can push a new commit with the f-string replaced, that should initiate a check job run to spot any hygiene issues.

One problem I have is that I'm not able to test this code, since none of my Sonos equipment is set up in HT configs, so I'd love to get a some other folks to try it out.

soco/core.py
rytilahti: Yeah, I realized they are missing, but the good thing with them is that there is no need to do a full conversion for them to be useful.

soco/core.py
rytilahti: Done.

soco/services.py
```diff
@@ -858,6 +858,22 @@ def __init__(self, soco):
...
+
+    def set_eq_variable(self, variable: str, value: int):
+        """Set the EQ variable to given (integer) value."""
+        return self.SetEQ(
+            [
+                ("InstanceID", 0),
+                ("EQType", variable),
+                ("DesiredValue", value),
+            ]
+        )
```
rytilahti, line 875: I was trying to write some tests for all the functionality using this and realized that it's a bit more complicated as these methods are themselves magicmocks inside `moco`...

I'll have to see how to go around that limitation, in case it is fine to introduce these helpers (as indicated as an open question in the PR description) to reduce code duplication.

rytilahti:
Thanks @pwt! I added some responses in the comments, and also created a separate PR (#832) for the audio input format so that's why I force-pushed this branch.

I was working to add tests to the functionality when I realized that the helper methods I introduced makes testing the API interface more complicated thanks to indirection. I'll have to look for a suitable solution for that, I'm open for suggestions!

soco/core.py
pwt: Yep, I know, but EOL was less than a year ago, and while there's nothing pressurising us to deprecate 3.5 support we might as well continue supporting it.

pwt:
> I was working to add tests to the functionality when I realized that the helper methods I introduced makes testing the API interface more complicated thanks to indirection. I'll have to look for a suitable solution for that, I'm open for suggestions!

I may not have understood this fully, but I'd be minded to keep the unit tests simple and just pattern them after the existing tests for simple getters/setters.

rytilahti:
Oops, looks like I forgot to push this earlier, but I think this should be good to go now. Mind taking a look @pwt & @jjlawren?

I added tests to all surround properties (including the already existing `surround_enabled`) and renamed the `surround_music_playback` to `surround_ambient_enabled`.

edit: pre-commit config pins down to a version that's different from what's used by the CI, that's why all the hassle with force pushes, sorry for that. `pre-commit autoupdate` should be run to pin the version to the newest, opinions on whether it should be done in this PR or in a separate one?

pwt:
Hi @rytilahti . I've reviewed the code and it looks good. I've also pushed a couple of suggested commits that:

1. Fix a docstring reference error and tidy up some of the doctstring text.
2. Remove the type hints. (I'm an advocate of using typing, but I think we should maintain consistency with the rest of the SoCo codebase until such time as we decide to add typing more universally.)

Please review and check that the changes look OK to you.

Unfortunately, I'm not able to test this PR properly as I don't have any Sonos HT setups, so I'll have to depend on the testing that you and others are able to do.

rytilahti:
Hi @pwt and thanks for the review and the changes you made, looks good to me! I re-tested all the new settings on my Beam setup with the most current software and they work as expected.

pwt:
@rytilahti Great! Thanks for the contribution.

soco/core.py
```diff
@@ -1251,6 +1263,96 @@ def dialog_mode(self, dialog_mode):
...
+
+        response = self.renderingControl.GetEQ(
+            [("InstanceID", 0), ("EQType", "SurroundMode")]
+        )
+        return not int(response["CurrentValue"])
+
+    @surround_ambient_enabled.setter
+    @only_on_soundbars
+    def surround_ambient_enabled(self, value):
+        """Toggle surround music playback mode. True = ambient mode.
```
rytilahti, line 1286: I just realized that having the ambient mode as True makes using the eventing API awkward (ambient mode is report as `0`), so this may need to be changed to be other way around. IIRC I didn't do it previously as I couldn't think of a good name for the "full mode".

tests/test_core.py
```diff
@@ -1355,6 +1355,72 @@ def test_soco_balance(self, moco):
...
+        * For boolean values, check toggling and payloads
+        * For ranged int values, check the ranges and payloads.
+        """
+
+        def _assert_seteq_call(prop, variable, value, type):
+            desired_value = value
+            setattr(moco, prop, value)
+
+            if variable == "SurroundMode":  # negate desired value for surround mode
+                desired_value = not value
```
rytilahti, line 1380: This could be removed if the surround mode would be reversed.

PR #781, opened 2020-12-28
Add the ability to remove alarms by ID

rvhu-cl:
In my clockradio project I was missing the possibility to add and remove alarms by its ID. so add 2 functions so I can do this.
in the alarms.py I added remove_id and save_id.
Hope this is a usefull adding.


pwt:
Hi @rvhu-cl:

Thanks for contributing this enhancement.

In preparation for adding this PR to the upcoming 0.22 release, I've proposed some changes and pushed them to your branch. **Please review when you get a moment and check that my changes meet your intended use cases**.

Specifically, there was substantial redundancy in your `save_id()` and `remove_id()` methods, when compared to the existing `save()` and `remove()` methods. Also, the `remove_id()` method, where you force-set `_alarm_id` looked a bit odd to me.

So, I've reworked as follows:

1. There is now an `alarm_id` property as part of the `Alarm` class. This allows one to obtain the ID of any `Alarm` instance once it's been saved (or once it's been obtained by `get_alarms()`). This is a substitute for your `save_id()` method returning the ID.
2. There is now a `remove_alarm_by_id()` function that (as the name implies), removes an alarm with a given ID from the system.

pwt:
I've made a change so that 'Alarm.save()' now returns the `alarm_id` string. Although this is an interface change, it shouldn't break anything because previously it always returned `None`.

Issue #591, opened 2018-02-23
Deprecating support for Python 2.7


stefankoegl:
Separate issued branched off from [#337](https://github.com/SoCo/SoCo/issues/337#issuecomment-367275787).

Python 2.7 is the last supported 2.x version and requires some campatability code (#337). Python 2.7 will be maintained until 2020 (https://pythonclock.org/). 

This issue is to discuss our plans for supporting 2.7 for SoCo.

stefankoegl:
My suggest would be as follows
 I'd be open to deprecating it now, with the following implications
* deprecate support for 2.7 with the next release
* set a horizon for removing support with the release following the official end of support for 2.7 in 2020
* until then, continue to fix any bugs and maintain existing code and features
* for new features, if additional effort for achieving 2.7-comatability would be too high, support 3.x only (under the condition that no other features are broken)

relevitt:
Please don't do this! Many people use python 2.7 and have no immediate intention of changing. For example, Ubuntu 16.04 LTS (the version I run) uses python 2.7 as its default. There are significant libraries that haven't been ported to python 3.

stefankoegl:
> Many people use python 2.7 and have no immediate intention of changing.

That would not mean that people have to rewrite all of their Python software. It would only mean that you will need Python 3 for running SoCo.

> Ubuntu 16.04 LTS (the version I run) uses python 2.7 as its default. 

The default version is not really relevant. Ubuntu allows you to install multiple Python versions in parallel. In fact you can simply run `apt-get install python3` to install Python 3 in addition to Python 2.

> There are significant libraries that haven't been ported to python 3.

I tend to disagree on that one. https://python3wos.appspot.com/ - I wouldn't call any of those few red entries as significant.

KennethNielsen:
@stefankoegl you beat me on the wall of superpower.

@relevitt note that we aren't actually doing anything yet. But with regards to at least starting the discussion, the time is now. Not only has most (or all) significant projects been ported, but several projects have also now gone Python 3 only, because the want to use the new features and improvements. In the scientific community e.g. the Jupyter project if I remember correctly.

relevitt:
@stefankoegl:

- An entire 2.7 application that imports soco would have to be ported to python 3.3. I think the codebase of my application is bigger than the soco codebase, as it does many other things beside acting as a sonos controller. So it is a problem for me. Perhaps there are others like me. I'm not aware that other libraries I use are planning to drop 2.7 support.

[...]

PR #786, opened 2021-01-12
Updates to 'setup.py' for the 0.21 release

pwt:
- Update the supported Python versions
- Change `Development Status` to `4 - Beta`
- Apply Black formatting
- Add python_requires=">=3.5"

PR #788, opened 2021-01-12
Drop support for EOL Python 2.7 and 3.5

hugovk:
For https://github.com/SoCo/SoCo/issues/591.

The most important bits are to update the Trove classifiers and add `python_requires` to help pip.

Also upgrades Python syntax using https://github.com/asottile/pyupgrade and removes redundant mock backport and compat submodule and some other unnecessary bits.


pwt:
Hi @hugovk : this is great, thanks!

Given the scale of the proposed changes, I'm going to target the PR for the 0.22 release, not the imminent 0.21 release for which we're now in feature freeze. We'll take a closer look once 0.21 is released on 17th January.

I think we also need some group discussion about whether the time is right to remove support for Python 3.5, for which EOL was less than 6 months ago.

hugovk:
Yep, sounds sensible.

Is the aim for 0.21 to not work on 2.7? https://github.com/SoCo/SoCo/issues/591 said it already doesn't work on 2.7?

If so, it would make sense to have minimal changes for that: remove the 2.x Trove classifiers, and add `python_requires` excluding 2.7. That means modern pip won't only install 0.21 on Python 2.7, and will fall back to 0.20.

pwt:
> If so, it would make sense to have minimal changes for that: remove the 2.x Trove classifiers, and add `python_requires` excluding 2.7. That means modern pip won't only install 0.21 on Python 2.7, and will fall back to 0.20.

Excellent -- thanks. Pending PR #786 included the trove changes, and I've now added the python 3.5+ requirement.

pwt:
Returning to this after the release of 0.21.

A little bit of rework is required following the 0.21 changes. Also, let's retain Python 3.5 compatibility for the time being. So, please could you:

1. Start with current `master` and run `pyupgrade --py3-plus`
2. Run `black *.py soco/*.py tests/*.py` to make the formatting Black-compliant once again

This will result in code that passes all current SoCo check jobs, as well as the integration tests. (The two steps above are idempotent -- i.e., re-running pyupgrade will not make any changes.)

3. With the 0.21 changes, and if we stick with Python 3.5 compatibility, the following files no longer need to change: `.travis.yml`, `README.rst`.
4. I'd retain the Black import line as-is in `requirements-dev.txt`.
5. The other changes look good.

Take look at the above and let me know what you think.

Sorry for the extra work. This is a very nice 'cleanup' PR, and as it changes lots of files I'd like to get it in early in the v0.22 cycle, so other PRs can sync with it.

rytilahti:
Note that even Debian stable which is known for its long release cycles ships Python 3.7 already, so if some new features python 3.6+ are considered useful (type hints, f-strings come to my mind), I wouldn't try to keep 3.5 compatibility. :-)

pwt:
> Note that even Debian stable which is known for its long release cycles ships Python 3.7 already, so if some new features python 3.6+ are considered useful (type hints, f-strings come to my mind), I wouldn't try to keep 3.5 compatibility. :-)

Fair point, although Debian 9/Stretch, for which the default Python is 3.5, is supported until June 2022. (I still have machines running Stretch and SoCo).

Unless there's something compelling in 3.6 that we need (or badly want) for SoCo, I don't favour dropping support for 3.5 too hastily.

rytilahti:
> Fair point, although Debian 9/Stretch, for which the default Python is 3.5, is supported until June 2022. (I still have machines running Stretch and SoCo).

Oh, that's new to me, interesting! But indeed, stretch is separately supported by the LTS "subproject" (https://wiki.debian.org/LTS), and there's even ELTS (which supportes still jessie, too). Debian itself has EOL'd both of them already though (https://wiki.debian.org/DebianReleases). 

> Unless there's something compelling in 3.6 that we need (or badly want) for SoCo, I don't favour dropping support for 3.5 too hastily.

Sure thing, I personally like type hints & f-strings for my own personal projects, and use those wherever I can. In the end, I don't personally mind that much, I just thought I'd mention it when I saw this issue in my notifications :-)

hugovk:
Updated! How does this look?

pwt:
> Updated! How does this look?

Looks good -- thanks!

- I've just cleared up some pylint and flake8 issues and pushed the changes to your branch.
- Passes unit and integration tests on Python 3.9.1 and Python 3.5.10.

@KennethNielsen: I'd appreciate if you could take a quick look at this, as it touches a large numbers of files. (In that regard, I think it's similar to the Black formatting changes you made a while ago.) Clears up a lot of Python 2 cruft.

pwt (approved):

pwt:
Merged. Thanks to @hugovk for initiating this major tidy-up.

PR #795, opened 2021-01-22
Do not refresh zone group cache on join/unjoin

amelchio:
We currently reload the zone group cache after group changes (join/unjoin). As this is a synchronous operation, it actually slows the client down.

If we end up using the cache, the time we save has already been spent filling it. If we do not use the cache within 5 seconds, filling it was redundant.

Also note that group changes often happen to multiple speakers in turn and we are currently reloading the cache after each such operation.

So it seems that _not_ reloading the cache is always at least as good. Thus, this PR removes the preloading.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1243,7 +1243,6 @@ class SoCo(_SocoSingletonBase):
             ]
         )
         self._zgs_cache.clear()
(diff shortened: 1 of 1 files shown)


pwt:
LGTM.

PR #796, opened 2021-01-22
Catch client exceptions on event thread

amelchio:
This keeps our event thread from dying due to uncaught exceptions in the client event handling code.


pwt:
LGTM.

PR #797, opened 2021-01-22
Always restore volume

amelchio:
Due to races, our idea of a speaker being a coordinator could be wrong. In that case, restoring its queue will fail. This PR ensures that this failure does not keep the volume from being restored.

I have split this change into three commits to make it easier to review.


pwt:
LGTM.

Issue #778, opened 2020-12-11
PR help request: Adding direct Plex control support


jjlawren:
After a lot of packet sniffing and trial and error, I'm now able to initiate playback of media served from my Plex server. However, I'm not sure of the best way to add clean support of this feature back into SoCo. I'd appreciate any help or suggestions based on what I'll share below.

The Sonos app allows you to browse your Plex library by Artist->Album->Track. Once you select a track, it replaces the current queue with the track's album with playback starting at the chosen track. It does this with the "ReplaceAllTracks" command as noted [here](https://github.com/SoCo/SoCo/wiki/Sonos-UPnP-Services-and-Functions#queue).

The required arguments follow this pattern:
```python
args = [
    ("QueueID", 0),
    ("UpdateID", 0),
    ("ContainerURI", ""),
    ("ContainerMetaData", ""),
    ("CurrentTrackIndex", 0),
    ("NewCurrentTrackIndices", TRACK_NUMBER),
    ("NumberOfURIs", 1),
    ("EnqueuedURIsAndMetaData", METADATA),
]
```
The `METADATA` payload is XML with the following structure:
```xml
<URIs>
  <URI uri="x-rincon-cpcontainer:{ENCODED_ALBUM_ID}?sid=212&amp;flags=8300&amp;sn=9">
    <DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/">
      <item id="{ENCODED_ALBUM_ID}" parentID="{ENCODED_ARTIST_ID}" restricted="true">
        <dc:title>{ALBUM_TITLE}</dc:title>
        <upnp:class>object.container.album.musicAlbum.#titleSummary</upnp:class>
        <upnp:album>{ALBUM_TITLE}</upnp:album>
        <dc:creator>{ARTIST_NAME}</dc:creator>
        <upnp:albumArtURI>{URL_TO_ALBUM_ART}</upnp:albumArtURI>
        <r:description>{ARTIST_NAME}</r:description>
        <desc id="cdudn" nameSpace="urn:schemas-rinconnetworks-com:metadata-1-0/">SA_RINCON12345_X_#Svc12345-0-Token</desc>
      </item>
    </DIDL-Lite>
  </URI>
[...]

PR #813, opened 2021-02-26
Plex plugin

jjlawren:
This is a first attempt at addressing #778 to add support for playback using the Plex service. I'm somewhat unfamiliar with the various data structures, so I welcome any suggestions on how to better implement this in the library.

This requires the Sonos system to be linked to the same Plex server/account which is used by the example script. This also implies that the Sonos system can reach the Plex server over the network at the provided `baseurl` address.

The example script uses the [plexapi](https://github.com/pkkid/python-plexapi) library to simplify gathering some of the internal Plex identifiers.


pwt:
Tentatively milestoning this feature for v0.22.

jjlawren:
I don't think this is quite ready for prime time yet. It was suggested this could be moved into a plugin but I haven't had time to attempt that yet.

jjlawren:
After taking a fresh look at this with some more traffic sniffing, I realized this can be accomplished as a plugin without any core changes. This requires use of the `plexapi` library and currently supports adding tracks/albums/artists/playlists to the queue. There's also a convenience method to insert the new media next in the queue & immediately begin playback. This preserves the existing playback queue on the speaker which will resume after the new media finishes playing.

The command to replace all tracks in the queue will require some small changes to core and can be done in a followup PR.

pwt:
Apologies that I'd missed the removal of the 'draft' tag on this. Looks great! I'd like to get it in to 0.25.0, due for release shortly.

It's working fine for me with light testing, although there's a usage wrinkle with `plexapi` that I don't understand. Specifically, the following doesn't work, even though I definitely have a `Music` section in the library:
```
music = plex.library.section("Music")
```
Instead I have to use the name of my Plex server:
```
music = plex.library.section("pi4-plex")
``` 
It works fine from then on. Clearly, this is purely a `plexapi` wrinkle and not related to the plugin. Perhaps it's something to do with using username/password auth instead of a token.

I'm assuming we don't want to add `plexapi` to the standard imports.

We need to add the plex plugin to the api docs section, and make some changes to the docstrings to address sphinx issues. I'll aim to make the changes and push to your repo.

jjlawren:
> Instead I have to use the name of my Plex server

Hmm, that doesn't make sense to me. The `plex` object here represents the `PlexServer` object. The string you're passing must match an existing library section title or it would throw an exception: https://github.com/pkkid/python-plexapi/blob/a4976137d2578b9c62b5f0e6d16f439792700d72/plexapi/library.py#L56

> I'm assuming we don't want to add plexapi to the standard imports.

No, not for a plugin.

pwt:
> Hmm, that doesn't make sense to me. The `plex` object here represents the `PlexServer` object. The string you're passing must match an existing library section title or it would throw an exception: https://github.com/pkkid/python-plexapi/blob/a4976137d2578b9c62b5f0e6d16f439792700d72/plexapi/library.py#L56

Doesn't make sense to me either, but here's what I get:
```
>>> account = MyPlexAccount("<username>", "<password>")  # Redacted
>>> plex = account.resource("raspberrypi-4").connect()
>>> type(plex)
<class 'plexapi.server.PlexServer'>
>>> plex.library.sections()
[<MusicSection:3:Pi4-Plex>]
>>> music = plex.library.section("Music")
  File "<input>", line 1, in <module>
  File "/Users/pwt/pythons/SoCo/lib/python3.9/site-packages/plexapi/library.py", line 67, in section
    raise NotFound('Invalid library section: %s' % title) from None
plexapi.exceptions.NotFound: Invalid library section: Music
>>> music = plex.library.section("pi4-plex")
>>> type(music)
<class 'plexapi.library.MusicSection'>
```
Once I have the `MusicSection` object, everything works fine.

As I said, this is not directly relevant to the plugin, just documenting for completeness. I'll do some further digging if I get time.

jjlawren:
> >>> plex.library.sections()
> [<MusicSection:3:Pi4-Plex>]

It appears that's the actual name of your music library section. It should show this label if you browse your media using a Plex client.

PR #814, opened 2021-03-05
Improve handling of network errors

amelchio:
Add timeout values so network requests cannot hang indefinitely.

Also ignore a couple of error situations during unsubscribe because no error handling is needed in those situations.

These issues were found by running a network conditioner with pretty bad settings.

### DIFF
--- a/soco/events.py
+++ b/soco/events.py
@@ -383,8 +383,20 @@ class Subscription(SubscriptionBase):
                 of response headers as its only parameter.
 
         """
-        response = requests.request(method, url, headers=headers)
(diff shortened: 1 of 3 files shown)


pwt:
Great -- thanks for this.

Oddly, I was looking at exactly the same issue a couple of days ago, and trying to settle on what the default timeouts should be. Your choices look good to me. I think the longest-lived actions relate to library searches with a large number of returned values.

amelchio:
We have been using these values for 18+ months in Home Assistant with no reported issues so I think they are at least not too low.

PR #820, opened 2021-04-17
Prevent binary installs of 'multidict' and 'yarl' (fixes 819)

pwt:
This change forces a Python-only installation of the `multidict` and `yarl` packages, to prevent installation issues on some platforms on which binary wheels need to be built as part of the installation.


pwt:
Nope, this approach is not correct.

PR #827, opened 2021-05-14
iss826: added EndDirectControlSession to core.py avtransport class call

amaten69:


pwt:
Hi @amaten69: 

Thanks for the PR. I've made a minor name change to the method, and added a unit test. Let me know if it looks OK.

PR #829, opened 2021-05-17
Raise exceptions on aiohttp status in async event handler

jjlawren:
The current async events module does not actually raise `aiohttp` exceptions inside of `try:` blocks as the `raise_for_status` argument was missing. This can lead to problems such as failed subscription renewals being silently ignored.


jjlawren:
Identical to https://github.com/amelchio/pysonos/pull/87

bdraco (approved):

pwt:
LGTM. Thanks for upstreaming.

pwt:
Cherry-picked to v0.22.fixes.

PR #831, opened 2021-05-17
Extract SNG metadata

amelchio:
This adds extraction of `TYPE=SNG` metadata in radio streams. I don't know exactly which standard this is but it has been reported being using at least with SiriusXM and Apple Music.

The first commit moves the existing extraction into a dedicated function to avoid a too-many-branches Pylint warning.


pwt:
LGTM -- thanks.

pwt:
Cherry-picked to v0.22.fixes.

PR #833, opened 2021-05-18
Cleanup async_events renewal logging and exceptions

jjlawren:
The logging and exception handling was a bit noisy and redundant in the new async event handler.

This change will not make subscription auto-renew calls `strict`. This matches the behavior in `events_twisted.py`: https://github.com/SoCo/SoCo/blob/bb71285d5d9255a2eca942b357f292d613891f93/soco/events_twisted.py#L325-L329)

If a call to the `auto_renew_fail` callback is made, we will no longer log an exception traceback from `SoCo` as it's assumed the caller now has that information.


bdraco (approved):
Looks good. Makes sense to me especially for the users with flakey wifi

pwt:
Cherry-picked to v0.22.fixes. Will be included in v0.22.3.

Issue #792, opened 2021-01-20
SoCo raising exception when playing from YouTube Music|


Barry4679:
I am subscribed to SoCo AVT events. 
If my Sonos is playing from YouTubeMusic it generates the AVT event OK, but logs the following exception to the log for each track played.

01-20 23:28:15 158.0095 - WARNING soco.events_base:  110 parse_event_xml; Event contains illegal metadatafor 'enqueued_transport_uri_meta_data'.
Error message: 'UPnP class is incorrect. Expected 'object.container.album.musicAlbum.ALBUM', got 'object.container.album.musicAlbum''
The result will be a SoCoFault.

I added a message to display the data causing  the problem:
01-20 23:28:15 159.0295 - WARNING soco.events_base:  118 parse_event_xml; data value was <DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/"><item id="1004004cALkSOiGdMl9FQFJGE1Ic8S2rmLFHkNfJ69CFi-KR6SopsRn2" parentID="10fe1044ALkSOiEkIOVrL0HH1ihrgR_Gj1GTz0ZqOUgv4vwS1aNttlQt6jwZTkbBPq7CgzFt3E_p20LznoFNX_xmBB3JTSKcZIve9E98FBLjXTA4m6U3xXATU1i3H_DU6ym61uEu" restricted="true"><dc:title>Schnittke: Tango (Arr. by Andriy Rakhmanin for Violin and Piano) (From &quot;Agony&quot;)</dc:title><upnp:class>object.container.album.musicAlbum.#ALBUM</upnp:class><upnp:album>Schnittke: Tango (Arr. by Andriy Rakhmanin for Violin and Piano) (From &quot;Agony&quot;)</upnp:album><dc:creator>Daniel Hope, Alexey Botvinov</dc:creator><upnp:albumArtURI>https://lh3.googleusercontent.com/ictx7j6cS24dlX-MTtEYFoHrORicbEabuFfpgSS4Le9P8g0mnrTukCCNVeK7zF-lHAz
[...]

PR #835, opened 2021-05-22
Adjust log levels

amelchio:
This downgrades some logging statements so SoCo does not log above the DEBUG level in normal use.

A single log level has been increased since it could be a real problem (client code raising an exception while handling an event).

Some related discussion in #792.


pwt:
LGTM -- thanks.

amelchio:
Thank **you**, that was quick!

PR #837, opened 2021-05-23
Add get_current_media_info() method to the SoCo class

amelchio:
This adds a `get_current_media_info()` method which is very similar to `get_current_track_info()`, but for the `GetMediaInfo` service.

We are using this in Home Assistant to get the channel name when playing radio. The `uri` is useful for finding out whether a favorite is playing since it matches those, unlike `get_current_track_info()['uri']`.

Screenshot

No other information from `GetMediaInfo` seemed interesting, here is an example of the full data:
```
{'CurrentURI': 'x-sonosapi-stream:s216726?sid=254&flags=8224&sn=0',
 'CurrentURIMetaData': '<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" '
                       'xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" '
                       'xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" '
                       'xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/"><item '
                       'id="-1" parentID="-1" '
                       'restricted="true"><dc:title>myROCK</dc:title><upnp:class>object.item.audioItem.audioBroadcast</upnp:class><desc '
                       'id="cdudn" '
                       'nameSpace="urn:schemas-rinconnetworks-com:metadata-1-0/">SA_RINCON65031_</desc></item></DIDL-Lite>',
 'MediaDuration': 'NOT_IMPLEMENTED',
 'NextURI': '',
 'NextURIMetaData': '',
 'NrTracks': '2',
 'PlayMedium': 'NETWORK',
 'RecordMedium': 'NOT_IMPLEMENTED',
 'WriteStatus': 'NOT_IMPLEMENTED'}
```

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -319,7 +319,7 @@ class TestGetCurrentTrackInfo:
     )
 
     def test_get(self, soco):
-        """Test is the return value is a dictinary and contains the following
+        """Test that the return value is a dictionary and contains the following
         keys: album, artist, title, uri, playlist_position, duration,
         album_art and position.
         """
@@ -339,7 +339,7 @@ class TestGetCurrentMediaInfo:
     )
 
     def test_get(self, soco):
-        """Test that the return value is a dictinary and contains the expected
+        """Test that the return value is a dictionary and contains the expected
         keys.
         """
         info = soco.get_current_media_info()
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -328,6 +328,25 @@ class TestGetCurrentTrackInfo:
         assert sorted(info.keys()) == self.info_keys
 
 
+class TestGetCurrentMediaInfo:
+    """Integration test for the get_current_media_info method."""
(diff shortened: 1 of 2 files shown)


tests/test_integration.py
```diff
@@ -328,6 +328,25 @@ def test_get(self, soco):
...
+
+    info_keys = sorted(
+        [
+            "uri",
+            "channel",
+        ]
+    )
+
+    def test_get(self, soco):
+        """Test that the return value is a dictinary and contains the expected
```
pwt, line 342: Minor typo: 'dictinary'.

amelchio: Thanks. I fixed that, also for the `TestGetCurrentTrackInfo` where I have copied from :-)

pwt:
LGTM -- thanks.

PR #841, opened 2021-06-07
Fix unsubscription logic

jjlawren:
I noticed that unsubscription messages are _never_ sent to speakers because of the ordering of guards in `soco.events_base.unsubscription`.

Once `_cancel_subscription()` is called, the `_timestamp` attribute is set to `None`. This in turns causes the `time_left` property to always return 0. This aborts the unsubscription call early and the outgoing request is never fired.

### DIFF
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -546,12 +546,12 @@ class SubscriptionBase:
         if self._has_been_unsubscribed or not self.is_subscribed:
             return None
 
-        self._cancel_subscription()
-
         # If the subscription has timed out, an attempt to
         # unsubscribe from it will fail silently.
         if self.time_left == 0:
             return None
+
+        self._cancel_subscription()
 
         # Send an unsubscribe request like this:
         # UNSUBSCRIBE publisher path HTTP/1.1


jjlawren:
It also seems that the call to `_cancel_subscription` can cause the event listener to stop prematurely. Once this occurs I'm not seeing the remaining unsubscription requests sent out.

jjlawren:
Moved the internal `Subscription`/`SubscriptionsMap` cleanup method after the successful unsubscription request.

soco/events_base.py
```diff
@@ -546,8 +546,6 @@ def unsubscribe(self):
         if self._has_been_unsubscribed or not self.is_subscribed:
             return None
 
-        self._cancel_subscription()
```
bdraco, line 549: If the speaker is now offline, `self._cancel_subscription()` will never be called with this change. which means we will not cancel auto renew attempts.

bdraco: Maybe we need a `self._cancel_subscription_success` or the similar that sets `self._timestamp` ?

jjlawren: Added an optional `unconditional` parameter that takes a function in addition to `success`. It's always called after the request is complete. For now, only utilized by the unsubscription calls in both event handlers.

bdraco (approved):

pwt:
When I run this against SoCo-CLI, I'm getting an exception in event unsubscription:
```
2021-06-08 09:59:06,498 events_base.py:664 - _cancel_subscription() - An Exception occurred. Subscription to http://192.168.0.39:1400/MediaRenderer/AVTransport/Event, sid: uuid:RINCON_B8E93790A1A001400_sub0000000634 has been cancelled
2021-06-08 09:59:06,498 api.py:81 - run_command() - Exception: _request() takes 5 positional arguments but 6 were given
```

My code simply calls `sub.unsubscribe()` on a subscription object.

I think the defect is hit in `events.py` line 434, with the exception caught in line 436.

pwt:
Ok, the issue is that `events.py` was not updated to fit the new pattern. The following changes are required:

```
diff --git a/soco/events.py b/soco/events.py
index 667943b..4c6c670 100644
--- a/soco/events.py
+++ b/soco/events.py
@@ -370,7 +370,7 @@ class Subscription(SubscriptionBase):
         self._auto_renew_thread_flag.set()
 
     # pylint: disable=no-self-use, too-many-arguments
-    def _request(self, method, url, headers, success):
+    def _request(self, method, url, headers, success, unconditional=None):
         """Sends an HTTP request.
 
         Args:
@@ -381,6 +381,9 @@ class Subscription(SubscriptionBase):
             success (function): A function to be called if the
                 request succeeds. The function will be called with a dict
                 of response headers as its only parameter.
+            unconditional (function): An optional function to be called after
+                the request is complete, regardless of its success. Takes
+                no parameters.
 
         """
         response = None
@@ -400,6 +403,9 @@ class Subscription(SubscriptionBase):
         if success:
             success(response.headers)
 
+        if unconditional:
+            unconditional()
+
     # pylint: disable=inconsistent-return-statements
     def _wrap(self, method, strict, *args, **kwargs):

```

@jjlawren: Please could you review and include these changes, and resubmit a PR? Thanks.

jjlawren:
Apologies, I forgot about the original. Will re-submit a new PR.

PR #842, opened 2021-06-08
Fix event unsubscription logic v2

jjlawren:
The currrent event handlers are not able to send unsubscription requests to speakers because of a logic issue. This can lead to unintentionally leaving subscriptions active.

This change defers the internal unsubscription cleanup steps until after the outbound request is sent. This is achieved by adding an `unconditional` parameter to each implementations' request handler which is called after any request is sent. Only the `unsubscribe` method uses this for now.

This is an updated PR of #841 which adds missing changes in the default event handler.


jjlawren:
I also scanned through the documentation and didn't see anything that needed to be updated.

PR #844, opened 2021-07-07
Add pre-commit hooks option; code/docstring tidy-ups

rytilahti:
pre-commit is the go-to method for making it simple to verify that all linters and other checks are performed before a commit can be done, or when executing `pre-commit run -a` manually. Having it in-place makes it easier to contribute without a need to wait for the CI to inform about styling issues.

This PR adds the same checks (black, pylint, flake8) as running `make` does (hopefully), but also removes the blanket ignores for blake8 and uses `# noqa: <id>` for two cases (import order for docs, unused variable for propertymocks for tests). This PR also includes the changes wanted by the checks that were not caught previously for some reason.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,11 @@ repos:
   - id: black
     language_version: python3
 
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.9.2
+  hooks:
+  - id: flake8
+
 - repo: https://github.com/pycqa/pylint
   rev: v2.9.3
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,22 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.0.1
+  hooks:
+  - id: check-yaml
+  - id: check-json
+  - id: debug-statements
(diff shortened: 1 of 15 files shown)


pwt:
Presumably, we also need to add the `pre-commit` package to `requirements-dev.txt`? And to run `pre-commit install` locally to set up the hooks?

rytilahti:
If that'd be the preferred way to run the checks, it makes indeed sense to add pre-commit to the dev requirements & instruct contributors to use it by adding a note to the developer documentation. On the other hand, it is also helpful to have the configuration available even if it is not instructed anywhere.

What do you think?

pwt:
> If that'd be the preferred way to run the checks, it makes indeed sense to add pre-commit to the dev requirements & instruct contributors to use it by adding a note to the developer documentation. On the other hand, it is also helpful to have the configuration available even if it is not instructed anywhere.
> 
> What do you think?

I thought about this after making the comment, and I think it's fine just to include the yaml file (and the miscellaneous
changes) without adding to `requirements-dev.txt`. It's then there as a convenience for people that want to use `pre-commit`, but doesn't affect this that don't.

Note that I'll hold off merging this until after v0.23 is released.

pwt:
Thanks for all the work on this, and sorry for the delay in including it. Will be included in v0.26.0.

PR #848, opened 2021-07-16
Hardcode service numbers for ShareLink plugin

### LINKED ISSUE #847: Spotify ShareLink plugin fails on Sonos S2 (UPnP 800)
Using the new ShareLink plugin (#838) in SoCo v0.23.0, I find that it works for Spotify on Sonos S1 (11.2.9) but fails on Sonos S2 (13.1.4), returning a UPnP 800 error.

I'm in the UK, using a Spotify Premium subscription.

@amelchio : Any ideas? Happy to perform debugging suggestions.

amelchio:
@pwt I expect this to fix #847, can you please test?

It now hardcodes the service numbers and tries both US and global Spotify services in turn because there is apparently no reliable way to find the proper one.

### DIFF
--- a/soco/plugins/sharelink.py
+++ b/soco/plugins/sharelink.py
@@ -3,16 +3,11 @@
 import re
 
 from ..plugins import SoCoPlugin
-from ..music_services import MusicService
 from ..exceptions import SoCoException
 
 
 class ShareClass:
     """Base class for supported services."""
-
-    def __init__(self, soco):
-        """Initialize the share object."""
-        self.soco = soco
 
     def canonical_uri(self, uri):
         """Recognize a share link and return its canonical representation.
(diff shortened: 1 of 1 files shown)


pwt:
Great -- thanks for this.

Works for me with my S1 and S2 systems, although note that I can only test against Spotify/UK.

amelchio:
I would love to know the difference between your systems, maybe we could avoid this hardcoding after all.

However, as this PR does seem like an improvement with our current knowledge, I will remove the draft status.

pwt:
> I would love to know the difference between your systems, maybe we could avoid this hardcoding after all.

I'll see if I can figure out what's happening.

> However, as this PR does seem like an improvement with our current knowledge, I will remove the draft status.

Thanks. Sounds good for now.

PR #851, opened 2021-08-02
Refactor alarm handling

jjlawren:
When using the `soco.alarms` module in tandem with subscription callback events, I found it was difficult to know if the current state of the alarms was accurate. When an alarm is changed and events are sent out (from each speaker), it can take some time for updated alarms to propagate to all speakers. The `zone.alarmClock.ListAlarms()` call returns a `CurrentAlarmListVersion` which can be matched against the same one provided in the event payload, but this does not exist in the current implementation.

While adding support for this value, I found it was easier to think of things in terms of a parent storage class (`Alarms`) which contains all `Alarm` instances instead of using the `weakref` class attribute.

This PR adds new functionality while attempting to preserve existing behavior:
* Adds a new `Alarms` storage class to hold all `Alarm` instances (acts as singleton like `SoCo` instances)
* Parses and stores `CurrentAlarmListVersion` to allow synchronizing with event callbacks
* Throttles unnecessary updates using `CurrentAlarmListVersion`
* Attempts to preserve `Alarm` instances by updating them in place (legacy behavior)
* ~~Adds typing to all methods and attributes~~

Breaking changes:
* ~~`Alarm.alarm_id` is an `int` instead of `str`~~

Non-breaking changes to existing behavior:
* `Alarm.remove()` returns a `bool` to indicate success

### DIFF
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -95,7 +95,7 @@ class Alarms(_SocoSingletonBase):
 
     def __init__(self) -> None:
         """Initialize the instance."""
-        self.alarms: Dict[int, Alarm] = {}
+        self.alarms: Dict[int, 'Alarm'] = {}
         self._last_zone_used: Optional[SoCo] = None
         self._last_alarm_list_version: Optional[str] = None
         self.last_uid: Optional[str] = None
@@ -191,10 +191,10 @@ class Alarm:
         self,
         zone: SoCo,
         start_time: Optional[datetime.time] = None,
-        duration: datetime.time | None = None,
+        duration: Optional[datetime.time] = None,
         recurrence: str = "DAILY",
         enabled: bool = True,
-        program_uri: str = None,
+        program_uri: Optional[str] = None,
         program_metadata: str = "",
         play_mode: str = "NORMAL",
         volume: int = 20,
@@ -372,7 +372,7 @@ class Alarm:
         return self._alarm_id
 
 
-def get_alarms(zone: Optional[SoCo] = None) -> Set[Alarm]:
+def get_alarms(zone: Optional[SoCo] = None) -> Set['Alarm']:
     """Get a set of all alarms known to the Sonos system.
 
     Args:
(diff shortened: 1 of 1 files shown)


pwt:
> Breaking changes:
> 
> * `Alarm.alarm_id` is an `int` instead of `str`

I'm concerned that this will cause widespread breakages, and I'd like to explore ways of retaining backward compatibility. I appreciate that this could make the code suboptimal.

jjlawren:
> I'm concerned that this will cause widespread breakages, and I'd like to explore ways of retaining backward compatibility. I appreciate that this could make the code suboptimal.

To clarify, the values used remain the same, but they're now cast to `int`.

pwt:
> > I'm concerned that this will cause widespread breakages, and I'd like to explore ways of retaining backward compatibility. I appreciate that this could make the code suboptimal.
> 
> To clarify, the values used remain the same, but they're now cast to `int`.

Yes, I realise that. The type change will break code, though, as you noted. (It would break my code, for example.)

jjlawren:
Fair point. I can revert that change after I make all the CI versions happy.

jjlawren:
Breaking change has been reverted. I've also removed type hinting for now as using the "old" pre-3.7 hinting is messy and no other code in the library uses it yet.

pwt:
> Breaking change has been reverted. I've also removed type hinting for now as using the "old" pre-3.7 hinting is messy and no other code in the library uses it yet.

Thanks for retaining backward compatibility.

I gave the refactored code a quick test run against the alarms actions in my code ([SoCo-CLI](https://github.com/avantrec/soco-cli#alarms)), and it all worked fine.

jjlawren:
I've refactored the Home Assistant integration against the current state of this PR and I haven't run into any issues. I'm now able to coordinate and throttle alarm updates based on event callbacks across multiple speakers.

pwt:
LGTM -- thanks.

PR #852, opened 2021-08-05
Add boot count attribute to zones

jjlawren:
Having the boot sequence number on a zone can be useful for determining when a given zone has lost power or is otherwise rebooted. This information is available from SSDP & Zeroconf discovery broadcasts, but this is the only way I've been able to find it in direct queries to the speaker.


pwt:
LGTM, assuming you're OK with my additions.

jjlawren:
> LGTM, assuming you're OK with my additions.

Absolutely, thanks for polishing it up a bit.

PR #853, opened 2021-08-06
Clean up subscription if unsubscription request fails

jjlawren:
If an unsubscription request fails, the `Subscription` object is not properly cleaned up. This can lead to dangling failed subscriptions which conflict with attempts to recreate new subscriptions.

The call to `_cancel_subscription` is currently called in request failure scenarios for both subscriptions and renewals, but not after unsubscription request failures:

https://github.com/SoCo/SoCo/blob/5602d2880af8a3daa9158a2982fca8130d41b13f/soco/events_asyncio.py#L367-L368

https://github.com/SoCo/SoCo/blob/5602d2880af8a3daa9158a2982fca8130d41b13f/soco/events_asyncio.py#L418-L419

### DIFF
--- a/soco/events_asyncio.py
+++ b/soco/events_asyncio.py
@@ -452,6 +452,7 @@ class Subscription(SubscriptionBase):
                 return
             await unsub
         except Exception as exc:  # pylint: disable=broad-except
+            self._cancel_subscription(exc)
             if strict:
                 raise
             self._log_exception(exc)
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -661,7 +661,7 @@ class SubscriptionBase:
         # Cancel any auto renew
         self._auto_renew_cancel()
         if msg:
-            log.info(msg)
+            log.debug(msg)
 
     @property
     def time_left(self):


soco/events_asyncio.py
```diff
@@ -452,6 +452,7 @@ async def unsubscribe(
                 return
             await unsub
         except Exception as exc:  # pylint: disable=broad-except
+            self._cancel_subscription(exc)
```
bdraco, line 455: This mirrors the exception handling in `subscribe`.

Throwing away the subscription makes sense here since its clearly not usable
if we get an exception.

bdraco (approved):

jjlawren:
Along with this I think it would be prudent to downgrade this message level to debug:
https://github.com/SoCo/SoCo/blob/5602d2880af8a3daa9158a2982fca8130d41b13f/soco/events_base.py#L664

pwt:
> Along with this I think it would be prudent to downgrade this message level to debug:
> 
> https://github.com/SoCo/SoCo/blob/5602d2880af8a3daa9158a2982fca8130d41b13f/soco/events_base.py#L664

LGTM. Do you want to make the log severity change as part of this PR, or as a separate one? If the latter, I'll proceed with the merge.

Issue #790, opened 2021-01-17
Reinstate Check Jobs for SoCo Development


pwt:
Travis CI withdrew the ability for open source projects to run check jobs, during the development of SoCo v0.21.

It's important that we reinstate an alternative process, otherwise SoCo development velocity will be significantly impeded and there'll need to be excessive manual testing.

rytilahti:
Take a look at https://hynek.me/articles/python-github-actions/ (for github actions) or https://hynek.me/articles/simple-python-azure-pipelines/ (for azure pipelines, which requires quite a bit more effort but is not bound to github).

KennethNielsen:
I'm already working on a tox configuration for SoCo, which makes it easier to setup Github Actions. I will post a PR shortly.

pwt:
As a temporary measure, I've instantiated a 'brute force' Github actions workflow that performs the same checks as the previous Travis job, across Python 3.5 to 3.9. It appears to be working correctly for PRs and merges to master.

This can be junked when we have a tox configuration.

PR #854, opened 2021-08-13
Use matrix in CI config

jjlawren:
Instead of duplicating the test configs, loop through using a matrix and specific job exceptions.

Bumps the 3.10 version to the latest available, and also disables building docs on 3.10 as https://github.com/sphinx-doc/sphinx/issues/9505 is an open issue causing CI failures.

### DIFF
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -12,16 +12,19 @@ on:
       - 'v0.23.fixes'
 
 jobs:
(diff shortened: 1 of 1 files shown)


jjlawren:
It should be possible to improve the CI performance a bit with venv caching, but that can be done in a followup.

pwt:
Excellent -- I'm glad someone tidied this up! Thanks.
I'm going to declare that this fixes #790.

PR #855, opened 2021-08-13
Remove Python 3.5 support

jjlawren:
Python 3.5 has been considered EOL since September 2020 and `requests` (a primary dependency) has dropped support as of [2.26.0](https://docs.python-requests.org/en/latest/community/updates/#id1). This removes a workaround for handling `black` in the CI.

Dropping support for 3.5 will allow use of some newer features, probably most visible being [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings).

### DIFF
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.7, 3.8, 3.9, "3.10"]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9, "3.10"]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10"]
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
@@ -42,7 +42,6 @@ jobs:
         run: |
           pytest --cov-config .coveragerc --cov soco .
       - name: Test formatting with Black
-        if: matrix.python-version != 3.5
         run: |
           black --check soco tests
       - name: Test documentation build
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ would be appreciated.**
 Installation
 ------------
 
-SoCo requires Python 3.7 or newer.
+SoCo requires Python 3.6 or newer.
 
 Use pip:
 
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ would be appreciated.**
 Installation
 ------------
 
-SoCo requires Python 3.5 or newer.
+SoCo requires Python 3.6 or newer.
 
 Use pip:
 
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ would be appreciated.**
 Installation
 ------------
 
-SoCo requires Python 3.6 or newer.
+SoCo requires Python 3.7 or newer.
 
 Use pip:
 
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,6 +7,6 @@ pylint
 coveralls
 pytest-cov<2.6.0
 wheel
-black >= 20.0b0; python_version >= "3.6" and implementation_name == "cpython"
+black >= 20.0b0
 requests-mock
 twine
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ CLASSIFIERS = [
     "Intended Audience :: Developers",
     "License :: OSI Approved :: MIT License",
     "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3.6",
     "Programming Language :: Python :: 3.7",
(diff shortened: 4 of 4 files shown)


pwt:
Personally, I think this needs a bit of discussion as to the timing for making this change. While I understand the 'keeping up to date' argument, I don't see any current, compelling reason that would drive SoCo to drop 3.5 support.

Note that Debian 9/Stretch has LTS support for 3.5 through to June 2022.

@KennethNielsen: any thoughts?

jjlawren:
Agree, I wanted to get this on the radar. With `requests` dropping 3.5 support it seemed like a good time to start discussing.

jjlawren:
I'm working on adding Python venv caching to the CI runs and 3.5 support is a blocker. To ensure the CI is always testing with the latest release, it's necessary to pin each library's version in `requirements-dev.txt` as the hash of that file is used as a cache key. However if we pin to the latest release, several of the libraries no longer support <3.6 and will fail to install. This includes `requests` and `black`, possibly others.

jjlawren:
Note that once #857 is merged, we can rebase this PR and clean up the changes to `requirements-dev.txt` which the other PR introduces.

KennethNielsen:
@pwt @jjlawren I think we should probably develop make a decision on what we want to with support of older version in general, so that we have that to lean on the next time this problem pops up. From my point of view, the answer is somewhere in between the things you have already discussed.

As @pwt, I would prefer the measuring stick for which version we do support to be the oldest python versions that LST distributions support _in as far as that can be achieved with reasonable effort_. So, personally I would be willing to compromise on it, the the burden is too high, but I too have heard of clever version pinning tools that apply to the dependencies dependencies, so if we could set one of those up **and** note in internal documentation, which distributions old python is limiting which dependency etc. etc. so that we would know when to update the pin, then that would be awesome. I will try and have a look at the other PRs soon.

pwt:
> @pwt @jjlawren I think we should probably develop make a decision on what we want to with support of older version in general, so that we have that to lean on the next time this problem pops up. From my point of view, the answer is somewhere in between the things you have already discussed.
> 
> As @pwt, I would prefer the measuring stick for which version we do support to be the oldest python versions that LST distributions support _in as far as that can be achieved with reasonable effort_. So, personally I would be willing to compromise on it, the the burden is too high, but I too have heard of clever version pinning tools that apply to the dependencies dependencies, so if we could set one of those up **and** note in internal documentation, which distributions old python is limiting which dependency etc. etc. so that we would know when to update the pin, then that would be awesome. I will try and have a look at the other PRs soon.

My (personal) position on this is that we might as well support older releases unless there's a compelling reason not to do so -- e.g., libraries on which we depend dropping support in some significant way, or language features that it's highly inconvenient to live without. It has not been onerous to maintain 3.5 support, and I'm sure it's been a service to plenty of legacy systems out there.

Having said that, once 3.5 stops receiving **any** security patches (June 2022), we should probably drop support to 'nudge' people into updating their systems.

I'm not really in favour of complex support matrices or elaborate version pinning. Not the best use of our energies.

jjlawren:
Python 3.5 reached EOL over a year ago. In fact, 3.6 reached EOL a few weeks ago in Dec 2021.

The difficulty with supporting old versions is going to be more around the dependencies than the distributions. Dependencies will drop support for python versions as they reach EOL, and we'll have little control over problems caused by outdated dependencies when those arise.

As noted above, both `requests` and `black` dropped support for 3.5 a while back (and `requests` plans on dropping 3.6 support in 2022). Having divergent dependency versions based on distribution may become a problem over time. For example, `requests` may change fix a bug or change behavior, or old versions of `black` may format PRs differently than what our CI uses.

We already have a similar issue today as versions are not pinned and unexpected dependency upgrades lead to CI failures for things unrelated to the PR.

pwt:
> Python 3.5 reached EOL over a year ago. In fact, 3.6 reached EOL a few weeks ago in Dec 2021.

I'm aware of this, but note that the Debian LTS project is supporting Stretch (and Python 3.5) through to the end of June 2022.

> The difficulty with supporting old versions is going to be more around the dependencies than the distributions. Dependencies will drop support for python versions as they reach EOL, and we'll have little control over problems caused by outdated dependencies when those arise.

Agree with the risk; currently hypothetical. Assess and deal with it when we encounter it?

> As noted above, both `requests` and `black` dropped support for 3.5 a while back (and `requests` plans on dropping 3.6 support in 2022). Having divergent dependency versions based on distribution may become a problem over time. For example, `requests` may change fix a bug or change behavior, or old versions of `black` may format PRs differently than what our CI uses.

Agreed; also currently hypothetical. (Black has never supported 3.5, FWIW.)

> We already have a similar issue today as versions are not pinned and unexpected dependency upgrades lead to CI failures for things unrelated to the PR.

Agreed, but this is not a version deprecation issue.

As I've alluded, I have in mind to deprecate 3.5 in June unless there's a significant cost we're incurring in the meantime.

jjlawren:
Agreed it's not causing any critical issues now and June is a reasonable goal. The biggest benefits I see from dropping 3.5 would be:
1. Allowing us to have a _clean_ set of pinned dependencies. Right now we'll have this mess while 3.5 is around: https://github.com/SoCo/SoCo/pull/857/files
2. Ability to use new Python features, like f-strings in 3.6.

jjlawren:
I just reviewed #857 to see what the current state of dependencies looks like. As of right now, `flake8`, `pylint`, `pytest`, and `requests` no longer support 3.5 and we're using outdated versions of those on the CI for the 3.5 tests.

jjlawren:
If we're aligning with Debian LTS support timelines, then we should be able to remove both 3.5 and 3.6 support in when Debian 9 ("Stretch") LTS support ends on June 30, 2022. The next LTS release will be based on Debian 10 ("Buster") which comes packaged with 3.7.

pwt:
I'm in agreement with dropping 3.5 support for the upcoming v0.28 release, but also dropping 3.6 requires further discussion, at least from my point of view.

jjlawren:
The latest release of `requests` has officially dropped support for py36: https://github.com/psf/requests/releases/tag/v2.28.0

pwt:
> The latest release of `requests` has officially dropped support for py36: https://github.com/psf/requests/releases/tag/v2.28.0

Meaning that folks using 3.6 will not get `requests` updates. Everyone else will. SoCo will keep working, at least for the time being.

I will shortly release v0.28. This PR will be included if it is restricted to removing Python 3.5 support. Otherwise, it won't be included.

jjlawren:
I'll revert the 3.6 removal for now.

jjlawren:
CI failures are from a new unpinned `pylint` version and are addressed here: https://github.com/SoCo/SoCo/pull/916.

pwt:
> I'll revert the 3.6 removal for now.

Thanks for doing that. Let's discuss 3.6 removal for the a future release.

My general philosophy, FWIW, is that we shouldn't unnecessarily inconvenience SoCo users unless we have solid engineering reasons to do so.

PR #856, opened 2021-08-13
Use Github Action to upload to PyPI

jjlawren:
This is a Github Actions workflow to automatically publish tagged releases to PyPI as described [here](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries).

This will require a repository owner to:

1. Create a new token on PyPI under "Account Settings > API tokens"
2. Set it as `PYPI_TOKEN` in the repository's encrypted secrets

More details about the encrypted secrets here: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,0 +1,36 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: PyPI
+
+on:
+  release:
+    types: [published]
+
+env:
+  DEFAULT_PYTHON: 3.9
+
+jobs:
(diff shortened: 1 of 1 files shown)


KennethNielsen:
@pwt I remember we discussed it at some point and as I recall back then you seemed to prefer to keep it a manual process. As the one that does all the releases, I will the decision on this one completely up to you.

pwt:
> @pwt I remember we discussed it at some point and as I recall back then you seemed to prefer to keep it a manual process. As the one that does all the releases, I will the decision on this one completely up to you.

Building/pushing the release is the least onerous step in the overall release process (I execute one small local script, and I'm done.)

That's not to say that I don't appreciate the value in this automation enhancement, and I'd like to keep it open as a future possibility.

PR #857, opened 2021-08-13
Use venv cache between CI runs

jjlawren:
This will use the `actions/cache` action to store and load a venv cache between runs using the same dependencies. In order to do this, each entry in `requirements-dev.txt` must be explicitly pinned as a hash of that file is used to determine if a cache should be reused.

Since the dependencies must be pinned, I've also added a Dependabot config to automatically keep those libraries up to date.

Both `twine` and `wheel` have been removed as hard dependencies. An alternative to using `twine` to upload from a local dev environment is to use Github Actions to do this for us. The PR for that feature is here: #856.

~~This may be dependent on dropping Python 3.5 support as several of the dependencies have already dropped support. This is proposed in #855.~~

This does _not_ affect the standard `requirements.txt` file and no pinning is necessary there. This will only affect development environments & CI runs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "11:00"
+    open-pull-requests-limit: 10
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
+      time: "11:00"
+    open-pull-requests-limit: 10
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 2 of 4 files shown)


jjlawren:
Figured out how to make this compatible with 3.5, but it makes the `requirements-dev.txt` file a little ugly. That can be cleaned up when support is actually dropped.

`black` and `sphinx` dependencies are not installed in 3.5 environments as the CI currently doesn't use them.

jjlawren:
Rebased on #855 and removed the 3.5-specific pins. Unfortunately since this PR was opened, several of the dependencies have dropped support for 3.6 and new version-specific pins are now required.

jjlawren:
After working through a lot of version-specific linting issues, I think this is in a good state again. There are still more version-specific pins than I'd like in `requirements-dev.txt`, but it will be that way until we drop (official) support for 3.6 (and 3.7!).

@pwt is this PR something you can consider? Also, are you open to dropping 3.6 (at least in the CI) as it's been [EOL](https://devguide.python.org/versions/) since Dec 2021? It would clean up several of the version pins.

pwt:
> @pwt is this PR something you can consider? Also, are you open to dropping 3.6 (at least in the CI) as it's been [EOL](https://devguide.python.org/versions/) since Dec 2021? It would clean up several of the version pins.

Thanks for this PR, @jjlawren. I'm sorry it has languished -- I'll get round to looking at it as soon as I can. I just want to make sure I understand it, because I've often had to tweak the check jobs to deal with emergent issues.

PR #858, opened 2021-08-24
Add Deezer support to the sharelink plugin

tipofthesowrd:
I added a DeezerShare link interpreter for the plugin.
Seems to work for albums / playlists / tracks on my end. 

The sharelink is similar to a Spotify sharelink:

deezer:playlist:5390258182

### DIFF
--- a/soco/plugins/sharelink.py
+++ b/soco/plugins/sharelink.py
@@ -109,6 +109,24 @@ class TIDALShare(ShareClass):
         encoded_uri = tidal_uri.replace("tidal:", "").replace(":", "%2f")
         return (share_type, encoded_uri)
 
+class DeezerShare(ShareClass):
+    """Deezer share class."""
+
+    def canonical_uri(self, uri):
+        match = re.search(r"https://deezer.*[:/](album|track|playlist)[:/]([\w-]+)", uri)
+        if match:
+            return "deezer:" + match.group(1) + ":" + match.group(2)
+
+        return None
+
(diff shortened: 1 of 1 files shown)


amelchio:
Awesome, I am happy to see this plugin grow.

Can you clarify, did you invent the `deezer:playlist:5390258182` format? Spotify has its own `spotify:` scheme which is available to share from their app but they are moving away from that and maybe we should stop the proliferation within the plugin at this point.

The intention is to use official share links like `https://www.deezer.com/en/playlist/5390258182` seems to be (though I don't know the Deezer service, only browsed the site just now). This way copy-paste works and one does not have to carefully construct the URI by hand.

tipofthesowrd:
Correct, I set the format to be identical to the Spotify format myself.

In an earlier commit I was using the URL as you described. These work, they
show up when browsing the website (Format:
https://www.deezer.com/playlist/5390258182
<https://www.deezer.com/en/playlist/5390258182>)

However fyi the actual share links from the app generated by Deezer are
being shortened:
https://deezer.page.link/w9zgJuBES9iezNT9A


Op wo 25 aug. 2021 08:32 schreef Anders Melchiorsen <
***@***.***>:

pwt:
If you run `black` on `sharelink.py`, and recommit, that should sort out the check job failures.

pwt:
Looks great! I can't test this myself, so we'll proceed optimistically and assume that your testing generalises.

PR #860, opened 2021-10-05
Migrate check job from Python 3.10-dev to 3.10 (released)

pwt:


pwt:
Note that specifying `3.10` in the matrix does not work. It erroneously looks for version `3.1`.

Version `3.11-dev` is not yet available in the [Python versions manifest](https://github.com/actions/python-versions/blob/main/versions-manifest.json).

PR #862, opened 2021-10-10
Add methods to add/separate satellite (surround) speakers

rytilahti:
This implements methods to allow combining a set of speakers to a surround setup, as well as disbanding them.
This could be useful for some use cases where you want to control the rear speakers individually.
However, as disbanding (or separating) a setup loses trueplay tuning, this turned out not to be as useful for my setup as I thought it would be.

Considering that I dug into how this works and wrote the code already, I'm publishing this to save time and effort if someone else is looking into this.


pwt:
Hi @rytilahti: Thanks for this contribution.

I'm not able to test this, but I think it's useful, missing functionality, so I've tidied it up, added tests, and I'll release it for user feedback. I'm sure folks will let us know if it doesn't work!

rytilahti:
Hey @pwt, happy to see this getting merged! My sole intention for this PR was to make this experiment public to save time for anyone who is looking into this, thanks for cleaning it up!

Issue #864, opened 2021-10-15
Multicast discovery not working when wireguard device present


towo:
Cheers, ran into a rather weird issue just now. Starting `mkchromecast` as normal (both versions in AUR, doesn't matter):

```
Mkchromecast v0.3.9
Creating Pulseaudio Sink...
Open Pavucontrol and Select the Mkchromecast Sink.
Starting Local Streaming Server
[Done]
Selected backend: parec
Selected audio codec: mp3
Default bitrate used: 192k
Default sample rate used: 44100Hz.
PID of main process: 284407
PID of streaming process: 284414
 * Serving Flask app 'mkchromecast.audio' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
 * Running on http://192.168.178.11:5000/ (Press CTRL+C to quit)
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/soco/discovery.py", line 212, in any_soco
    d for d in cls._instances[cls._class_group].values() if d.is_visible
KeyError: 'SoCo'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/mkchromecast", line 311, in <module>
    mk()
  File "/usr/bin/mkchromecast", line 59, in __init__
    self.audio_linux()
  File "/usr/bin/mkchromecast", line 78, in audio_linux
    self.cc.initialize_cast()
  File "/usr/share/mkchromecast/mkchromecast/cast.py", line 118, in initialize_cast
    zone = soco.discovery.any_soco()
  File "/usr/lib/python3.9/site-packages/soco/discovery.py", line 215, in any_soco
    devices = discover(allow_network_scan=allow_network_scan, **network_scan_kwargs)
  File "/usr/lib/python3.9/site-packages/soco/discovery.py", line 124, in discover
    _sock.sendto(really_utf8(PLAYER_SEARCH), (MCAST_GRP, MCAST_PORT))
OSError: [Errno 126] Required key not available
```

So I'm guessing it's an issue with my network interface, as it's sitting on a bond interface and probably has some deeper issues trying to do multicast there. Since other apps (e.g. `noson`) appear to have no problem finding the SONOS device, I'm guessing it's something to either be fixed in soco… or Python. (welp.)

FWIW, the networki device for `192.168.178.0/24` is just a simple bond device with

```
[NetDev]
Name=xxxx
Kind=bond

[...]

PR #866, opened 2021-10-19
Improvements to 'discover()'

pwt:
- Close sockets opened for multicast discovery
- Fix logic for 'allow_network_scan' fallback
- Additional DEBUG logging


pwt:
In fixing #864, I noticed that the sockets created for multicast discovery are not closed, potentially resulting in file descriptor leakage. I also noticed that the logic for fallback to network scan discovery was wonky.

This PR fixes these issues.

PR #868, opened 2021-11-16
Adjust timeouts for services and async event handler

jjlawren:
This adjusts the `requests` and `aiohttp` timeouts in two places:

1. Reduce the default service call timeout to 5s from 20s, allow passing in a timeout per-call.
2. Set an explicit async event handler timeout instead of using the default of 300s. (https://docs.aiohttp.org/en/stable/client_quickstart.html#aiohttp-client-timeouts)

### DIFF
--- a/soco/events_asyncio.py
+++ b/soco/events_asyncio.py
@@ -68,7 +68,7 @@ import time
 import asyncio
 
 try:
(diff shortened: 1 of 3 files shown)


pwt:
LGTM -- thanks.

PR #871, opened 2021-12-05
Support reporting of additional audio input formats

jjlawren:
Add more audio input formats reported by users. Let's keep this open for a bit to collect feedback.

Thanks @zacwest to get things started.


jjlawren:
As the information required to troubleshoot this is contained in the value of `soundbar_audio_input_format`, I've also reduced the log severity from warning to debug to make this less noisy.

jjlawren:
This value also feels like it should be a bitmask, but I haven't found a pattern to decode it quite yet.

jjlawren:
I'm starting to wonder if these bitmasks contain information besides just the audio codec, like the transport (HDMI/optical), speaker layout, etc.

rytilahti:
If it would contain information about the layout, I'd expect that the values would vary among users with and without sub, no? But it definitely has some sort of pattern, I think it'd be great to get some more input (for atmos, dts) to see if that simplifies the puzzle :)

pwt:
LGTM -- thanks. Curious to understand the model Sonos is using to determine the codes.

terriblyvile:
Unknown audio input format: 84934716 showing as Dolby TrueHD 5.1 in the Sonos app.

pwt:
> 84934716 showing as Dolby TrueHD 5.1 in the Sonos app.

Added in SoCo v0.30.7 (https://github.com/SoCo/SoCo/commit/64be77d2ae9c4dda0d3bf09342ebe585b10deb38).

PR #872, opened 2021-12-05
Address new pylint issue

jjlawren:
Addresses an unrelated CI issue uncovered in https://github.com/SoCo/SoCo/pull/871.

Since `pylint` is not pinned to a specific version, a newer release has marked previously accepted code with a warning. CI pinning is included with https://github.com/SoCo/SoCo/pull/857 (and improved with https://github.com/SoCo/SoCo/pull/855) which would avoid shifting test requirements in the future.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -526,7 +526,7 @@ class SoCo(_SocoSingletonBase):
     def play_mode(self, playmode):
         """Set the speaker's mode."""
         playmode = playmode.upper()
-        if playmode not in PLAY_MODES.keys():
+        if playmode not in PLAY_MODES:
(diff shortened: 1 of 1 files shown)


pwt:
Thanks for fixing this.

> Since `pylint` is not pinned to a specific version, a newer release has marked previously accepted code with a warning. CI pinning is included with #857 (and improved with #855) which would avoid shifting test requirements in the future.

I'll need to take the time to understand what #857 is doing, unless someone more knowledgeable than me is prepared to take a look. (I haven't been treating it as urgent since CI throughput/efficiency isn't obviously a problem for SoCo.)

PR #874, opened 2021-12-12
Add `dialog_level` convenience property to mirror `dialog_mode`

jjlawren:
~~Renames the `SoCo.dialog_mode` property to `dialog_level` to match both the service call and subscription event callback payload. Adds a compatibility shim to avoid breaking existing usage.~~

Adjusted to simply add a convenience property named `dialog_level` which wraps `dialog_mode`.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -205,7 +205,7 @@ class SoCo(_SocoSingletonBase):
         loudness
         balance
         night_mode
-        dialog_mode
+        dialog_level
         supports_fixed_volume
         fixed_volume
         soundbar_audio_input_format
@@ -1143,7 +1143,17 @@ class SoCo(_SocoSingletonBase):
 
     @property
     def dialog_mode(self):
-        """bool: The speaker's dialog mode.
+        """Compatability wrapper for dialog_level getter."""
+        return self.dialog_level
+
+    @dialog_mode.setter
+    def dialog_mode(self, dialog_mode):
+        """Compatability wrapper for dialog_level setter."""
+        self.dialog_level = dialog_mode
+
+    @property
+    def dialog_level(self):
+        """bool: The speaker's dialog level.
 
         True if on, False if off, None if not supported.
         """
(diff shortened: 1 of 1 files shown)


pwt:
While it's inconsistent with the name of the service call, it seems to me that `dialog_mode` is a more accurate description of the property than `dialog_level`?

jjlawren:
The XML description and subscription event payloads received from the speakers use `DialogLevel` and `dialog_level` respectively. Right now, mapping the `dialog_level` event payload key to the associated `SoCo.dialog_mode` property/attribute requires a special mapping which is not needed for `night_mode`, `cross_fade`, `bass`, `treble`, `sub_enabled`, etc.

pwt:
> The XML description and subscription event payloads received from the speakers use `DialogLevel` and `dialog_level` respectively. Right now, mapping the `dialog_level` event payload key to the associated `SoCo.dialog_mode` property/attribute requires a special mapping which is not needed for `night_mode`, `cross_fade`, `bass`, `treble`, `sub_enabled`, etc.

So whoever implemented this therefore made a conscious choice to remap `level` to `mode`, presumably because it's a discrete on/off mode, not a continuous level.

jjlawren:
Logically "mode" makes more sense, but "level" is cleaner to work with from an API standpoint. That's why I wanted to keep a compatibility shim in place.

pwt:
> Logically "mode" makes more sense, but "level" is cleaner to work with from an API standpoint. That's why I wanted to keep a compatibility shim in place.

Yes, I understand that backward compatibility is retained. It's just that the property rename seems like a regression to me, at least from a UX perspective. I'll seek another opinion.

KennethNielsen:
Hi @jjlawren and @pwt

As requested, I'll offer an opinion and I can certainly see both points to this argument.

I tend to agree with @pwt. The reason for this is that I view SoCo as a more of a "driver", than an API compatibility mapper in Python, meaning that SoCo should map the Sonos API to Python concepts and values. To me this also means mapping concepts in the API to have good understandable names, which to me is what the properties had to begin with.

If anything, to fix the point that @jjlawren makes about matching properties with events, I would suggest to reverse the fix and rename the incoming events to match the property (and then of course there keep a back compatibility fix by in practice copying the event and rename the copy, more than just rename).

Regards Kenneth

jjlawren:
I don't think the event payloads should be modified as the current convention today is to (mostly) pass those through as-is. It would also break existing usage of events.

How about leaving the `dialog_mode` property is it was, and adding a wrapper on top to better fit the Sonos API conventions?

pwt:
Thanks for the discussion on this PR. This approach LGTM.

PR #876, opened 2021-12-21
Revert service call timeout reduction

jjlawren:
I've received several reports from users that previously working service calls are now timing out. This coincides with the reduction of timeout from 20s to 5s in #868. This PR retains the ability to set a timeout per-call but reverts the default timeout to 20s.

### DIFF
--- a/soco/services.py
+++ b/soco/services.py
@@ -428,7 +428,7 @@ class Service:
         return (headers, body)
 
     def send_command(
-        self, action, args=None, cache=None, cache_timeout=None, timeout=5, **kwargs
+        self, action, args=None, cache=None, cache_timeout=None, timeout=20, **kwargs
     ):  # pylint: disable=too-many-arguments
         """Send a command to a Sonos device.
 
--- a/tests/test_services.py
+++ b/tests/test_services.py
@@ -243,7 +243,7 @@ def test_send_command(service):
             "http://192.168.1.101:1400/Service/Control",
             headers=mock.ANY,
             data=DUMMY_VALID_ACTION.encode("utf-8"),
-            timeout=5,
+            timeout=20,
         )
         # Now the cache should be primed, so try it again
         fake_post.reset_mock()


pwt:
> I've received several reports from users that previously working service calls are now timing out. This coincides with the reduction of timeout from 20s to 5s in #868. This PR retains the ability to set a timeout per-call but reverts the default timeout to 20s.

Thanks for responding to these reports. I'll merge this now and issue a patch release.

pwt:
SoCo v0.25.1 is now released, including this change.

relevitt:
Hi @pwt,

I was also having timing out errors with timeout=5. These have reduced with timeout=20, but they haven't been eliminated. In an application that adds a hundred spotify tracks to the queue (this has to be done in batches of 16, being the maximum Sonos permits), with timeout=5 I get an error after the first 16 tracks have been added and with timeout=20 I get an error after the first 48 have been added.

I've manually updated my services.py file to change this back to timeout=None. That fixes the problem.

Is there a reason not to have timeout=None? I haven't been keeping track of how this issue arose. Absent some problem that was being fixed by this change (timeout=5 or timeout=20), perhaps we could go back to how things were before?

If we must have a default timeout value (other than None), perhaps there could be an option to change it via a configuration variable in config.py, like config.TIMEOUT = 20. Then I could stop this problem from recurring in my app without having to make manual changes to services.py.

Hope that makes sense.

R

pwt:
Hi @relevitt:

@jjlawren may wish to comment on this, but the intent of introducing a timeout is to prevent the possibility of very long execution delays in the event of network duress. I think this is reasonable, and the 20s default probably works for the majority of use cases. We'll keep this under review, obviously, in case we see evidence of an increasing number of problem reports.

I'm interested, however, in the idea of setting the timeout via a config variable.

jjlawren:
A new timeout was not introduced, it was changed from a default of 20s -> 5s. After some users started hitting this lower threshold it was reverted back to the original value of 20s in this PR.

You _can_ now pass in a `timeout` kwarg to raw service calls to override the default, which was not possible before. For example:
```python
soco.avTransport.AddURIToQueue(
    [
        ("InstanceID", 0),
        ("EnqueuedURI", queueable_item.resources[0].uri),
        ("EnqueuedURIMetaData", metadata),
        ("DesiredFirstTrackNumberEnqueued", position),
        ("EnqueueAsNext", int(as_next)),
    ],
    timeout=None,
)
```
However this won't help with convenience methods such as `SoCo.add_to_queue()` which don't expose that parameter today.

pwt:
Thanks for the clarification, @jjlawren. So, we've just restored the original behaviour (20s timeout), while allowing some additional flexibility to adjust the timeout.

What do you think of the idea of being able to override the timeout universally, using a `config.py` setting?

relevitt:
Hi @jjlawren and @pwt,

Thank you for looking into this. A timeout must have been added within the last couple of years, because I was using a rather old version of soco (on my system the files are dated 20 Feb 2020) and there was no timeout then (or rather, timeout was set explicitly as None). Someone must have added a timeout since then, but I haven't checked how or why.

I updated to the  version of soco with timeout=5 only recently, and then when I saw it had introduced this error into my app, I updated again. However, timeout=20 still introduces errors (as I described in my previous post).

I can't say I've ever experienced a problem arising from timeout=None. However, if there has to be a different timeout, surely it makes sense for users to be able to tweak it to suit their own needs through a config variable? I would still then wonder if the default option of timeout=None might not be better, unless there was a good reason for changing it.

R

jjlawren:
Perhaps the speakers are making synchronous calls to Spotify for each track added to the queue which takes a long time. I haven't done any investigations myself so this is just a guess based on nothing specific.

The timeout for service calls was introduced about a year ago in #814. Setting a timeout to `None` has the chance of hanging execution indefinitely and is not a good default.

Fixing a problem by increasing a timeout is likely covering up a different issue. The current 20 second timeout is a _very_ long time and exposing a global config timeout for users to tweak on their own feels like a band-aid approach.

In the interest of providing a short-term workaround, I'm not quite sure what I'd propose which would avoid leading to issues in the future. For example, a global config option could easily be forgotten and lead to confusing behavior when diagnosing these types of issues in the future.

@relevitt have you encountered these timeouts with any calls besides `add_multiple_to_queue`? Perhaps we could adjust the timeout dynamically based on the number of items passed into that call. Or (way easier) just increase the timeout for that specific and potentially heavy call to something like 30-60s.

pwt:
PR #880 implements a possible approach to a configurable timeout. For most users, there's no change, and the default timeout will remain at 20.0s (now set in `config.py`), but it's easy to change this dynamically or to override it using the `timeout` kwarg.

relevitt:
@pwt thank you very much for PR #880. I've tested it and it works for me. Problem solved.

@jjlawren the bit of code I use to load multiple tracks is:

```
self.device.avTransport.AddMultipleURIsToQueue([
                    ('InstanceID', 0),
                        # Don't provide UpdateID if tracks are being added to end of
                        # the queue, as the add should be able to succeed regardless
                        ('UpdateID', self.status.updateid if var['pos'] else 0),
                        ('NumberOfURIs', len(chunk)),
                        ('EnqueuedURIs', uris),
                        ('EnqueuedURIsMetaData', meta),
                        ('ContainerURI', ''),
                        ('ContainerMetaData', ''),
                        ('DesiredFirstTrackNumberEnqueued', var['pos']),
                        ('EnqueueAsNext', 0 if var['pos'] else 1)
                ])
```
I don't currently provide any *args or **kwargs, but I assume I could.

As to the philosophical point you raise, I'm not sure what to say. Twenty seconds is very long, but evidently not long enough. Adding Spotify files to a Sonos speaker with soco is a bit hit and miss. Some of us seem to have found a way to do this, but I don't think it's the way a native Sonos app would do it.

pwt:
> @pwt thank you very much for PR #880. I've tested it and it works for me. Problem solved.

Thanks for testing. This will be included in SoCo v0.26.0, which I expect to release once we're happy with the new music service capabilities (#883) ... hopefully sometime next week.

PR #878, opened 2021-12-30
Add audio_delay property for TV dialog sync.

jjlawren:
Adds controls to the TV Dialog Sync audio delay exposed for home theater devices.

image

Oddly, this value doesn't seem to update in the native Sonos app even when navigating away and back. However its value is correctly reflected in realtime from `renderingControl` subscription events as `audio_delay`.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -223,6 +223,7 @@ class SoCo(_SocoSingletonBase):
         treble
         loudness
         balance
+        audio_delay
         night_mode
         dialog_mode
         supports_fixed_volume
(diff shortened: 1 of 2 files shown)


pwt:
Do we know what units are used for the adjustment, and if so could that be included in the documentation?

Tests would be good, if possible.

jjlawren:
Doesn't seem like it's expressed in units, just a 0 to +5 range. Here is where Sonos describes it: https://support.sonos.com/s/article/3336. That page is linked from [here](https://www.sonos.com/support/en-us/sonos-user-guide/index.html#rhtocid=_0_4_1&t=sonos-user-guide%2Fhome-theater%2Fhome-theater.htm).

jjlawren:
A side note, I noticed that individual tests within the `TestRenderingControl` class are written in a way where they're affected by the end state of previous tests and therefore cannot be run individually (e.g., with the `-k` flag). For example, that's why I had to re-set the `_is_soundbar` attribute at the start of the test to pass the full suite.

pwt:
Thanks for the comments and for adding the tests. LGTM.

PR #879, opened 2022-01-03
Add `mic_enabled` property to the SoCo class

jjlawren:
Adds a new read-only boolean property which reports the current state of the device's microphone. For devices without a microphone or where a voice service has not yet been configured, the return value will be `None`.

Modified the `ZoneGroupState` XML in the tests to mark two speakers as voice-enabled in order to allow testing this new feature. Should have no effect on existing tests.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -233,6 +233,7 @@ class SoCo(_SocoSingletonBase):
         trueplay
         status_light
         buttons_enabled
+        mic_enabled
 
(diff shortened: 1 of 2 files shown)


pwt:
LGTM, works on my system. Thanks.

pwt:
Just to note that I've observed a situation where the `mic_enabled` state is incorrectly reported as `False`, using a Move speaker. Toggling the microphone state resolved it.

PR #881, opened 2022-01-04
Refactor ZGS parsing; add `voice_service_configured` property

jjlawren:
Slight refactor to minimize the number of statements in `_parse_zone_group_state`. This may allow removal of the `pylint: disable=too-many-statements` pragma introduced in #879.


jjlawren:
Refactored the changes from #879 to fit the proposed style in this PR. Also added a new `voice_service_configured` boolean property since it's readily available.

Perhaps the raw `VoiceConfigState` value represents the specific voice service. That can be added later.

PR #869, opened 2021-11-26
Bring music services back based on #763 PR

goldyfruit:
This PR is 99% based on #763 from @KennethNielsen 

- The tests have been fixed *(from a `pytest` and `pylint` perspective)*
- Returns `linkDeviceId` if the music service requires it
- Tested on Spotify _(AppLink)_, Deezer _(DeviceLink without `linkDeviceId`)_, Amazon Music _(DeviceLink)_
- Rename `device_or_app_link_auth_part1()` to `get_device_link_code()`
- Rename `device_or_app_link_auth_part2()` to `get_device_auth_token()`

I'm currently using this PR with the [Mycroft AI Sonos Controller skill](https://github.com/smartgic/mycroft-sonos-controller-skill).

PR #763, opened 2020-11-29
[WIP] Bring music services back

KennethNielsen:
This PR contains the WIP for bringing music services back. I've been working of different sources for this and will of course credit them before merging.

**NEEDS TESTING**

The short history of music service authentication with Sonos is that Sonos used to store and make available _account_ information (which covered the main access to a music service), which meant that the only thing we had to do was to query the speaker for all the information we needed to talk to the music services. Then at some point, some guess due to pressure from the music services, this was changed. One of the mechanisms which replaced this was different forms of browser authenticated token access, which involves authenticating in the browser, to get an access token. Two of these methods are called DeviceLink and AppLink and it is these that this PR aims to implement.

The code has only so far been tested successfully on one music service, Tidal, which uses DeviceLink, so I am in need of lots more testing and especially on an AppLink music service, such as Spotify.

You can test the new auth features with code like this:
```python
from soco.music_services import MusicService
music_service = MusicService("Spotify")
reg_url, link_code = music_service.device_or_app_link_auth_part1()
print(reg_url)
print(link_code)
```

Visit the url and authenticate and store the link_code for later. Then, after having authenticated, call part 2, with the link code as only argument
```python
music_service.device_or_app_link_auth_part2("THE_LINK_CODE_FROM_BEFORE")
```
and then, if everything goes well, we should be authenticated. **It is very important to not call part1 again before when calling part 2, since then it will be a new link code, a wrong one. The link code passed into part2 has to be the one returned along with the reg URL actually used**. The tokens by default are stored in a json file in the users default config dir so e.g. on Linux you can see the stored token with the command:
```bash
cat ~/.config/SoCo/token_store.json
```

Now the music service should be available for use like before, where items can be browsed and searched for and the results enqueued. You can use code along the lines of this to quickly test it (only editing the name of the speaker away from "Kontor" and specify the name of the music service at the bottom:

```python
from time import sleep
from random import choice
import soco
from soco.music_services import MusicService
from soco.discovery import by_name

def confirm_work(service):
    t = MusicService(service)
    z = by_name("Kontor")

    # Print the search categories. TIDAL e.g. has 4 search categories:
    # ['artists', 'albums', 'tracks', 'playlists']
    print(t.available_search_categories)

    for search_category in t.available_search_categories:
        z.clear_queue()
        sleep(1)
        # Clear queue
        print("Search category:", search_category)
        results = t.search(search_category, "Metallica")
        print(len(results))
        try:
            z.add_to_queue(choice(results))
[...]

PR #883, opened 2022-01-16
Reintroduce support for music services (experimental)

KennethNielsen:
I finally got around to finishing this up. I apologize for taking so long. It was a bit of a beast for me at first and it has been difficult to break up, but I probably should have tried that harder, so as to get something merged. Or at the very least let people know that I would not be able to finish it any time soon, so someone else could have finished it. But I kept thinking I would find the time and I really wanted to finish it up :(

This PR supersedes the draft PR #763 and hopefully also #869, if @goldyfruit gives it his blessing. This PR:
* Includes everything that was discussed in #763 (including per household token store brought up by  @amelchio in https://github.com/SoCo/SoCo/pull/763#issuecomment-740441769)
* Has been updated with recent upstream master, so it should merge cleanly
* Includes most of the improvements added in #869
* Includes the fix for Navidrome + bonob mentioned here https://github.com/SoCo/SoCo/pull/763#issuecomment-992667867
* I also added some known problems with the code in the module docstring

I now consider this ready for review. When reviewing please keep in mind that it is known that there are more improvements that should be done, but that it is probably important to get it merged soon, so others can contribute to that.

If anyone can find the time, please also test it as described in #763 and then it's ready for review.


goldyfruit:
@KennethNielsen Thanks the PR and don't apologize!

I tested your branch with my Mycroft Sonos Controller skill on Spotify _(AppLink)_, Deezer _(DeviceLink without linkDeviceId)_, and Amazon Music _(DeviceLink)_. I was able to authenticate for `part1` and part2` for all of them!

The only comment I have would be to add a note in the code or the changelog about the commented code.

Good job @KennethNielsen

pwt:
> @KennethNielsen Thanks the PR and don't apologize!

Seconded!

> I tested your branch with my Mycroft Sonos Controller skill on Spotify _(AppLink)_, Deezer _(DeviceLink without linkDeviceId)_, and Amazon Music _(DeviceLink)_. I was able to authenticate for `part1` and `part2` for all of them!

Thanks so much for all the testing.

> The only comments I have would be:
> 
> * Add a note in the code or the changelog about the commented code

@KennethNielsen: I'm probably missing something obvious, but is there a reason for commenting out the old code rather than deleting it?

> * Rebase with 0.25.3 release

No need to do this. (The 0.25.fixes branch, from which v0.25.3 is released, diverges from master because it generally only consists of defect fixes cherry-picked from master, and omits other changes.)

> Good job @KennethNielsen

Indeed!

goldyfruit:
> 
> No need to do this. (The 0.25.fixes branch, from which v0.25.3 is released, diverges from master because it generally only consists of defect fixes cherry-picked from master, and omits other changes.)
> 
Thanks for the information.

KennethNielsen:
@goldyfruit thanks for all the extra testing.
@pwt no, you are not missing anything with the commented code. For so long I was struggling to make it do anything, so I kept the old code around in case I needed some of it. But it is silly, we can just go back in time with git if we ever need it. I will either delete it or add a comment to explain why it is there but commented out.

KennethNielsen:
I remove the old commented out code. You should  know, that removing the commented code removes an old idea about the queue items being upgraded to proper music service items whenever possible, so that they could be re-added to the queue and browser etc. However, that was no longer possible, since we had no music services, and AFAICT it is no longer possible with the new music servies, so I simply dispensed with it. In any case:
* It was something I implemented long ago
* I don't think it was ever terrible useful and so I don't think anyone ever really used it
* It was broken anyway, due to the no music services
... so I don't feel that badly about removing it. But now you are informed.

Note also, that with the removal of this silly queue item upgrade feature, all reason for the existence of the music_service_entry module has disappeared (I think)! I will see about removing that in another PR.

goldyfruit:
So, I failed my tests, sorry about that...

During some testing with @jjlawren about `Plex`, I retried the authentication with the `Amazon Music` service and I got the same error as here https://github.com/SoCo/SoCo/pull/763#issuecomment-742775798 with the same diagnostic as here https://github.com/SoCo/SoCo/pull/763#issuecomment-975933681

In my PR I was able to solve this by adding `link_device_id` as argument to `part2`.

https://github.com/SoCo/SoCo/pull/869/files#diff-79b6eaf47af5904eadf59b9a8176c268b4612006b2301ee3fdec13cd52c8c597R264-R285

jjlawren:
After some testing, I made some suggestions in [this branch](https://github.com/KennethNielsen/SoCo/compare/bring_music_service_back...jjlawren:music_service_tweaks):
* Add `linkDeviceId` handling into AppLink flows.
* Simplify the `part1`/`part2` handling by caching the code/ID for the user instead of passing back tuples. However, still allow them to override these values by accepting optional arguments in `part2`. For example, to continue the flow elsewhere.
* Uncomment the timezone handling.
* Reduce unnecessary log levels, fix typos.

This will need more testing, but it worked for the few services I had available.

Here's how auth can be done now:
```
>>> from soco.music_services.music_service import MusicService
>>> plex = MusicService("Plex")
>>> reg_url = plex.device_or_app_link_auth_part1()

 -> Go complete authentication

>>> plex.device_or_app_link_auth_part2()
>>> for result in plex.search("artists", "Stevie Wonder"):
...     result
...
<soco.music_services.data_structures.MSArtist object at 0x7f8b5b1bf3d0>
```

KennethNielsen:
@jjlawren @goldyfruit thanks for the testing.

@jjlawren about your tweaks, how do you want to attack this? If your commit is on top of my branch, should I pull them in, or do you want to do a separate PR for it?

jjlawren:
Either way, as long as it's tested well. Since it would be a breaking change it should be considered before next release.

goldyfruit:
> Either way, as long as it's tested well. Since it would be a breaking change it should be considered before next release.

I can't get the `link_code` or the link_device_id` anymore.

```
  File "/tmp/plex.py", line 3, in <module>
    reg_url, link_code, link_device_id = music_service.device_or_app_link_auth_part1()
ValueError: too many values to unpack (expected 3)
```

jjlawren:
Correct, `part1` only returns the registration URL now. Check my example in the comment above.

goldyfruit:
> Correct, `part1` only returns the registration URL now. Check my example in the comment above.

Yes, I tested with `reg_url` but I need to retrieve the `link_device_id` to pass it in `part2`.

jjlawren:
Are you instantiating a new `MusicService` instance between steps? That will lose the cached `link_code` and `link_device_id` values.

goldyfruit:
> Are you instantiating a new `MusicService` instance between steps? That will lose the cached `link_code` and `link_device_id` values.

Yes I'am, because the process is not fully automated because of the web authorization.

1. `part1`
2. Web authorization
3. `part2`

Maybe the `link_device_id` could be stored in a temporary file during `part1` and read then delete during `part2`.

jjlawren:
I just tested linking Amazon Music successfully. If you need to restart the flow with a new `MusicService` instance, you can do this:
```
>>> from soco.music_services import MusicService
>>> az = MusicService("Amazon Music")
>>> reg_url = az.device_or_app_link_auth_part1()
>>> print(az.link_code, az.link_device_id)  # Save these for later

 -> Complete auth, restart flow

>>> az = MusicService("Amazon Music")
>>> az.device_or_app_link_auth_part2(link_code=link_code, link_device_id=link_device_id)
```

goldyfruit:
> I just tested linking Amazon Music successfully. If you need to restart the flow with a new `MusicService` instance, you can do this:
> 
> ```
> >>> from soco.music_services import MusicService
> >>> az = MusicService("Amazon Music")
> >>> reg_url = az.device_or_app_link_auth_part1()
> >>> print(az.link_code, az.link_device_id)  # Save these for later
> 
>  -> Complete auth, restart flow
> 
> >>> az = MusicService("Amazon Music")
> >>> az.device_or_app_link_auth_part2(link_code=link_code, link_device_id=link_device_id)
> ```
Yep, it should work but it is not really leveraging the cache then.
Using a cache file could avoid to have to deal with `link_code` and `link_device_id` during `part2`.

jjlawren:
> Yes I'am, because the process is not fully automated because of the web authorization.

A note that this process is easier to test if you're running in an interactive shell as you won't lose the state of the instance. In an application that leverages `soco`, the `MusicService` instance should not be discarded during the authorization flow and the cache should persist. A file cache is not necessary.

goldyfruit:
> > Yes I'am, because the process is not fully automated because of the web authorization.
> 
> A note that this process is easier to test if you're running in an interactive shell as you won't lose the state of the instance. In an application that leverages `soco`, the `MusicService` instance should not be discarded during the authorization flow and the cache should persist. A file cache is not necessary.

Yeah that will not be applicable for me, you could have a look how I does the authentication here: https://github.com/smartgic/mycroft-sonos-controller-skill/blob/21.2.1/utils.py#L32-L88

But as long as I'm able to retrieve the `link_code` and the `link_device_id`, I'm fine.

jjlawren:
Submitted https://github.com/KennethNielsen/SoCo/pull/1 which should merge cleanly into this branch if accepted.

KennethNielsen:
> > > Yes I'am, because the process is not fully automated because of the web authorization.
> > 
> > 
> > A note that this process is easier to test if you're running in an interactive shell as you won't lose the state of the instance. In an application that leverages `soco`, the `MusicService` instance should not be discarded during the authorization flow and the cache should persist. A file cache is not necessary.
> 
> Yeah that will not be applicable for me, you could have a look how I do the authentication here: https://github.com/smartgic/mycroft-sonos-controller-skill/blob/21.2.1/utils.py#L32-L88
> 
> But as long as I'm able to retrieve the `link_code` and the `link_device_id`, I'm fine.

I tend to agree with @jjlawren that the general assumption of the auth is that it is on the same instance. That being said, we can of course make improvements for your workflow. To make it even easier and future proof (if other things would even need to be cached and entered into part2, we could add a `cached_auth_info` property which returns a dict with keynames that match kwargs in part2, so it would always be possible to **expand it into it:
```python
reg_url = ms.part1()
auth_info = ms.auth_info
# save auth info in whatever way you choose and reg at url

# Break into new instance, load auth info
ms.part2(**auth_info)
```

KennethNielsen:
@pwt Fixes was added from @jjlawren's branch. I consider it ready for whatever the next step is.

pwt:
> @pwt Fixes was added from @jjlawren's branch. I consider it ready for whatever the next step is.

Thanks for pulling this in @KennethNielsen. @jjlawren, @goldyfruit: thanks for all the contributions and testing. 

**Once you three let me know you consider this ready for merge**, I'll do so. I don't feel appropriately qualified to make the readiness judgment myself on this one.

I'll quickly follow the merge by a v0.26.0 release which should give us a broader testing base.

KennethNielsen:
The original PR was reviewed and I reviewed the changes from @jj, so I consider it ready for merge.

goldyfruit:
Agree with @KennethNielsen, ready to merge.
Thanks guys.

soco/music_services/music_service.py
```diff
@@ -155,50 +196,138 @@ def call(self, method, args=None):
...
+            auth_parts = result["authorizeAccount"]["deviceLink"]
+            if "linkDeviceId" in auth_parts:
+                link_device_id = auth_parts["linkDeviceId"]
+            return auth_parts["regUrl"], auth_parts["linkCode"], link_device_id
+        raise MusicServiceAuthException(
+            "device_or_app_link_auth_part1 is not implemented "
+            "for auth type {}".format(self.music_service.auth_type)
+        )
+
+    def device_or_app_link_auth_part2(self, link_code, link_device_id=None):
```
jjlawren, line 309: This (and its equivalent in `MusicService`) should probably return a `bool` for success.

soco/music_services/music_service.py
```diff
@@ -155,50 +196,138 @@ def call(self, method, args=None):
...
         except SoapFault as exc:
+            if "Client.AuthTokenExpired" in exc.faultcode:
+                raise MusicServiceAuthException(
+                    "Authorization for {} expired or invalid: [{} / {} / {}]".format(
+                        self.music_service.service_name,
+                        exc.faultcode,
+                        exc.faultstring,
+                        exc.detail,
+                    )
+                ) from exc
```
jjlawren, line 207: This fault code can occur when the token store does not contain auth info for a service. Should the case of missing credentials be reported to the user?

soco/music_services/music_service.py
```diff
@@ -155,50 +196,138 @@ def call(self, method, args=None):
...
-                self.music_service.account.oa_device_id = auth_token
-                self.music_service.account.key = private_key
+
+                # Create new token pair and save it
+                token_pair = (auth_token, private_key)
+                self.token_store.save_token_pair(
+                    self.music_service.service_id,
+                    self._device.household_id,
+                    token_pair,
+                )
```
jjlawren, line 239: If the payload does not contain a token/key pair could the token store save empty values?

jjlawren (commented):
Some design things to consider before potential modifications will lead to breaking changes:

1. Does it make sense to instantiate the token store with the household ID and maintain separate token stores per household? That way the lookups could be single-argument values instead of tuples which are a bit unwieldy.
2. Does it make sense to rename `part1`/`part2` to reflect what's happening in the flow? Perhaps something like `begin_auth`/`get_auth_url` and then `complete_auth`?
3. It seems the proper way to browse a service is to use the `get_metadata()` method. This was not intuitive to me. Should this either be renamed or a `browse` method added which wraps this call?

soco/music_services/music_service.py
KennethNielsen: I don't think so. For Python I don't think it is very common or idiomatic to return success or failure. If it doesn't raise an exception it succeeded.

goldyfruit:
> 2. Does it make sense to rename `part1`/`part2` to reflect what's happening in the flow? Perhaps something like `begin_auth`/`get_auth_url` and then `complete_auth`?

Briefly discussed on the chat, here is the discussion between @KennethNielsen and I

image

KennethNielsen:
@jjlawren 

> Some design things to consider before potential modifications will lead to breaking changes:
> 
>     1. Does it make sense to instantiate the token store with the household ID and maintain separate token stores per household? That way the lookups could be single-argument values instead of tuples which are a bit unwieldy.

I know but this feature came at the direct request of @amelchio here: https://github.com/SoCo/SoCo/pull/763#issuecomment-740441769 and since this isn't a public api, but only used internally, I didn't see much problem with it..

> 
>     2. Does it make sense to rename `part1`/`part2` to reflect what's happening in the flow? Perhaps something like `begin_auth`/`get_auth_url` and then `complete_auth`?

Yeah, it might. @goldyfruit also wanted to rename but more in the direction of what they did, like "get_auth_token" which I didn't like too much. My main reason for that is that I really want the names to convey that it is a two part process, to make sure users don't misunderstand this, as that will probably cause issues. `begin_auth` and `complete_auth` also hints at this, so that could work as well. The question then is: There is more than DeviceLink and AppLink authentication supported by Sonos, so should the methods names include this information to make sure we don't get a name clash later, so something like `begin_device_or_app_link_auth` and ditto `complete`? I willing to adjust according to popular opinion on this, as long as we follow general Python naming guidelines.

> 
>     3. It seems the proper way to browse a service is to use the `get_metadata()` method. This was not intuitive to me. Should this either be renamed or a `browse` method added which wraps this call?

Yeah. I've been thinking something along the same lines. I think, since we can just add a method that wraps it, this could be a separate enhancement and be saving it for later, we have the option to think about whether it should do something more than just wrap. Since it does not need to be in this PR I would suggest to not do it here :D

jjlawren:
> There is more than DeviceLink and AppLink authentication supported by Sonos, so should the methods names include this information to make sure we don't get a name clash later

The only other auth methods I know of are either `Anonymous` which _should_ not need authentication, and `UserId` which is a legacy method where we have no working examples. The `DeviceLink` and `AppLink` methods use a _very_ similar two-step process, with no indication that `UserId` would be similar. I would suggest keeping it simple with `begin_auth` and `complete_auth` and add a separate flow for `UserId` if we figure that one out.

pwt:
I've had a chance to do some testing, with Spotify only. Tokens seem to expire after ~30 minutes (that's a guess, I haven't measured it), after which:

```
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/music_services/music_service.py", line 197, in call
    result_elt = message.call()
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/soap.py", line 324, in call
    raise SoapFault(faultcode, faultstring, faultdetail)
soco.soap.SoapFault: ns8eeLx:Client.AuthTokenExpired: authTokenExpired
```

What steps are required to recover from the `AuthTokenExpired` case? Does web-based authentication ("part 1") need to be performed again? Apologies if I'm missing something.

jjlawren:
@pwt is that the complete traceback, and are there any other log messages in context? Are you on the latest commit from this branch?

That exception _should_ be caught [here](https://github.com/SoCo/SoCo/blob/d16c44ce6a667bda19c63b9af974b275fc29aa03/soco/music_services/music_service.py#L198) and then refreshed. However, `AuthTokenExpired` is a different exception than `TokenRefreshRequired`. This might require an adjustment.

pwt:
> @pwt is that the complete traceback, and are there any other log messages in context? Are you on the latest commit from this branch?

I'm on the current state of this branch (https://github.com/SoCo/SoCo/pull/883/commits/d16c44ce6a667bda19c63b9af974b275fc29aa03). I don't have logging enabled, and this is the full SoCo stack trace.

> That exception _should_ be caught [here](https://github.com/SoCo/SoCo/blob/d16c44ce6a667bda19c63b9af974b275fc29aa03/soco/music_services/music_service.py#L198) and then refreshed. However, `AuthTokenExpired` is a different exception than `TokenRefreshRequired`. This might require an adjustment.

Yes, it's the `AuthTokenExpired` exception that's being hit: the exception is immediately raised back to the caller and there's no attempt to refresh the token.

pwt:
Here's the full stack trace, with the re-raised exception:
```
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/music_services/music_service.py", line 197, in call
    result_elt = message.call()
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/soap.py", line 324, in call
    raise SoapFault(faultcode, faultstring, faultdetail)
soco.soap.SoapFault: ns8eeLx:Client.AuthTokenExpired: authTokenExpired

The above exception was the direct cause of the following exception:

  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/spotify.py", line 79, in <module>
    test_loop(ms)
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/spotify.py", line 22, in test_loop
    result = ms.search("artists", term="U2", count=1)
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/music_services/music_service.py", line 876, in search
    response = self.soap_client.call(
  File "/Users/pwt/Dropbox/github-projects/SoCo-pwt/soco/music_services/music_service.py", line 200, in call
    raise MusicServiceAuthException(
soco.exceptions.MusicServiceAuthException: Authorization for Spotify expired or invalid: [ns8eeLx:Client.AuthTokenExpired / authTokenExpired / None]
```

pwt:
With a little bit of background testing, looks like the Spotify token expires after 50-55 minutes if used every 5 minutes, and after about 30 minutes if there's no regular activity.

Possibly consistent with the [Spotify documentation](https://developer.spotify.com/documentation/general/guides/authorization/code-flow/):

> Access tokens are deliberately set to expire after a short time, after which new tokens may be granted by supplying the refresh token originally obtained during the authorization code exchange.

jjlawren:
Ok, that makes more sense with the chained traceback. That exception apparently also needs a token refresh attempt.

I saw this exception when testing a service previously that had no credentials in the token store yet. Perhaps we should only attempt to refresh if tokens currently exist.

pwt:
I've just re-synced this branch with `master`, with a view to v0.26.0 release readiness, and to address a conflict with a recent merge. Sorry for the raft of changes.

KennethNielsen:
@jjlawren 

> Some design things to consider before potential modifications will lead to breaking changes:
> 
>     1. Does it make sense to instantiate the token store with the household ID and maintain separate token stores per household? That way the lookups could be single-argument values instead of tuples which are a bit unwieldy.

I know but this feature came at the direct request of @amelchio here: https://github.com/SoCo/SoCo/pull/763#issuecomment-740441769 and since this isn't a public api, but only used internally, I didn't see much problem with it..

> 
>     2. Does it make sense to rename `part1`/`part2` to reflect what's happening in the flow? Perhaps something like `begin_auth`/`get_auth_url` and then `complete_auth`?

Yeah, it might. @goldyfruit also wanted to rename but more in the direction of what they did, like "get_auth_token" which I didn't like too much. My main reason for that is that I really want the names to convey that it is a two part process, to make sure users don't misunderstand this, as that will probably cause issues. `begin_auth` and `complete_auth` also hints at this, so that could work as well. The question then is: There is more than DeviceLink and AppLink authentication supported by Sonos, so should the methods names include this information to make sure we don't get a name clash later, so something like `begin_device_or_app_link_auth` and ditto `complete`? I willing to adjust according to popular opinion on this, as long as we follow general Python naming guidelines.

> 
>     3. It seems the proper way to browse a service is to use the `get_metadata()` method. This was not intuitive to me. Should this either be renamed or a `browse` method added which wraps this call?

Yeah. I've been thinking something along the same lines. I think, since we can just add a method that wraps it, this could be a separate enhancement and be saving it for later, we have the option to think about whether it should do something more than just wrap. Since it does not need to be in this PR I would suggest to not do it here :D

amelchio:
> I know but this feature came at the direct request of @amelchio here: [#763 (comment)](https://github.com/SoCo/SoCo/pull/763#issuecomment-740441769) and since this isn't a public api, but only used internally, I didn't see much problem with it..

I didn't mean to insist on a particular implementation. That comment was just to note that we should preferably be able to support a multi-household setup with different accounts on each system, for example a child having its own Sonos with its own Spotify account.

soco/music_services/music_service.py
KennethNielsen: Ah yes, that option should probably be explicitly mentioned in the error message.

KennethNielsen: Added in 1a432b5

soco/music_services/music_service.py
KennethNielsen: Good point. Fixed in 976922d.

KennethNielsen:
> > I know but this feature came at the direct request of @amelchio here: [#763 (comment)](https://github.com/SoCo/SoCo/pull/763#issuecomment-740441769) and since this isn't a public api, but only used internally, I didn't see much problem with it..
> 
> I didn't mean to insist on a particular implementation. That comment was just to note that we should preferably be able to support a multi-household setup with different accounts on each system, for example a child having its own Sonos with its own Spotify account.

@amelchio apologies of that came out wrong, I didn't intend to imply that the particular implementation was on your request, only that you made me aware that we should probably support multi-household auth, that I agreed, and that I think the current implementation is ok for the purpose :)

KennethNielsen:
> > There is more than DeviceLink and AppLink authentication supported by Sonos, so should the methods names include this information to make sure we don't get a name clash later
> 
> The only other auth methods I know of are either `Anonymous` which _should_ not need authentication, and `UserId` which is a legacy method where we have no working examples. The `DeviceLink` and `AppLink` methods use a _very_ similar two-step process, with no indication that `UserId` would be similar. I would suggest keeping it simple with `begin_auth` and `complete_auth` and add a separate flow for `UserId` if we figure that one out.

@jjlawren agreed. @goldyfruit you ok with these names?

goldyfruit:
> > > There is more than DeviceLink and AppLink authentication supported by Sonos, so should the methods names include this information to make sure we don't get a name clash later
> > 
> > 
> > The only other auth methods I know of are either `Anonymous` which _should_ not need authentication, and `UserId` which is a legacy method where we have no working examples. The `DeviceLink` and `AppLink` methods use a _very_ similar two-step process, with no indication that `UserId` would be similar. I would suggest keeping it simple with `begin_auth` and `complete_auth` and add a separate flow for `UserId` if we figure that one out.
> 
> @jjlawren agreed. @goldyfruit you ok with these names?

All good for me!

KennethNielsen:
@jjlawren methods renamed. I also a few extra logging statements to the auth methods, which logs the instance ids, so that we can easily detect if the auth is being run on different objects.

jjlawren (approved):
A few suggestions for docstring cleanups and one additional suggestion to guard for missing parameters in `MusicService.complete_authentication()`.

I think any additional functional changes or fixes can be addressed in followups after release. 👍

soco/music_services/music_service.py
```diff
@@ -155,50 +196,146 @@ def call(self, method, args=None):
         try:
             result_elt = message.call()
         except SoapFault as exc:
+            if "Client.AuthTokenExpired" in exc.faultcode:
```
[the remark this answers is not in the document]
pwt: Let's work on this in the coming days.

pwt:
@KennethNielsen, @goldyfruit, @jjlawren: **great work** on this, folks!

I've applied all of the code review changes from @jjlawren. The one (known) open item is how to deal with expired tokens. Let's work on this for the next patch release.

Now, with a great sense of ceremony, I'm now going to merge this feature and prepare the v0.26.0 release.

KennethNielsen:
Thanks @pwt.

With regards to marking it experimental there should probably be a change to the README, changing the warning about no music services to saying they are back but highly experimental for this release. Do you want to do it or should I do a mini-PR?

pwt:
> Thanks @pwt.
> 
> With regards to marking it experimental there should probably be a change to the README, changing the warning about no music services to saying they are back but highly experimental for this release. Do you want to do it or should I do a mini-PR?

I've already changed it (#885).

PR #884, opened 2022-01-21
Use LRU caches for `from_didl_string` and `parse_event_xml`

jjlawren:
The `from_didl_string()` function is regularly called with the same input data when processing subscription events. By caching the results with `lru_cache`, we can avoid some processing overhead and also allow equality comparisons downstream for metadata as identical `DidlObject` instances are re-sent.

This uses the default size of 128 entries.

### DIFF
--- a/soco/data_structures_entry.py
+++ b/soco/data_structures_entry.py
@@ -3,7 +3,7 @@ objects from both music library and music service data structures
 
 """
 
-
+from functools import lru_cache
 import logging
 from urllib.parse import urlparse
 
@@ -19,6 +19,7 @@ _LOG.addHandler(logging.NullHandler())
 _LOG.debug("%s imported", __name__)
 
 
+@lru_cache
 def from_didl_string(string):
     """Convert a unicode xml string to a list of `DIDLObjects <DidlObject>`.
 
--- a/soco/data_structures_entry.py
+++ b/soco/data_structures_entry.py
@@ -19,7 +19,7 @@ _LOG.addHandler(logging.NullHandler())
 _LOG.debug("%s imported", __name__)
 
 
-@lru_cache
+@lru_cache()
 def from_didl_string(string):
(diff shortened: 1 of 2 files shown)


bdraco:
Will give this a test when I get home

soco/data_structures_entry.py
```diff
@@ -19,6 +19,7 @@
 _LOG.debug("%s imported", __name__)
 
 
+@lru_cache()
```
bdraco, line 22: Tested and working


You can check how effective it is with
https://docs.python.org/3/library/functools.html#functools.lru_cache

```
To help measure the effectiveness of the cache and tune the maxsize parameter, the wrapped function is instrumented with a cache_info() function that returns a named tuple showing hits, misses, maxsize and currsize.
```

jjlawren: I'm gathering some statistics and will share shortly. I'm also testing with adding a decorator to `parse_event_xml()` but that cache rate is a bit lower. Perhaps still useful, though.

jjlawren: After a few hours in a recent test:
* `from_didl_string`: CacheInfo(hits=347, misses=263, maxsize=128, currsize=128)
* `parse_event_xml`: CacheInfo(hits=220, misses=776, maxsize=128, currsize=128)

Even though the `parse_event_xml` hit rate is lower, I think it's still worth adding.

bdraco: That's still a decent hit rate and xml is painfully slow to parse

pwt:
LGTM! I'll hold for `parse_event_xml` cache addition, unless you'd like me to merge this separately.

jjlawren:
This is ready now.

pwt:
> This is ready now.

Sorry, I'd somehow missed that the change for `parse_event_xml` had already been added.

PR #886, opened 2022-01-23
Add Apple Music support to the ShareLink plugin

amelchio:


pwt:
Great addition -- thanks. Can't test it personally, but LGTM. Will be in v0.26.0, to be released shortly.

PR #894, opened 2022-02-14
Detect Sonos Radio as radio source

jjlawren:
This adds an entry for Sonos Radio URIs to properly categorize playing streams as radio for `music_source` related features.


pwt:
LGTM -- thanks.

PR #895, opened 2022-02-14
Detect Spotify Connect as music source

jjlawren:
Report a new `music_source` value when playing via Spotify Connect (i.e., casting from the native Spotify app), similar to how AirPlay is reported.


pwt:
LGTM -- thanks.

PR #896, opened 2022-02-14
Reprocess radio tracks if metadata incomplete

jjlawren:
Some services (notably Sonos Radio) act as radio stations, but do not encode the current track's metadata in a `streamContent` element. This change will reprocess metadata if the radio decoding process does not populate the necessary fields.

Note: We only check if `artist` is empty as `title` is almost always set and `album` is less commonly provided. If `artist` is missing we can be fairly confident that it's worth reprocessing with the standard method.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1896,16 +1896,21 @@ class SoCo(_SocoSingletonBase):
 
             return radio_track
 
-        # Duration seems to be '0:00:00' when listening to radio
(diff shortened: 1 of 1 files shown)


jjlawren:
One possible regression is that the reprocessing step will undo the cleanup which avoids using URIs as the title:
https://github.com/SoCo/SoCo/blob/master/soco/core.py#L1886-L1893

pwt:
LGTM --thanks.

PR #897, opened 2022-02-15
Fix python iteration error

relevitt:
This minor change fixes an iteration error I've encountered since upgrading to python 3.8. Without this fix, I get:

`builtins.RuntimeError: dictionary keys changed during iteration`


pwt:
LGTM -- thanks for contributing this fix.

PR #893, opened 2022-02-10
Attempt to refresh expired music service tokens (fixes#890)

goldyfruit:
This will try to find first the token and the key into the namespace
and fallback to `findtext()` if nothing has been found.

PR #899, opened 2022-02-17
Check for empty token refresh to avoid exceptions

jjlawren:
Minor followup to #893 which prevents exceptions if the `.find()` calls do not succeed and return `None`.

Example traceback this PR addresses:
```python
>>> my_music_service.search("artists", "M83")
  File "/src/SoCo/soco/music_services/music_service.py", line 199, in call
    result_elt = message.call()
  File "/src/SoCo/soco/soap.py", line 309, in call
    raise SoapFault(faultcode, faultstring, faultdetail)
soco.soap.SoapFault: Client.TokenRefreshRequired: Token has expired

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/src/SoCo/soco/music_services/music_service.py", line 916, in search
    response = self.soap_client.call(
  File "/src/SoCo/soco/music_services/music_service.py", line 233, in call
    auth_token = exc.detail.find(
AttributeError: 'NoneType' object has no attribute 'text'
```

### DIFF
--- a/soco/music_services/music_service.py
+++ b/soco/music_services/music_service.py
@@ -230,10 +230,14 @@ class MusicServiceSoapClient:
                 # </detail>
                 auth_token = exc.detail.find(
                     ".//xmlns:authToken", {"xmlns": self.namespace}
-                ).text
+                )
+                if auth_token:
+                    auth_token = auth_token.text
                 private_key = exc.detail.find(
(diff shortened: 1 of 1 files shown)


pwt:
LGTM -- thanks.

Issue #903, opened 2022-02-28
Missing title in get_current_track_info


LionelCons:
When investigating a missing title in my audio software, I traced it back to some incorrect output from `get_current_track_info`.

In short, the following code:
```
print(device.get_current_track_info())
```
gives:
```
{'title': '', 'artist': 'Angel City', 'album': 'Face to Face', 'album_art': '...', 'position': '0:01:59', 'playlist_position': '9', 'duration': '0:03:14', 'uri': '...', 'metadata': '<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/" xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/" xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/"><item id="-1" parentID="-1" restricted="true"><res protocolInfo="x-file-cifs:*:audio/mpeg:*" duration="0:03:14">...</res><r:streamContent></r:streamContent><upnp:albumArtURI>...</upnp:albumArtURI><dc:title>Waiting for the World</dc:title><upnp:class>object.item.audioItem.musicTrack</upnp:class><dc:creator>Angel City</dc:creator><upnp:album>Face to Face</upnp:album><upnp:originalTrackNumber>9</upnp:originalTrackNumber><r:albumArtist>Angel City</r:albumArtist></item></DIDL-Lite>'}
```

The track title is present in the metadata but not in the dict.

I'm using the latest version from `pip`: 0.26.3.

jjlawren:
I'm investigating this in a downstream project now. It's likely a side-effect of #896 in regards to local music via CIFS/Samba file shares. You redacted the value, but the `uri` is very important here.

LionelCons:
Here is the `uri` in the dict, less redacted:
```
'uri': 'x-file-cifs://xxx/music/Angel City - Face to Face/09 - Angel City - Waiting for the World.mp3'
```

jjlawren:
Yep, that's the same issue I'm seeing elsewhere. Working on a fix.

jjlawren:
@LionelCons would you be willing to test #906 and provide feedback?

LionelCons:
I've tried with the new `core.py` and the title information is now back.

Thanks!

pwt:
Fix released in SoCo v0.26.4. Thanks for spotting this issue.

PR #906, opened 2022-03-01
Check metadata title for URI hints before discarding

jjlawren:
A change in #896 caused certain titles (most obviously from CIFS shares) to be incorrectly discarded as they were contained in the track URI. This was naive as even single-word strings would trigger the discard behavior. This PR adds an additional hurdle which requires the track to contain some URI components before it is discarded, such as path and query arguments or an `.m3u8` extension.

I've also added tests to validate expected processing of metadata from a variety of sources. This should help ensure we're going all we can to get correct results and also prevent future regressions.

Note: This includes the commit from #904 in order to pass tests. Once that PR is merged this can be rebased to simplify the changes.

Fixes: #903

### DIFF
--- a/tests/data/media_metadata_payloads/tunein.json
+++ b/tests/data/media_metadata_payloads/tunein.json
@@ -1,0 +1,23 @@
+{
+    "input": {
+        "Track": "1",
+        "TrackDuration": "0:00:00",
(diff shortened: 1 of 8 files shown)


tests/data/media_metadata_payloads/tunein.json
```diff
@@ -0,0 +1,23 @@
...
+        "TrackURI": "hls-radio://http://n30a-e2.revma.ihrhls.com/zc000/hls.m3u8?TGT=TuneIn&maxServers=2&partnertok=somesuperlongtoken&rj-ttl=5&DIST=TuneIn&rj-tok=000000000000000000000_0000",
+        "RelTime": "0:04:14",
+        "AbsTime": "NOT_IMPLEMENTED",
+        "RelCount": "2147483647",
+        "AbsCount": "2147483647"
+    },
+    "result": {
+        "title": "text=\"Spot Block End\" amgTrackId=\"1234567\"",
```
jjlawren, line 13: This is not handled ideally, but it mirrors what is displayed in the native Sonos app. Perhaps should also be discarded?

jjlawren:
Radio stations _usually_ provide the station name in a separate call (`SoCo.get_current_media_info()["channel"]`) which could be used as a fallback title, but that should probably be handled by the downstream application.

pwt:
> Radio stations _usually_ provide the station name in a separate call (`SoCo.get_current_media_info()["channel"]`) which could be used as a fallback title, but that should probably be handled by the downstream application.

Just to note that to get radio show names, I fall back to using avTransport events with something like:
```Python
sub = speaker.avTransport.subscribe()
event = sub.events.get(timeout=0.5)
radio_show = event.variables[
       "current_track_meta_data"
].radio_show.rpartition(",")[0]
```

jjlawren:
> Just to note that to get radio show names, I fall back to using avTransport events with something like:

Yep, that's the exact same data as what `get_current_media_info()` provides and how I solve this problem downstream today. I was talking in terms of a pure polling workflow using the built-in methods and making this "automatic" for simpler use-cases.

pwt:
> > Just to note that to get radio show names, I fall back to using avTransport events with something like:
> 
> Yep, that's the exact same data as what `get_current_media_info()` provides and how I solve this problem downstream today. I was talking in terms of a pure polling workflow using the built-in methods and making this "automatic" for simpler use-cases.

`get_current_media_info()` gets the radio channel, but not the radio show (i.e., the equivalent of the title). At least, not in my experimentation.

jjlawren:
> `get_current_media_info()` gets the radio channel, but not the radio show (i.e., the equivalent of the title). At least, not in my experimentation.

Ah, yes, I didn't read the attributes closely enough. You're absolutely right and I have not seen that payload in polling calls, either.

I was thinking about the radio _station_, which would require an additional call to the speaker. The radio show is simply not available without separately consuming subscription events.

pwt:
I've merged #904. @jjlawren Do you want to rebase this PR, or should I commit it as-is? Ah ... you just answered my question :)

Once committed, I'll issue a patch release.

Nice work. Great to see all the validation tests.

PR #907, opened 2022-03-01
Handle lists of media in the Plex plugin; allow insertion by position

jjlawren:
This adds two new features to the Plex plugin:
1. Allows inserting media into the queue at a specific index position.
2. Accepts lists of Plex media items to both `play_now()` and `add_to_queue()`.

The `add_next` parameter has been removed and `as_next` has been added instead. The `as_next` parameter will enqueue the item next in the queue _only_ if currently in shuffle mode. This matches the model used by `ShareLinkPlugin` and now relies on default Sonos behavior. As `position` is now exposed, it will be up to the user if they would like to enqueue media into the "next" playback position.

### DIFF
--- a/soco/plugins/plex.py
+++ b/soco/plugins/plex.py
@@ -111,7 +111,7 @@ class PlexPlugin(SoCoPlugin):
         self.soco.play_from_queue(position)
 
     def add_to_queue(
-        self, plex_media, add_next=False
+        self, plex_media, add_next=False, position=None
     ):  # pylint: disable=too-many-locals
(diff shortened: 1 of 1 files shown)


jjlawren:
I'm reviewing the implementation here and think it makes sense to normalize it to match how `ShareLinkPlugin` works with the queue. Specifically, the `as_next` being passed through as-is and only having an effect when in shuffle mode. This will make things a bit simpler as there will be (slightly) less magic happening behind the scenes.

pwt:
LGTM -- thanks.

PR #909, opened 2022-03-08
Add support for subwoofer gain controls

jjlawren:
Adds support for controlling the subwoofer gain using the `SoCo.sub_gain` property. This property name matches the key from the `RenderingControl` event payload.

Needs validation testing from a system that contains a subwoofer.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1141,15 +1141,15 @@ class SoCo(_SocoSingletonBase):
     def sub_gain(self, level):
         """Set the subwoofer gain level.
 
-        :param level: Desired subwoofer gain level (-10 to 10)
+        :param level: Desired subwoofer gain level (-15 to 15)
         :type level: int
         """
         if not self.has_subwoofer:
             raise NotSupportedException("This group does not have a subwoofer")
 
-        if not -10 <= level <= 10:
+        if not -15 <= level <= 15:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


soco/core.py
```diff
@@ -1122,6 +1123,43 @@ def sub_enabled(self, enable):
...
+    def sub_gain(self, level):
+        """Set the subwoofer gain level.
+
+        :param level: Desired subwoofer gain level (-10 to 10)
+        :type level: int
+        """
+        if not self.has_subwoofer:
+            raise NotSupportedException("This group does not have a subwoofer")
+
+        if not -10 <= level <= 10:
```
bachya, line 1150: In my Sonos iOS app, the gain goes from -15 to 15.

bachya (changes requested):
Tested both the setter and getter and they look good: the Sonos app properly reflects the changes in both cases. Need to change the level limits.

jjlawren:
@bachya Thanks for testing, adjusted thresholds based on your feedback. 👍

pwt:
I'll test (and presumably merge) this later today. Sorry for the delay.

pwt:
Working perfectly with my Play:1 x 2 + Sub rig. Also behaving as expected when tested against a device without a Sub. 
Thanks for your work on this -- it's something I've (selfishly) wanted for a while.

I'll merge as-is, but it would be nice to add a test in a subsequent PR.

PR #911, opened 2022-04-05
Pass kwargs for ShareLink/Plex queue methods

jjlawren:
Allow passing `**kwargs` to the underlying service call for the ShareLink `add_share_link_to_queue()` and Plex `add_to_queue()` plugin methods. My primary reason is to allow differing `timeout` arguments for different services. Spotify, for example, seems to need a longer timeout when adding playlists with many tracks.


jjlawren:
Lint complaints can be more easily cleaned up after #912.

pwt:
LGTM, thanks.

PR #912, opened 2022-04-05
Update pylint config, remove unnecessary pragmas

jjlawren:
This updates the pylint configuration to disable more `too-many-*` checks and add comments with reasoning.

Also removes any existing pragmas in the codebase which match these new global disables.


pwt:
Yes, might as well factor these out. Thanks.

PR #913, opened 2022-04-12
Refactor ZoneGroupState parsing

jjlawren:
This takes a few new approaches to `ZoneGroupState` handling:
* Creates a new ZoneGroupState module/class
* Tracks ZGS per-household instead of per-device
* Skips reprocessing ZGS payloads when possible
* Normalizes ZGS XML using XSL transformation (adds `lxml` dependency)
* Processes ZGS payloads received from subscription events (only `events_asyncio` for now)
* Uses extended cache timeout (5s -> 60s) for payloads received from event callbacks as we can assume future updates will be delivered as they occur
* Tracks statistics to measure effectiveness

These payloads are (mostly) identical across all speakers in a given household, so they can be processed centrally. Each speaker has its own "view" and the groups/members may be shown in different orders. An XSL transform is used to normalize the payloads into identical formats. This also means we can keep a single cache instead of a cache per-speaker, which makes things much more efficient can avoid more network requests.

In my local setup over about a day of activity, only about 200 out of 3000 ZGS payloads/requests needed to be fully processed. The rest were loaded from the cache or discarded as they were identical to the last-processed payload.


jjlawren:
I've been running this branch with positive results as a drop-in replacement in a downstream project.

It's a change to core functionality and I'm interested in any feedback.

pwt:
Just to note for other reviewers: this PR introduces a new external requirement **[lxml](https://lxml.de)** which has a binary component. It appears to be a widely used and well-supported library.

As a datapoint, I've found that prebuilt wheels are available for all the platforms I've experimented with (macOS, Linux x64, Linux ARM, Windows), with various versions of Python (3.5 and up) and I've experienced no installation glitches.

Note that it's on the large side for a dependency. A complete SoCo installation currently occupies ~4.3MB. lxml adds about 7MB, taking the footprint up to ~11.3MB.

jjlawren:
There is also the future potential to migrate all XML handling to the `lxml` library. It's unfortunately not a drop-in replacement for this project, but it's built to be mostly compatible with the ElementTree APIs. Once those (hopefully minor) differences are worked through it could improve the performance of the whole project.

KennethNielsen:
@jjlawren @pwt just so we are all on the same side here, I'm actually in favor of migration _away_ from ElementTree as it has more than one deficiency that has annoyed me for years working on SoCo, although I'm unsure if any of them has been fixed. Amongst the ones I remember which are strong arguments for are;

1. Clumsy, bordering on unworkable, namespace handling 
2. Only a restricted sub-set of XPath being supported and the subset is **just a little** to small

So, I think my only concern here was that if we did it, someone would show up and say "you broke my SoCo script on [(EXOTIC) PLARFORM X]", where the most likely  not so exotic one would be Raspberry Pi. So, since @pwt did the research on that already and confirms that it runs on ARM, I no longer worry about that. With regards to the installation size going from 5 to 12 MB, I am very unconcerned about this, as I don't know of a platform where SoCo would be used with more restricted available memory than the Raspberry Pi, which always comes with a GB data card anyway.

So, all in all I don't think there is a problem with lxml. I'll review the PR as soon as I can.

pwt:
One additional installation datapoint: `lxml` installed without issue on my new M1 Pro MacBook, but there wasn't a pre-built wheel, so a build was required. I don't know if the build process depends on development tools that may not be present on all Apple Silicon Macs.

I suspect that the absence of a pre-built wheel is a temporary issue, however.

soco/zgs.py
```diff
@@ -0,0 +1,267 @@
...
+similar contents, payloads are passed through an XSL transformation
+to normalize the data, to allow simple equality comparisons, and to
+avoid unnecessary reprocessing of identical data.
+
+Since the payloads are identical between all speakers, we can use a
+common cache per household.
+
+Satellite devices can sometimes deliver outdated payloads when they
+are directly polled. These requests are instead redirected to their
+parent device.
```
rytilahti, line 19:
```suggestion
As satellite devices can sometimes deliver outdated payloads when they
are directly polled, these requests are instead redirected to their
parent device.
```
If I'm reading correctly, that only the parent is being accessed no matter whether "these" requests are made to any of the children?

soco/zgs.py
```diff
@@ -0,0 +1,267 @@
...
+
+Example payload contents:
```
rytilahti, line 21: Instead of having an example payload here, perhaps it would be better to have this as a fixture file and add tests that run on that? That would also be better for maintainability (just in case the structure changes at some point).

soco/zgs.py
```diff
@@ -0,0 +1,267 @@
...
+
+        for group_element in zone_groups.findall("ZoneGroup"):
+            coordinator_uid = group_element.attrib["Coordinator"]
+            group_uid = group_element.attrib["ID"]
+            group_coordinator = None
+            members = set()
+            for member_element in group_element.findall("ZoneGroupMember"):
+                zone = self.parse_zone_group_member(member_element)
+                zone._is_satellite = False
+                zone._satellite_parent = None
```
rytilahti, line 237: It would be cleaner if you would set these inside `parse_zone_group_member`. Same applies probably on all other cases below, where a protected member is being set based on accesses on `member_element`.

rytilahti (commented):
My personal opinion is that it's great to see `core.py` being cleaned up, so :+1: I'm just wondering if it'd make sense to name the new file `zonegroupstate.py` instead of its abbreviation (which I think is fine to use inside the module itself)?

soco/zgs.py
jjlawren: It could, but `parse_zone_group_member` is called for both the primary & satellite devices in this method. In the latter case it would be set these values to the defaults and then immediately override them after the call returns. I think the current logic is _slightly_ easier to follow as it sets both to the expected value once.

soco/zgs.py
jjlawren: Yes, I agree.

jjlawren: I forgot, there are already tests in place for ZGS parsing: https://github.com/SoCo/SoCo/blob/48e3b705d3ec05acd993de33aacda3ce8f96349e/tests/test_core.py#L162-L166

soco/zonegroupstate.py
```diff
@@ -0,0 +1,267 @@
...
+            # Satellites can return outdated information, use the parent
+            _LOG.debug(
+                "Poll request on satellite (%s), using parent (%s)",
+                soco.ip_address,
+                soco._satellite_parent.ip_address,
+            )
+            soco = soco._satellite_parent
+
+        zgs = soco.zoneGroupTopology.GetZoneGroupState()["ZoneGroupState"]
+        self.process_payload(zgs, "poll", soco.ip_address)
```
KennethNielsen, line 152: Maybe add keywords to the last two argument, to indicate their function. I know that at present the called method is right below, but that might change.

soco/zonegroupstate.py
```diff
@@ -0,0 +1,267 @@
...
+
+        def update_cache():
+            if source == "event":
+                timeout = EVENT_CACHE_TIMEOUT
+            else:
+                timeout = POLLING_CACHE_TIMEOUT
+            self._cache_until = time.monotonic() + timeout
+            _LOG.debug("Setting ZGS cache to %ss", timeout)
+
+        tree = normalize_zgs_xml(payload)
```
KennethNielsen, line 166: That's a neat trick

soco/zonegroupstate.py
```diff
@@ -0,0 +1,267 @@
...
+        State, create a SoCo instance for the member, set basic attributes
+        and return it."""
+        # pylint: disable=protected-access
+
+        # Create a SoCo instance for each member. Because SoCo
+        # instances are singletons, this is cheap if they have already
+        # been created, and useful if they haven't. We can then
+        # update various properties for that instance.
+        member_attribs = member_element.attrib
+        ip_addr = member_attribs["Location"].split("//")[1].split(":")[0]
```
KennethNielsen, line 199: For extraction of information from strings I find it tremendously helpful with an example string in a comment above right above it. I know that it is mentioned in dosctring at the module level, but I find that the locality makes it easier to read.

soco/zonegroupstate.py
```diff
@@ -0,0 +1,267 @@
...
+        zone = config.SOCO_CLASS(ip_addr)
+        for key, attrib in ZGS_ATTRIB_MAPPING.items():
+            setattr(zone, attrib, member_attribs.get(key))
+
+        for channel_map in list(
+            filter(None, [zone._channel_map, zone._ht_sat_chan_map])
```
KennethNielsen, line 205: Ok so, programming style preferences aside, I think the `filter(None` trick is anything but easily parseable, especially compared to its list comprehension counterpart:
```python
       for channel_map in [
            map for map in [zone._channel_map, zone._ht_sat_chan_map] if map is not None
       ]:
```

soco/zonegroupstate.py
```diff
@@ -0,0 +1,267 @@
...
+        ):
+            for channel in channel_map.split(";"):
```
KennethNielsen, line 207: Again, might be helpful with a local example of what the channel_map looks like

KennethNielsen (approved):
I finally found time to review this. I have no comments on semantics, it all looks great. I only have a few minor and optional style requests. Besides from that, tests would have been great, but I realize it is also late in the commenting cycle to request that, so maybe in another PR.

soco/zonegroupstate.py
jjlawren: This was copied over from `core.py` as-is. I'll add a comment with an example.

soco/zonegroupstate.py
jjlawren: Done.

soco/zonegroupstate.py
jjlawren: Done.

soco/zonegroupstate.py
jjlawren: Done.

soco/zonegroupstate.py
jjlawren: I personally find the `filter` method easier to follow, but I'm open to suggestions. Anyone else want to weigh in here? 😄

jjlawren:
As for tests, I first wanted to make sure the existing ZGS tests passed untouched. The new functionality here is heavily focused on optimizations around throttling, payload normalization, and handling event callbacks. Those will take a bit more time to set up proper tests for, but I've validated those features fairly extensively in a downstream project.

soco/zonegroupstate.py
pwt: The `filter()` approach looks fine to me. Appears to be quite standard usage: https://docs.python.org/3/library/functions.html#filter.

(I had to remind myself how `filter()` works, but that's just me.)

pwt:
> As for tests, I first wanted to make sure the existing ZGS tests passed untouched. The new functionality here is heavily focused on optimizations around throttling, payload normalization, and handling event callbacks. Those will take a bit more time to set up proper tests for, but I've validated those features fairly extensively in a downstream project.

I agree with @KennethNielsen that tests can be added later. Obviously, their major utility is in preventing future regressions; I'm sure you're correct that the changes are solid as they currently stand.

soco/zonegroupstate.py
KennethNielsen: > I personally find the filter method easier to follow

really, even with the None. I mean, first argument is supposed to be a callable that will determine whether we pass the object on or not. The fact that you can pass None is there instead seems a little special.

> (I had to remind myself how filter() works, but that's just me.)

:D my point exactly. Did you have to look stuff up to read the list comprehension? Or would you have to if the list comprehension had covered a `map?` or `reduce` :D ?

Oh man, I really regret that time that Guido let his guard down over summer and let all these things in ;p In any case, never mind me, I like the prospect of functional programming and all the little tools that come with it, I just find it inherently more implicit and "_explicit is better than implicit_". In any case, this is style _of semantics_ and I don't dictate that for SoCo, in general we tend to go with consensus on that. So if you are both ok with it, just ignore me :)

KennethNielsen (approved):
Looks good to me.

pwt:
> Looks good to me.

Thanks for taking the time to review this, @KennethNielsen.
I'll merge this now, and create the v0.28 release in the next day or two.

KennethNielsen:
> > Looks good to me.
> 
> Thanks for taking the time to review this, @KennethNielsen.
> I'll merge this now, and create the v0.28 release in the next day or two.

No problem

PR #915, opened 2022-05-27
Add convenience wrappers for surround_level properties

jjlawren:
Adds wrappers to the TV & music `surround_volume` properties to allow direct mapping to the native subscription event payload names.

A similar solution was previously implemented in https://github.com/SoCo/SoCo/pull/874.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1372,12 +1372,12 @@ class SoCo(_SocoSingletonBase):
 
     @property
     def surround_level(self):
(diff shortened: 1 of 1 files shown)


pwt:
LGTM. These convenience functions aren't listed in the docstrings at the start of the file. I think that's fine given that they are essentially duplicate functionality, but perhaps you could confirm this is your intention, @jjlawren.

PR #924, opened 2022-10-14
Pass kwargs to queue add service calls

jjlawren:
This will pass through any kwargs provided to `add_to_queue`, `add_uri_to_queue`, and `add_multiple_to_queue` methods. The primary reason is to allow setting a (usually much longer) service `timeout` value per-call.


pwt:
LGTM.

PR #925, opened 2022-10-14
Explicitly set default timeouts for remaining `requests` calls

jjlawren:
Explicitly add the `timeout` parameter to the remaining `requests` calls.

A current bump to `pylint` (which is not currently pinned) added this check which is making CI runs fail.

### DIFF
--- a/soco/music_services/accounts.py
+++ b/soco/music_services/accounts.py
@@ -9,7 +9,7 @@ import weakref
 
 import requests
 
-from .. import discovery
+from .. import config, discovery
 from ..xml import XML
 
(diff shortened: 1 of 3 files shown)


pwt:
The change to `tests/test_soap.py` breaks the test for me when running them locally. Try running: `pytest -k test_call .`.

I'll start looking at getting the check jobs functioning again.

pwt:
Hi @jjlawren: can you pull in the changes from the current master and massage the conflicts in `soco/music_services/accounts.py` and `soco/soap.py` where I've temporarily added the `pylint: disable=` comments.

Then, let's see if the check jobs run successfully, because the `test_call` test is still failing for me locally.

PR #926, opened 2022-10-14
Attempt to fix flake8 runs on 3.7

jjlawren:
This attempts to fix the CI by pinning a `flake8` dependency to an older version on py37.


jjlawren:
This seems to fix the `flake8`/py37 issue for now. The remaining CI failures are addressed in https://github.com/SoCo/SoCo/pull/925.

pwt:
Thanks for figuring this out.

Issue #935, opened 2022-11-15
SoCo seems not work if there are more than 24 Sonos devices in one household


tomhht:
Hi, 
My English is poor, any inconvenience caused is regretted.
SoCo could do nothing when there are more than 24 devices in one household. The discover feature would fail, throwing SoCoUPnPException. None is contactable (soco.discovery.contactable).
I'm not sure if I missed some answers about it. Have you ever met this?

Cheers
Tom

pwt:
I haven't seen this reported before. Could you try discovery using:
```
soco.discovery.scan_network()
```
and see what that returns? Thanks.

It's possible we may need to do something specific to handle large populations of speakers; perhaps Sonos works differently for large systems.

In the meantime, if you know the IP address of a speaker, you can circumvent discovery by creating a SoCo object directly, e.g.:

```
from soco import SoCo
speaker = SoCo("192.168.0.50")
zones = speaker.all_zones
```

amelchio:
> It's possible we may need to do something specific to handle large populations of speakers; perhaps Sonos works differently for large systems.

https://github.com/SoCo/SoCo/issues/518#issuecomment-569832252

tomhht:
@pwt
Probably I have exactly the same issue as that @amelchio mentioned. `soco.discover`   `soco.discovery.scan_network` and _creating SoCo object directly_ all throw `soco.exceptions.SoCoUPnPException: UPnP Error 501 received: Action Failed from 192.168.1.5`.

Many thanks to you both.

I'm fresh meat here, so not sure if I should close this issue? Since it seems no update regarding #518 for now.

pwt:
> I'm fresh meat here, so not sure if I should close this issue? Since it seems no update regarding #518 for now.

Hi @tomhht : Thanks for checking. Leave it open for now. I'll take a look when I get a bit of time, and reconcile the two issues.

The issue I'll face is that I won't be able to test any solutions, since I don't have the required population of speakers. You may be able to help with testing.

tomhht:
@pwt 
I'm glad if I could be helpful.

jjlawren:
(Cross-posting from https://github.com/home-assistant/core/issues/82513#issuecomment-1356604402)

Based on nothing but a hunch, I'm wondering if there's a limitation in the Sonos firmware which causes it to fail when ZoneGroupState payloads exceed some set size. Since it seems like the native Sonos app can get away without _polling_ this endpoint, perhaps the subscription callbacks don't have this limitation.

[...]

Issue #518, opened 2017-09-15
Getting a SoCoUPnPException: UPnP Error 501 received: Action Failed from <player ip> on any request that calls GetZoneGroupState()


powertieke:
soco.discover() and SoCo('<ip>').player_name give this error...

soco.exceptions.SoCoUPnPException: UPnP Error 501 received: Action Failed from <ip>

DPH:
Hi @powertieke,
Assume you can play music on your Sonos with the normal controllers?
Assume you are running your code on the same network as your Sonos is?
Try running this (with one of your IP addresses) and post the results:
```
import soco
s = soco.SoCo('192.168.1.99')
print(s.player_name)
```
Cheers David

powertieke:
All assumptions are correct.

Created a script containing the lines above.

> (soco-env) reallymycomputername:sonos_webinterface reallymyusername$ python showname.py
> Traceback (most recent call last):
> File "showname.py", line 4, in <module>
print(pl.player_name)
> File "/Users/reallymyusername/code/sonos_webinterface/soco-env/lib/python3.5/site-packages/soco-0.12_-py3.5.egg/soco/core.py", line 249, in player_name
 >  File "/Users/reallymyusername/code/sonos_webinterface/soco-env/lib/python3.5/site-> packages/soco-0.12_-py3.5.egg/soco/core.py", line 939, in _parse_zone_group_state
>   File "/Users/reallymyusername/code/sonos_webinterface/soco-env/lib/python3.5/site-packages/soco-0.12_-py3.5.egg/soco/services.py", line 644, in GetZoneGroupState
>   File "/Users/reallymyusername/code/sonos_webinterface/soco-env/lib/python3.5/site-packages/soco-0.12_-py3.5.egg/soco/services.py", line 408, in send_command
>   File "/Users/reallymyusername/code/sonos_webinterface/soco-env/lib/python3.5/site-packages/soco-0.12_-py3.5.egg/soco/services.py", line 469, in handle_upnp_error
> soco.exceptions.SoCoUPnPException: UPnP Error 501 received: Action Failed from 192.168.1.222

Requests for local settings like bass, treble, volume and loudness return the correct values, and can be set as expected. I can also get the player name using the `s.deviceProperties.GetZoneAttributes()["CurrentZoneName"]` method mentioned in the SoCo.player_name() method code.

KennethNielsen:
One step at a time.

[...]

PR #937, opened 2022-12-18
Improve collection of Zone Group State for large Sonos systems (partial fix for #518 / #935)

pwt:


jjlawren:
I'm a little hesitant about this approach:
1. Not every network can handle subscription callbacks properly.
2. This opens/closes a listening socket every time the ZGS is requested, which could cause other side effects.
3. What happens if a different event handler has already been loaded, such as the async implementation?

pwt:
Thanks for the comments, @jjlawren.

> 1. Not every network can handle subscription callbacks properly.

The events-based approach is only invoked if a normal query fails. It's a last resort, so if it fails due to network limitations it's still better than not trying.

> 2. This opens/closes a listening socket every time the ZGS is requested, which could cause other side effects.

The cacheing logic limits ZGS refreshes to at most once per 5 seconds by default (using `POLLING_CACHE_TIMEOUT`). That seems reasonable to me.

> 3. What happens if a different event handler has already been loaded, such as the async implementation?

The logic currently only handles the standard events module. I'll try adding support for the others, although I could apply the same 'it's better than nothing' rationale as for (1).

Of course, this is all predicated on this approach actually working for large speaker populations! Reading through the issues (e.g., https://github.com/SoCo/SoCo/issues/518#issuecomment-356750673) makes me optimistic that it might. I'm looking forward to seeing some results from the use of your test script on suitably large systems.

pwt:
Hi @jjlawren:

I've now added support for `events_asyncio` which appears to work but, due to my inexperience with asyncio development, clearly has room for improvement.

If you get a moment, perhaps you could take a look at ~ lines 181-207 and point out the right way to approach this? In particular I couldn't figure out how to retrieve the ZGS from the callback without resorting to a global.

jjlawren:
I'd really like to have confirmation that this is the actual underlying issue first.

In my downstream implementation that uses `soco`, I think I'd actually prefer to know that the polling call failed (at least optionally) so I can handle the issue myself by creating my own long-lived subscriptions. Perhaps with a special exception that better describes why this is failing, like `SonosNotSupported` to indicate it's a Sonos-side limitation. The automatic fallback fixes here are complex and are not risk-free.

pwt:
> I'd really like to have confirmation that this is the actual underlying issue first.

Agree.

> In my downstream implementation that uses `soco`, I think I'd actually prefer to know that the polling call failed (at least optionally) so I can handle the issue myself by creating my own long-lived subscriptions. Perhaps with a special exception that better describes why this is failing, like `SonosNotSupported` to indicate it's a Sonos-side limitation. The automatic fallback fixes here are complex and are not risk-free.

I think I'd like a default that works without plunging users into having to understand how to subscribe to and process ZGT events, etc. I take your point, though, so we should provide a config option to opt out of the fallback behaviour.

pwt:
In its current state this PR is a sticking plaster that falls back to using a ZoneGroupTopology event to determine the ZoneGroupState if a direct `GetZoneGroupState()` lookup fails. 

- The fallback behaviour can be disabled by setting `config.ZGT_EVENT_FALLBACK` to `False`, in which case the `SoCoUPnPException` can be handled elsewhere by the caller
- The fallback only works when using the standard `soco.events` module, not `soco.events_twisted` or `soco.events_asyncio`

It's likely that if `GetZoneGroupState()` fails once then it will continue to fail. Hence, an improvement after the first failure would be to avoid the expense of the initial `GetZoneGroupState()` call and use the event fallback to begin with. This should probably be coupled with a periodic reset to confirm that `GetZoneGroupState()` is still failing.

pwt:
After the first fallback to using events, this now becomes the default behaviour within `EVENT_FALLBACK_TIMEOUT` (currently set to 120s). This saves `GetZoneGroupState()` calls that will probably fail.

pwt:
Hi @jjlawren: From my perspective this is ready to merge now, and your review would be welcome.

There's support for falling back to using a ZGT event under the default `events` module and when using `events_asyncio`. Support for `events_twisted` can be added later by someone with the required expertise; the code scaffolding is there for it.

Hi @KennethNielsen FYI, in case you want to take a look.

I'd like to land this in v0.29, to be released within the next week or so.

soco/zonegroupstate.py
```diff
@@ -148,8 +150,84 @@ def poll(self, soco):
...
+        Update ZGS using events_asyncio. When the event is received,
+        the events_asyncio library will call 'process_payload' with
+        the updated ZGS.
+        """
+        from . import events_asyncio  # pylint: disable=C0415
+
+        sub = await speaker.zoneGroupTopology.subscribe()
+        await asyncio.sleep(1.0)
+        await sub.unsubscribe()
+        await events_asyncio.event_listener.async_stop()
```
jjlawren, line 230: What happens here if other subscriptions were created elsewhere?

pwt: Good point. The `async_stop()` will stop the event listener for all subscriptions. I'll add a test to stop the event listener only if it was not running prior to the `subscribe()`, in which case it would have been started by our `subscribe()` call.

soco/zonegroupstate.py
```diff
@@ -148,8 +150,88 @@ def poll(self, soco):
...
+            # Explicit asyncio event loop control required for Python 3.6
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+            loop.run_until_complete(ZoneGroupState.update_zgs_by_event_asyncio(speaker))
+            asyncio.set_event_loop(None)
+            loop.close()
+            # From Python 3.7, we can just use the single statement:
+            # asyncio.run(ZoneGroupState.update_zgs_events_asyncio(speaker))
+
+        if config.EVENTS_MODULE.__name__ == "soco.events_twisted":
```
KennethNielsen, line 202: Do you maybe want to create a if-elif-elif-else here, with an exception not only in the last elif (events_twisted), but also in the else case. It would make sure that it would fail loudly if we ever grow yet **another** framework for handling events.

soco/zonegroupstate.py
```diff
@@ -148,8 +150,88 @@ def poll(self, soco):
...
+            raise SoCoException(
+                "ZGT event fallback not supported when using the "
+                "'events_twisted' module"
```
KennethNielsen, line 205: The wording here could be "not yet implemented" to suggest that it is possible but waiting for eager fingers.

soco/zonegroupstate.py
```diff
@@ -148,8 +150,88 @@ def poll(self, soco):
...
+        """
+        Update ZGS using events_asyncio. When the event is received,
+        the events_asyncio notify handler will call 'process_payload' with
+        the updated ZGS.
+        """
+        from . import events_asyncio  # pylint: disable=C0415
+
+        event_listener_is_running = events_asyncio.event_listener.is_running
+        sub = await speaker.zoneGroupTopology.subscribe()
+        await asyncio.sleep(1.0)
```
KennethNielsen, line 229: I'm wondering of the "sleep 1" is necessary, when the other variants only have a timeout? But also I don't really understand the rest. Forgive me, async is not me strongest suit. But shouldn't this somehow get and process and events like the non async version and wont the awaiting of that event remove remove the need for the the sleep and be equivalent to the timeout?

KennethNielsen (commented):
I have a few questions, mostly related to async I probably just don't understand.

soco/zonegroupstate.py
pwt: Good point. I'll make this change.

soco/zonegroupstate.py
pwt: Agreed.

soco/zonegroupstate.py
pwt: Not my strongest area, either. I think a `sleep()` is needed to give other coroutines a chance to run. However, I'll experiment with shorter durations, and also with omitting it completely.

KennethNielsen: But what about the question about actually getting the payload. It is somehow happening behind the scenes or am I missing it?

KennethNielsen: > Not my strongest area, either. I think a sleep() is needed to give other coroutines a chance to run. However, I'll experiment with shorter durations, and also with omitting it completely.

I think is the sort of thing that async people use Futures or Promises or whatever else they are called in difference asyncio frameworks for. Wrapping a corutine in an awaitable objects which "returns" when the coroutine have been executed. But I'm basically just saying words here.

pwt: > But what about the question about actually getting the payload. It is somehow happening behind the scenes or am I missing it?

Yes, it's automatically handled in `events_asyncio` when the event notification is received: https://github.com/SoCo/SoCo/blob/f41b6e81b08ff6d615a1ae3a034c2dbb3be81cae/soco/events_asyncio.py#L140-L147

pwt: > I think a `sleep()` is needed to give other coroutines a chance to run. However, I'll experiment with shorter durations, and also with omitting it completely.

I think I understand what's happening now:
- The `asyncio.sleep()` is needed to introduce enough delay between `subscribe()` and `unsubscribe()` to allow an event to be generated by the Sonos player and received by SoCo, and it allows the event handler coroutine to run
- On my system I'm seeing an event response within ~25ms of the `subscribe()`
- Hence, I'm going to set the delay at 250ms; we can adjust this later (or make it configurable) if it turns out empirically to be too short

pwt:
Thanks @jjlawren and @KennethNielsen for taking the time to review this PR, and thanks also to @jjlawren for the insight that led to creating this workaround.

It still lacks support for `events_twisted` (I couldn't get this to work), but I suspect the Venn diagram intersection of users with very large Sonos systems and `events_twisted` users is rather small.

I'll merge this in the next day or two, and release SoCo v0.29.

PR #946, opened 2023-03-12
Remove Ubuntu version pin in check jobs

pwt:


pwt:
I guess we're going to have leave the Ubuntu 20.04 pin in place to retain Python 3.6 support.

PR #948, opened 2023-03-16
Add `sub_crossover` controls

jjlawren:
Adds support for subwoofer crossover controls. As of now, I believe this only applies to Amp devices.

Connect:Amp devices have a non-adjustable crossover at 80hz. I assume this will report it properly and that attempts to change will have no effect. Unfortunately I do not have a Connect:Amp (or Amp) to test with.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -236,6 +236,7 @@ class SoCo(_SocoSingletonBase):
         is_soundbar
         is_satellite
         has_satellites
+        sub_crossover
         sub_enabled
         sub_gain
         is_subwoofer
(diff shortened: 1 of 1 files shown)


pwt:
Briefly tested this on a variety of devices:
```
Living Room (Amp): Crossover Freq. = 60
Living Room (Amp): Set Crossover Freq. to 80 ... OK

Main Bedroom (Connect:Amp): Crossover Freq. = None
Main Bedroom (Connect:Amp): Set Crossover Freq. to 80 ... not supported

Diner (One): Crossover Freq. = None
Diner (One): Set Crossover Freq. to 80 ... not supported
```

Note that neither my Amp nor my Connect:Amp actually has a subwoofer attached.

Setting the Amp's crossover via the `SoCo.sub_crossover = frequency` sets it to **60Hz** regardless of what `frequency` is supplied, whereas setting it via the (S1) iOS app does change the crossover in the app and this change is also visible via the getter.

So, I'm not sure the setter is working; I'd expect it to be consistent with the iOS app.

>Connect:Amp devices have a non-adjustable crossover at 80hz. I assume this will report it properly and that attempts to change will have no effect. Unfortunately I do not have a Connect:Amp (or Amp) to test with.

I'm not sure why this is a concern, as there's an explicit test for an 'Amp' in both the getter and the setter?

jjlawren:
Hmm, another user implied the setter worked, so perhaps actually having a sub attached is needed for the system to accept the change.

As I don't have a sub myself, it's been a bit difficult to figure out how this is determined by the Sonos system. We've been using the channel layout to infer the existence of a subwoofer, but perhaps there's a better method. For example, this doesn't seem to show external 3rd party subs connected to an Amp.

> I'm not sure why this is a concern, as there's an explicit test for an 'Amp' in both the getter and the setter?

This comment was about there being a difference in behavior between the older `Connect:Amp` model vs the newer `Amp` devices. Only the newer ones support an adjustable crossover point.

pwt:
Let's release this into the wild to obtain additional empirical data.

PR #951, opened 2023-11-04
Add methods to get next alarm datetime

DaAwesomeP:
I have added methods to more easily determine:
- The next `datetime` for an `Alarm`
- The next `datetime` for an `Alarms` (i.e. a whole zone/room)

These methods are particularly useful for home automation scenarios.

~~This is still a draft because I have not tested it yet.~~

Maybe there is a better name for the `bypass_disabled` argument?

### DIFF
--- a/soco/alarms.py
+++ b/soco/alarms.py
@@ -185,7 +185,7 @@ class Alarms(_SocoSingletonBase):
             if not new_alarms.get(alarm_id):
                 self.alarms.pop(alarm_id)
 
-    def get_next_alarm_datetime(self, from_datetime=None, bypass_disabled=False):
+    def get_next_alarm_datetime(self, from_datetime=None, include_disabled=False):
         """Get the next alarm trigger datetime.
 
         Args:
@@ -194,7 +194,7 @@ class Alarms(_SocoSingletonBase):
                 exact time. Since alarms do not store timezone information,
                 the output timezone will match this input argument. Defaults
                 to `datetime.now()`.
-            bypass_disabled (bool, optional): If `True` then disabled alarms
+            include_disabled (bool, optional): If `True` then disabled alarms
                 will be included in searching for the next alarm. Defaults to
                 `False`.
 
@@ -208,7 +208,7 @@ class Alarms(_SocoSingletonBase):
         for alarm_id in self.alarms:
             this_alarm = self.alarms.get(alarm_id)
             this_next_datetime = this_alarm.get_next_alarm_datetime(
-                from_datetime, bypass_disabled
+                from_datetime, include_disabled
             )
             if next_alarm_datetime is None or this_next_datetime < this_alarm:
                 next_alarm_datetime = this_next_datetime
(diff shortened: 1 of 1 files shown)


pwt:
Thanks for the contribution.

> Maybe there is a better name for the bypass_disabled argument?

I'd be minded to invert this to be `include_disabled`, and to make the default `False`.

soco/alarms.py
```diff
@@ -179,6 +185,35 @@ def update(self, zone=None):
...
+                to `datetime().now()`.
+            bypass_disabled (bool, optional): If `True` then disabled alarms
+                will be included in searching for the next alarm. Defaults to
+                `False`.
+
+        Returns:
+            datetime: The next alarm trigger datetime or None if disabled
+        """
+        if from_datetime is None:
+            from_datetime = datetime().now()
```
pwt, line 205: This should be `datetime.now()`

soco/alarms.py
```diff
@@ -372,6 +407,71 @@ def alarm_id(self):
...
+                `False`.
+
+        Returns:
+            datetime: The next alarm trigger datetime or None if disabled
+        """
+        if not self.enabled and not bypass_disabled:
+            return None
+
+        if from_datetime is None:
+            from_datetime = datetime().now()
```
pwt, line 430: Replace with `datetime.now()`

pwt:
I've fixed the check jobs so that all checks on `master` are now passing. If you rebase, you'll pick up the fix.

You'll also need to address the following -- see code review comments.
```
************* Module soco.alarms
soco/alarms.py:205:28: E1120: No value for argument 'year' in constructor call (no-value-for-parameter)
soco/alarms.py:430:28: E1120: No value for argument 'year' in constructor call (no-value-for-parameter)
soco/alarms.py:470:35: E1101: Class 'datetime' has no 'timedelta' member (no-member)
```

soco/alarms.py
DaAwesomeP: Fixed!

soco/alarms.py
DaAwesomeP: Fixed!

soco/alarms.py
```diff
@@ -372,6 +407,71 @@ def alarm_id(self):
...
+
+        # Find first day
+        from_datetime_day = from_datetime.weekday()
+        offset_weekday = (from_datetime_day + offset) % 7
+        while offset_weekday not in recurrence_set:
+            offset += 1
+            offset_weekday = (from_datetime_day + offset) % 7
+
+        return datetime.combine(
+            from_datetime.date() + datetime.timedelta(days=offset),
```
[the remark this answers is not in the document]
DaAwesomeP: Fixed!

DaAwesomeP:
@pwt Thanks for the review! OK, I have fixed your notes (I did not see those errors for Python 3.12 before). I have also now actually tried this out and fixed some other bugs that I did not find before. I added a `zone_uid` filter to make it easier to for the next alarm by device zone.

pwt (changes requested):
This all looks great to me -- thanks.

I'd remove the `assert` statements (lines 453 and 461/462), unless there's a specific reason you think they should remain for the production release -- then I think we're all set to merge.

DaAwesomeP:
> I'd remove the `assert` statements (lines 453 and 461/462), unless there's a specific reason you think they should remain for the production release -- then I think we're all set to merge.

Ah, yeah sorry, all fixed! Thank you!

pwt (approved):

PR #955, opened 2023-12-23
Rework ZoneGroupState cache strategy

jjlawren:
A followup to the ZoneGroupState caching added in https://github.com/SoCo/SoCo/pull/913.

The last `ZoneGroupState` payload received for a Sonos system can be used indefinitely as long as a `ZoneGroupTopology` subscription remains active. With long-lived subscriptions, idle devices will have periods of time where no ZoneGroupState payload is published and the ZGS cache will expire long before the subscription needs to be renewed. This will require a poll to request a fresh payload and on large systems, this will fail (see https://github.com/SoCo/SoCo/pull/937).

This change ties the cache timeout from a default timeout of 60s since the last payload arrival to the lifetime of an active `ZoneGroupTopology` subscription.

### DIFF
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -453,6 +453,8 @@ class SubscriptionBase:
             # Autorenew just before expiry, say at 85% of self.timeout seconds
             interval = self.timeout * 85 / 100
             self._auto_renew_start(interval)
+            if service.service_type == "ZoneGroupTopology":
+                service.soco.zone_group_state.extend_cache(self.timeout)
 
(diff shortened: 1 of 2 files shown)


jjlawren:
I've marked this as a draft as this is a bit more complex if there are multiple `ZoneGroupTopology` subscriptions active at any given time. Extending the cache when a subscription is created would be fine, but we shouldn't unconditionally invalidate the cache when _a_ subscription is cancelled as there may be others still active which will continue delivering valid, identical payloads.

PR #956, opened 2023-12-26
Cache ZoneGroupState while subscriptions are active

jjlawren:
A followup to the ZoneGroupState caching added in https://github.com/SoCo/SoCo/pull/913.

The last `ZoneGroupState` payload received for a Sonos system can be used indefinitely as long as a `ZoneGroupTopology` subscription remains active. With long-lived subscriptions, idle devices will have periods of time where no ZoneGroupState payload is published and the ZGS cache (60s since the last "event") will usually expire long before the subscription needs to be renewed. After it expires, it requires a poll to request a fresh payload and on large systems this will fail (see https://github.com/SoCo/SoCo/pull/937).

Instead of extending the ZGS cache by 60s every time a subscription event is received, this keeps the cache alive while _any_ `ZoneGroupTopology` subscription remains active. This should make the library more usable on large systems as long as a subscription is maintained.

This replaces https://github.com/SoCo/SoCo/pull/955.

### DIFF
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -438,6 +438,7 @@ class SubscriptionBase:
                 self.timeout = int(timeout.lstrip("Second-"))
             self._timestamp = time.time()
             self.is_subscribed = True
+            service.soco.zone_group_state.add_subscription(self)
             log.debug(
                 "Subscribed to %s, sid: %s",
                 service.base_url + service.event_subscription_url,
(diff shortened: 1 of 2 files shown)


jjlawren:
The linked issue has a couple reports of users with large Sonos systems using this branch successfully. No regressions reported so far.

pwt:
> The linked issue has a couple reports of users with large Sonos systems using this branch successfully. No regressions reported so far.

Thanks @jjlawren. This change is now released in v0.30.2.

jjlawren:
Thanks!

Issue #964, opened 2024-07-13
XML Parser error on linein update


PeteRager:
A XML parsing error is occuring repeatedly for a Home Assistant user. I have been able to duplicate this issue in a SoCo test by loading the XML from a file and calling parse_event_xml

The malformed XML looks like this:   "res x-rincon-stream" is an invalid xml element name

I submitted PR #965 that has the test case and XML that was captured in debugging that causes the error

```
<DIDL-Lite xmlns:dc="http://purl.org/dc/elements/1.1/"
    xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/"
    xmlns:r="urn:schemas-rinconnetworks-com:metadata-1-0/"
    xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/">
    <item id="playback" parentID="0" restricted="false">
        <dc:title>loadLineIn</dc:title>
        <upnp:class>object.item.audioItem.linein</upnp:class>
        <res x-rincon-stream>x-rincon-stream:RINCON_949F3E5352EC01400</res></item></DIDL-Lite>
```

https://github.com/home-assistant/core/issues/121638

```
2024-07-09 19:29:22.784 ERROR (MainThread) [aiohttp.server] Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/aiohttp/web_protocol.py", line 452, in _handle_request
    resp = await request_handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/aiohttp/web_app.py", line 543, in _handle
    resp = await handler(request)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/soco/events_asyncio.py", line 138, in notify
    variables = parse_event_xml(content)
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/soco/events_base.py", line 104, in parse_event_xml
    value = from_didl_string(value)[0]
            ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/soco/data_structures_entry.py", line 31, in from_didl_string
    root = XML.fromstring(string.encode("utf-8"))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #965, opened 2024-07-13
Fix - audio linein parsing error

PeteRager:
Fix for #964 

In this issue; Sonos is sending a new XML element called "object.item.audioItem.linein" and the XML for this element is causing a parsing error.

Two changes:
- Use XML parser with recovery=True to allow parsing this new event
- Add the new Dldl class to represent the audio line in.


PeteRager:
Updated code to ignore the XML error. This led to discovering a new DIDL type that was being sent object.item.audioItem.linein; that needed to be defined,

pwt:
Hi Pete: thanks for working on this. Let me know when it's ready to review.

Regarding the check job failures:

1. If you rebase against `master`, that will suppress the Pylint `possibly-used-before-assignment` warnings.
2. That leaves the following to be addressed:
```
************* Module soco.data_structures_entry
soco/data_structures_entry.py:32:13: I1101: Module 'lxml.etree' has no 'XMLParser' member, but source is unavailable. Consider adding this module to extension-pkg-allow-list if you want to perform analysis based on run-time introspection of living objects. (c-extension-no-member)
soco/data_structures_entry.py:33:11: I1101: Module 'lxml.etree' has no 'fromstring' member, but source is unavailable. Consider adding this module to extension-pkg-allow-list if you want to perform analysis based on run-time introspection of living objects. (c-extension-no-member)
soco/data_structures_entry.py:12:0: C0411: third party import "lxml.etree" should be placed before local imports "data_structures.didl_class_to_soco_class", "exceptions.DIDLMetadataError", "xml.ns_tag" (wrong-import-order)
``` 

Cheers,
Peter

PeteRager:
I've fixed the linter / formatting issues, and extended the test, ready for review.

I do have concerns on the changes. While the change to using the parser with recover=True gets us through this specific issue. It does prevent logging exceptions when encountering invalid XML which could make it harder to diagnose future issues.

We could try to parse it with recover=False; if parsing fails - then log a DEBUG message and then parse it with recover=True - this way we'd get the details when debug is turned on; with the benefit of more runtime resilience to Sonos changes.  Could be I'm also overthinking it.

pwt:
Hi @PeteRager : thanks for your work on this; it's now merged and included in SoCo patch release v0.30.5.

Apologies for the long delay.

PR #967, opened 2024-11-08
Add Arc Ultra support

jjlawren:
The new Arc Ultra devices identify themselves as "Sonos Arc Ultra". This adds support for the soundbar-related controls.


pwt:
Thanks for adding this. It's now released in SoCo v0.30.6.

PR #972, opened 2025-01-25
Parse hyphens in track metadata

PeteRager:
When a radio station provides track metadata using TYPE=SNG in the stream_content and the album name contains a hyphen the track title, artist and album would not be parsed correctly.

The fix is to check for TYPE=SNG first; rather than the presence of a hyphen.

Reference HA issue:  https://github.com/home-assistant/core/issues/136474

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -1961,6 +1961,8 @@ class SoCo(_SocoSingletonBase):
                 or ""
             )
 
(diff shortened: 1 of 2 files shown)


amelchio (approved):
That makes sense.

pwt:
LGTM -- thanks @PeteRager and @amelchio. I'll make a patch release.

PR #973, opened 2025-02-18
ContentDirectory is missing some favorites.

PeteRager:
Browsing favorites with the ContentDirectory was missing Audible favorites. The fix is for the ContentDirectory to pass user-agent in the header.  The speaker users this header to determine the API version of the client. And if the client is an older version (or in our case we were not specifying the header) then it does not return the Audible favorites.

https://github.com/home-assistant/core/issues/136840

Similar to this existing piece of code:

https://github.com/SoCo/SoCo/blob/a954565f3f6eb6addfeeab3f115fd9b42d62b91f/soco/music_services/music_service.py#L78-L93


mdwhitten:
@PeteRager I tested this fix. Calling `mySonos.music_library.get_sonos_favorites()` now returns the audiobooks as expected. Many thanks!

pwt:
LGTM. Released in SoCo v0.30.9. Thanks for the contribution and the testing.

PeteRager:
Thank you for merging and creating release.

PR #976, opened 2025-03-02
Adjust to allow both IPv4 and IPv6 addresses

CloCkWeRX:
https://github.com/home-assistant/core/issues/139617

- Adds a net_addr dependency, I haven't looked into if its actually necessary or if its better to just do the try-catch parsing approach/make wrappers
- Adjust discovery to work on IPv6, adding a simulated test case
- Does not address other areas in the code
- Swaps to inet_pton() in various places

Not tested with a real device.

I have a busy work week; so feel free to take over this PR/branch if there's other areas that should be addressed and you don't feel like waiting ~6-7 days.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -5,6 +5,8 @@ the main entry to the SoCo functionality
 import datetime
 import logging
 import re
+from netaddr.strategy.ipv6 import valid_str as valid_ipv6
+from netaddr.strategy.ipv4 import valid_str as valid_ipv4
 from functools import wraps
 import urllib.parse
 from xml.sax.saxutils import escape
@@ -325,6 +327,9 @@ class SoCo(_SocoSingletonBase):
         # Note: Creation of a SoCo instance should be as cheap and quick as
         # possible. Do not make any network calls here
         super().__init__()
+        if valid_ipv4(ip_address) != True:
+            if valid_ipv6(ip_address) != True:
+                raise ValueError("Not a valid IP address string")
 
         #: The speaker's ip address
         self.ip_address = ip_address
--- a/soco/core.py
+++ b/soco/core.py
@@ -5,7 +5,6 @@ the main entry to the SoCo functionality
 import datetime
 import logging
 import re
-import socket
 from functools import wraps
 import urllib.parse
 from xml.sax.saxutils import escape
--- a/soco/core.py
+++ b/soco/core.py
@@ -5,13 +5,13 @@ the main entry to the SoCo functionality
 import datetime
 import logging
 import re
-from netaddr.strategy.ipv6 import valid_str as valid_ipv6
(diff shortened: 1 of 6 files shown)


pwt:
Not seeing any files changed in this commit.

CloCkWeRX:
Ugh, github UI :/ two secs

pwt:
Has this been tested with any IPv6 Sonos systems? I'm wondering if the IPv4 assumption is widespread, beyond just the check. The discovery code, for example.

PeteRager:
Apparently this is happening during HA auto discovery.  We need a look at the packets coming in as it's possible it's going to be auto discovered with a v4 and a v6 address.  Another solution is to filter out v6 discovery packets.  

@CloCkWeRX we should create a test case in HA that generates an auto discovery ipv6 address that would allow some level of trace through to happen without the device in hand.  Likewise we'd need a similar test case here.

soco/core.py
```diff
@@ -326,12 +325,7 @@ def __init__(self, ip_address):
         # Note: Creation of a SoCo instance should be as cheap and quick as
         # possible. Do not make any network calls here
         super().__init__()
-        # Check if ip_address is a valid IPv4 representation.
-        # Sonos does not (yet) support IPv6
-        try:
-            socket.inet_aton(ip_address)
-        except OSError as error:
-            raise ValueError("Not a valid IP address string") from error
+
```
PeteRager, line 328: I believe ip address is a string at this point.  So we should validate it as either an ipv4 or ipv6. Removing the check will just allow anything to be passed in which isn't a good solution.

PeteRager: Sounds good, there are regression errors in the tests reported in CI that need to be fixed (and updated to verify a v6 address changes)

CloCkWeRX:
>  I'm wondering if the IPv4 assumption is widespread, beyond just the check.

So yes, I suspect that's likely. Unfortunately, I'm working off the backtrace in https://github.com/home-assistant/core/issues/139617 rather than having a physical device so its harder to build up a reasonable set of realistic IPv6 test cases.

Did what I could by following the existing patterns.

soco/discovery.py
```diff
@@ -661,32 +671,40 @@ def _find_ipv4_networks(min_netmask):
...
+def _find_ip_addresses():
+    """Discover and return all the host's IPv4 and IPv6 addresses.
 
     Helper function to return a set of IPv4 addresses associated
     with the network interfaces of this host. Loopback and link
     local addresses are excluded.
 
     Returns:
-        set: A set of IPv4 addresses (dotted decimal strings). Empty
+        set: A set of IPv4, IPv6 addresses (dotted decimal strings). Empty
```
CloCkWeRX, line 682:
```suggestion
        set: A set of IPv4, IPv6 addresses. Empty
```

PR #996, opened 2026-04-01
Alarms for powered off speakers may be discarded

PeteRager:
During the initial fetch of the alarm list; if a speaker with an alarm is not registered in Soco then the alarm will be discarded. Subsequent calls to re-fetch the alarm list when the speaker is registered do not re-add the alarm because the alarm list version has not changed. The fix is to store these alarms on a separate list and allow them to be recovered by the application when the speaker becomes available.

https://github.com/home-assistant/core/issues/166446

The issue can be reproduced using two speakers with one alarm on each speaker. 

Power both speakers off.  
Power one on.
Connect Soco to that speaker
Get the alarm list (it will have one alarm)
Power the other speaker on
Connect Soco to that speaker
Get the alarm list (it will still have one alarm)

### DIFF
--- a/.github/workflows/soco-check-jobs.yaml
+++ b/.github/workflows/soco-check-jobs.yaml
@@ -43,9 +43,7 @@ jobs:
           pytest --cov-config .coveragerc --cov soco .
       - name: Test formatting with Black
         if: >-
-          ${{ matrix.python-version == '3.8' ||
-              matrix.python-version == '3.9' ||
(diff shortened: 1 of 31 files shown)


PeteRager:
I'm not sure what is going on with the black failure.

PeteRager:
Thank you @pwt. 

Claude's changes look great and fix an edge case if the alarm was edited.  I retested on a Sonos system and it still works as expected.

pwt:
Now released in v0.30.15.

PR #997, opened 2026-04-01
Pin black version

PeteRager:
The failures in #996 seems to be cause by black using python 3.14 formatting.

Pinning black to 24.8.0 provides compatibility across the python version 3.8-3.14.
Pin the black formatting to 3.8


pwt:
Thanks for PR, @PeteRager.

I'd prefer not to pin Black, so I've adjusted that.

Also:
- Only run on Python 3.10+, the earliest version supported by the current version of Black
- Target Python 3.7 for formatting, the earliest version supported by Black.

PR #999, opened 2026-05-17
Add soco_reset function for singleton lifecycle management

PeteRager:
## Summary

The API utilizes singletons to return objects to the user — for example `SoCo` and `Alarm` objects. However, there is no mechanism to clear these singletons. This PR adds the function `soco_reset`  to the top-level `soco` package to provide API entry points for resetting this state.

```python
from soco import soco_reset

soco_reset()   # call before using the API
# ... use the api
```

## Motivation
**Home Assistant integration reloads** - Existing singletons are not recreated on reload, so users often have to fully restart Home Assistant to reach a clean state. With this change, Home Assistant can call soco_initialize() during reload to clear the singletons, making reload and restart behavior consistent.

**Test isolation** - Singleton state leaks across test cases, causing interference and brittleness. This PR also applies the fix to SoCo's own test suite via an autouse fixture in conftest.py that calls soco_reset() before after each test. Also fixes tests that were dependent on singleton leakage.

### DIFF
--- a/examples/play_local_files/play_local_files.py
+++ b/examples/play_local_files/play_local_files.py
@@ -39,7 +39,6 @@ except ImportError:
 
     print("Running as python 2")
 
-from soco import soco_initialize, soco_shutdown
 from soco.discovery import by_name, discover
 
 
@@ -120,8 +119,6 @@ def parse_args():
 
 
 def main():
-    soco_initialize()
-
     # Settings
     args = parse_args()
     print(
@@ -170,8 +167,6 @@ def main():
         time.sleep(10**8)
     except KeyboardInterrupt:
         server.stop()
-    finally:
-        soco_shutdown()
 
 
 main()
--- a/examples/play_local_files/play_local_files.py
+++ b/examples/play_local_files/play_local_files.py
@@ -39,6 +39,7 @@ except ImportError:
 
     print("Running as python 2")
 
+from soco import soco_initialize, soco_shutdown
 from soco.discovery import by_name, discover
 
 
@@ -119,6 +120,8 @@ def parse_args():
 
 
 def main():
+    soco_initialize()
+
     # Settings
     args = parse_args()
     print(
@@ -167,6 +170,8 @@ def main():
         time.sleep(10**8)
     except KeyboardInterrupt:
         server.stop()
+    finally:
+        soco_shutdown()
 
 
 main()
--- a/examples/plugins/socoplugins.py
+++ b/examples/plugins/socoplugins.py
@@ -40,8 +40,6 @@ def main():
     # do something with your plugin
     myplugin.music_plugin_stop()
 
-    soco_shutdown()
-
 
 if __name__ == "__main__":
     main()
--- a/examples/plugins/socoplugins.py
+++ b/examples/plugins/socoplugins.py
@@ -5,11 +5,13 @@
 
 import time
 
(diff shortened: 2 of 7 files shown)


pwt:
Hi @PeteRager 

I ran this through Claude Code and I'll include the unedited comments below. Have a quick check to see if they prompt any changes.

## Summary

Adds `soco_initialize()` and `soco_shutdown()` to the public API, both of which currently just call `_ArgsSingleton._instances.clear()`. Adds an autouse test fixture for isolation, fixes two tests that relied on singleton leakage, and updates two example scripts.

---

## Design concerns

1. **Both functions do the same thing.** Right now `soco_initialize()` and `soco_shutdown()` are identical — both just clear `_instances`. The PR justifies this by saying they'll diverge in future releases, but this is speculative. The resulting API is confusing: why call `soco_initialize()` at startup if it just destroys existing singletons? A single `soco_reset()` or `soco_clear_instances()` would be more honest about what it actually does today.

2. **`_ArgsSingleton._instances` is a private attribute of a metaclass.** Exposing its management via public module-level functions creates a leaky abstraction. A cleaner approach would be a `reset()` classmethod on `_SocoSingletonBase` (or a dedicated internal helper), so the public functions delegate to a stable internal contract rather than reaching directly into a metaclass private dict.

3. **Examples should not be required to call lifecycle functions.** Adding `soco_initialize()`/`soco_shutdown()` to example scripts implies they're mandatory. If they're just optional cleanup, the examples mislead users. If they *are* mandatory, that's a significant API contract that needs to be documented clearly and would break all existing user code that doesn't call them.

---

## Positives

- The autouse fixture in `conftest.py` is the right fix for test isolation — this is clearly a genuine problem.
- The two test fixes (`moco._uid = "RINCON_000XXX1400"`) correctly address tests that were relying on side-effect state rather than explicit setup.
- Since `_ArgsSingleton` is the metaclass for both `SoCo` and `AlarmList`, clearing `_instances` correctly resets all singleton types — not just `SoCo`.
- New tests in `test_singleton.py` directly verify the behavior.

---

## Smaller issues

- **Thread safety:** `_instances.clear()` while another thread holds a reference to a singleton instance is unsafe. Worth at least a note in the docstring.
- **`soco_shutdown()` in example `play_local_files.py`** is inside a `finally` block that only triggers after `server.stop()`, but `soco_initialize()` is called before the server starts. If the server never starts (e.g. an exception in `parse_args()`), `soco_shutdown()` still fires — harmless here, but the pattern is inconsistent.

---

## Bottom line

The test isolation benefit is real and valuable, but the public API shape needs more thought. Suggest either: (a) keep just the autouse fixture and don't expose public lifecycle functions yet, or (b) if lifecycle functions are the goal, rename to something accurate (`soco_reset()`), keep it one function, and don't add it to examples until it's clearly documented as required.

PeteRager:
That's good feedback, I'll work on it.

Funny how Claude wrote the code for me, recommended this pattern

PeteRager:
Updated to use soco_reset() pattern, much simpler.

pwt:
Thanks @PeteRager. Just added a couple of minor tidy-ups before merge. I'll release this shortly.

pwt:
Released in SoCo v0.31.1.

PeteRager:
Thanks for the fixes and releasing

PR #1001, opened 2026-05-20
events_asyncio: defer EventListener teardown so resubscribes reuse the running server

schuettc:
## Summary

Fixes a lifecycle race in `soco.events_asyncio.EventListener` that, on every unsubscribe → resubscribe cycle, runs `EventListener.async_stop()` as a fire-and-forget task against a socket the next subscription has just replaced. Symptoms in the wild:

- `ValueError: Invalid file descriptor: -1` raised from inside `aiohttp.web_runner.SockSite.stop()` → `_loop._stop_serving(sock)` → `sock.fileno()` against an already-closed socket.
- The exception escapes as `Task exception was never retrieved` because `stop_listening()` schedules `async_stop()` via `asyncio.ensure_future` without retaining a reference or attaching a done callback.
- `async_stop` partially completes against the *new* listener's resources, intermittently orphaning the `runner` / `session` / `sock` pointers belonging to the fresh subscription.

Reproduced reliably on a long-lived asyncio consumer (Raspberry Pi orchestrator subscribing to `AVTransport` on a Sonos zone, with a 10-minute no-events watchdog that periodically resubscribes). Every watchdog resubscribe triggered the traceback. After this PR, 47 hours of continuous run with **122 clean resubscribe cycles, zero `Invalid file descriptor` errors, zero `Task exception was never retrieved` log lines.**

Original reproduction traceback:

```
ERROR asyncio Task exception was never retrieved
future: <Task finished name='Task-6060' coro=<EventListener.async_stop() done,
 defined at .../soco/events_asyncio.py:278>
 exception=ValueError('Invalid file descriptor: -1')>
  File ".../soco/events_asyncio.py", line 282, in async_stop
    await self.site.stop()
  File ".../aiohttp/web_runner.py", line 77, in stop
    self._server.close()
  File ".../asyncio/base_events.py", line 342, in close
    self._loop._stop_serving(sock)
  ...
  File ".../selectors.py", line 42, in _fileobj_to_fd
    raise ValueError("Invalid file descriptor: {}".format(fd))
ValueError: Invalid file descriptor: -1
```

Environment: soco 0.31.0, Python 3.13.7 (Pi OS Trixie), aiohttp 3.13.x.

## Why it happens

1. The last subscription's `unsubscribe()` calls `event_listener.stop()` → `stop_listening()`.
2. `stop_listening()` schedules `async_stop()` with `asyncio.ensure_future(self.async_stop())` — fire and forget.
3. The caller immediately resubscribes. `async_start()` opens a **new** socket and `web.SockSite` and assigns them to `self.sock` / `self.site` on the singleton listener.
4. The previously-scheduled `async_stop` finally runs. `await self.site.stop()` is now operating on the new site whose `_server.close()` calls `_loop._stop_serving(sock)` on a socket whose `fileno() == -1` because a closer path already closed it → `ValueError`.

## Fix

Three commits, logically separated for review:

### `65318ca` — Make `async_stop` tolerant of socket-close races

Wraps `site.stop()` in a `try/except (ValueError, OSError)` so the rest of the cleanup (runner, session, sock) still runs to completion. Attaches a done-callback on the task spawned by `stop_listening()` to consume any exception (no more `Task exception was never retrieved`). This is the defensive layer that handles edge cases the structural fix below doesn't anticipate.

### `c8bb455` — Defer `EventListener` teardown so resubscribes reuse the running server

The structural fix. `stop_listening()` now schedules teardown after a short grace window (`_stop_grace_seconds`, default 5.0s). `async_start()`, before checking `is_running`, cancels any pending deferred-stop task — and if the listener's runtime resources are still intact, simply restores `is_running = True` and returns without rebuilding anything. The HTTP server is reused across the unsubscribe → subscribe cycle.

`async_stop()` is also reworked to be race-safe: acquires a stop-lock, snapshots `site` / `runner` / `session` / `sock` into locals and clears the instance attributes *inside* the lock, then closes the snapshots *outside* the lock. A concurrent `async_start` never observes a half-torn-down listener, and a second overlapping `async_stop` sees cleared attrs and returns early (idempotent).

### `457c1a7` — Live-network verification example

`examples/events_asyncio_verify_listener_fix.py` exercises the deferred-stop path against a real Sonos zone (using `renderingControl` to avoid colliding with any consumer's `avTransport` subscription) and asserts:

- After `unsubscribe()`, `stop_listening` schedules a deferred-stop task; resources stay alive during the grace window.
- A resubscribe inside the grace window cancels the deferred stop — `site` and `sock` object identity preserved, ``Event Listener resumed (deferred stop cancelled)`` debug log fires.
- After a final unsubscribe with no resubscribe, the deferred stop runs after the grace window expires; resources cleared.

## Tests

Six new tests in `tests/test_events_asyncio.py`:

- `test_async_stop_tolerates_closed_socket` — covers `65318ca`.
- `test_stop_listening_swallows_task_exception` — covers `65318ca`.
- `test_async_start_cancels_pending_deferred_stop_and_resumes` — fast-path: resubscribe inside grace reuses the listener.
- `test_deferred_stop_runs_when_grace_expires_and_count_zero` — slow path: real shutdown after grace.
- `test_deferred_stop_aborts_when_subscription_appears_in_grace` — subscription arriving during grace aborts teardown.
- `test_async_stop_idempotent_under_parallel_calls` — overlapping `async_stop()` calls.

Full suite: **251 passed, 67 skipped** (integration tests requiring `--ip`). No regressions.

## Behavior compatibility

- Existing callers see no API change. `event_listener.async_stop()` still tears down immediately when called directly (idempotent now). `stop_listening()` still results in teardown — just deferred by 5s, which a final-shutdown path will naturally wait through.
- The grace window is exposed as `event_listener._stop_grace_seconds` so consumers (or tests) can shorten/lengthen it.
- For consumers that don't resubscribe (single subscribe → process exit), behavior is identical apart from a 5s delay on `async_stop` running. If that matters, callers should `await event_listener.async_stop()` directly during shutdown.

## Production evidence

Deployed via a git pin on a Raspberry Pi orchestrator (Sonos UPnP `AVTransport` subscription, watchdog-based resubscribe every ~10 min on event silence). Continuous run since 2026-05-21:

- 47 h uptime, **122 watchdog-driven resubscribe cycles**, all clean.
- Zero new tracebacks of any kind.
- Original user-facing symptom (kiosk failing to update for ~10 min after a state change on the Sonos zone) has not recurred.

## How to verify locally

After installing this branch, against a Sonos zone on your LAN:

```
python examples/events_asyncio_verify_listener_fix.py --zone "Living Room"
```

Exits 0 on PASS. Useful both as a regression guard and as a worked example of asyncio subscription lifecycle.

### DIFF
--- a/examples/events_asyncio_verify_listener_fix.py
+++ b/examples/events_asyncio_verify_listener_fix.py
@@ -1,0 +1,164 @@
+"""Live-network verification of the events_asyncio EventListener fix.
+
+Exercises the deferred-stop refcounting path against a real Sonos zone
+and asserts the structural invariants that the fix guarantees:
+
+  1. After ``subscription.unsubscribe()``, ``stop_listening`` schedules a
+     deferred stop. The listener's runtime resources (``site``, ``sock``,
+     ``runner``, ``session``) remain alive during the grace window.
+
+  2. A resubscribe within the grace window cancels the deferred stop and
+     resumes the listener — the site/sock object identity is preserved
+     and the "Event Listener resumed (deferred stop cancelled)" debug log
+     fires. No teardown/rebuild churn.
+
+  3. After a final unsubscribe with no resubscribe, the deferred stop
+     runs once the grace window expires; listener resources are released
+     and ``is_running`` returns to ``False``.
+
+The test uses ``renderingControl`` rather than ``avTransport`` so it
+does not interfere with consumer applications that may have an
+``avTransport`` subscription active against the same zone (e.g. a
+"now playing" orchestrator).
+
+Pass a zone name (any zone on your local network) via ``--zone`` or set
+``SOCO_ZONE``. Discovery uses SSDP, so the host must be on the same
+LAN/VLAN as the Sonos players.
+
+Usage:
+
+    python examples/events_asyncio_verify_listener_fix.py --zone "Office"
+    SOCO_ZONE="Living Room" python examples/events_asyncio_verify_listener_fix.py
+
+Exit code 0 on PASS, non-zero on FAIL.
+"""
+
+import argparse
+import asyncio
+import logging
+import os
+import sys
+
+import soco
+from soco import config as soco_config
+from soco import events_asyncio
+
+# Must be set BEFORE any subscription is created so SoCo uses the asyncio
+# event module rather than the threaded one.
+soco_config.EVENTS_MODULE = events_asyncio
+
+
+# Keep the test fast — production default is 5.0s.
+SHORT_GRACE = 0.5
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Verify events_asyncio EventListener fix against a live Sonos zone."
+    )
+    parser.add_argument(
+        "--zone",
+        default=os.environ.get("SOCO_ZONE"),
+        help="Sonos zone name to bind against (default: $SOCO_ZONE).",
+    )
+    parser.add_argument(
+        "--discover-timeout",
+        type=float,
+        default=3.0,
+        help="SSDP discovery timeout in seconds (default: 3.0).",
+    )
+    return parser.parse_args()
+
+
+async def main(zone_name: str | None, discover_timeout: float) -> int:
+    if not zone_name:
+        print("ERROR: --zone (or $SOCO_ZONE) is required", file=sys.stderr)
+        return 2
+
+    logging.basicConfig(
+        level=logging.DEBUG,
+        format="%(asctime)s %(levelname)s %(name)s %(message)s",
+    )
+    # Silence aiohttp's chatter; keep soco at DEBUG so the
+    # "resumed (deferred stop cancelled)" line is visible.
+    logging.getLogger("aiohttp").setLevel(logging.WARNING)
+    logging.getLogger("urllib3").setLevel(logging.WARNING)
+
+    # Shorten the grace window so the test runs quickly.
+    events_asyncio.event_listener._stop_grace_seconds = SHORT_GRACE
+
+    # Capture log records emitted by soco.events_asyncio so we can assert
+    # on the structural log lines the patch emits.
+    captured: list[str] = []
+
+    class _RecordCapture(logging.Handler):
+        def emit(self, record):
+            captured.append(record.getMessage())
+
+    logging.getLogger("soco.events_asyncio").addHandler(_RecordCapture())
+
+    zones = soco.discover(timeout=discover_timeout) or set()
+    zone = next((z for z in zones if z.player_name == zone_name), None)
+    if zone is None:
+        print(
+            f"ERROR: zone {zone_name!r} not found via SSDP. "
+            f"Visible zones: {[z.player_name for z in zones] or 'none'}",
+            file=sys.stderr,
+        )
+        return 1
+    print(f"Using zone: {zone.player_name} @ {zone.ip_address}")
+
+    listener = events_asyncio.event_listener
+
+    print("\n--- Phase 1: initial subscribe ---")
+    sub = await zone.renderingControl.subscribe(auto_renew=False)
+    sub.callback = lambda ev: None
+    site1 = listener.site
+    sock1 = listener.sock
+    print(f"sid={sub.sid} site={id(site1):x} sock={id(sock1):x}")
+    assert listener.is_running, "listener should be running after subscribe"
+    assert site1 is not None and sock1 is not None
+
+    print("\n--- Phase 2: unsubscribe + immediate resubscribe (inside grace) ---")
+    await sub.unsubscribe()
+    pending = listener._stop_grace_task
+    print(
+        f"deferred task: {pending} "
+        f"done={pending.done() if pending else 'n/a'}"
+    )
+    assert pending is not None and not pending.done(), (
+        "stop_listening should have scheduled a deferred stop"
+    )
+    # Resources must still be alive during grace.
+    assert listener.site is site1, "site should not be replaced during grace"
+    assert listener.sock is sock1, "sock should not be replaced during grace"
+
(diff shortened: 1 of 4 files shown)


pwt:
Hi @schuettc : thanks for contributing to SoCo!

1. Please could you take a look at the missing import flagged in the check jobs.
2. I used Claude Code to review the PR, and I've included its review below. A lot of text, but just a couple of minor suggestions. Could you take a look at these, please?

Thanks,
Peter

# PR #1001 Review — events_asyncio: defer EventListener teardown so resubscribes reuse the running server

## Summary

Fixes a real, well-documented race in `EventListener` where `stop_listening()` fires `async_stop()` as a fire-and-forget task, and a racing resubscribe causes the deferred stop to tear down the new subscription's resources. Fix is in three layers: defensive exception handling in `async_stop()`, a 5-second deferred-stop grace window in `stop_listening()`, and cancellation of the pending stop in `async_start()` on resubscribe.

---

## Positives

- Root cause analysis is correct and the reproduction is solid (47 h / 122 cycles on real hardware is compelling evidence).
- The snapshot-then-clear pattern in `async_stop()` (capture locals inside the lock, close them outside) is the right approach to idempotency and concurrency.
- Done-callback on the `ensure_future` task correctly eliminates `Task exception was never retrieved`.
- Six focused unit tests cover all new paths: closed-socket tolerance, exception swallowing, fast-path resume, grace-window expiry, mid-grace subscription appearance, and parallel `async_stop` calls.
- The verification example (`events_asyncio_verify_listener_fix.py`) is well-written and doubles as a regression guard against a real zone.

---

## Concerns

1. **5-second shutdown delay for non-resubscribing callers.** Any consumer that does a single subscribe → unsubscribe → process exit now waits 5 s for the deferred stop to fire before resources are released. The PR mentions this only in passing ("callers should `await event_listener.async_stop()` directly during shutdown"). This is a meaningful behavior change and should be documented in the changelog and in `async_stop()`'s docstring, not just the PR description.

2. **Test isolation — module-level state mutation without full cleanup.** `test_stop_listening_swallows_task_exception`, `test_deferred_stop_runs_when_grace_expires_and_count_zero`, and `test_deferred_stop_aborts_when_subscription_appears_in_grace` all write directly to `events_asyncio.subscriptions_map.subscriptions` and `subscriptions_map._pending`. Only the last test cleans up (`subscriptions.clear()`). If any of these tests fails mid-run, state leaks into subsequent tests. Each test should save and restore (or clear) those module-level globals in a `try/finally` or fixture teardown.

3. **`pytest-asyncio` dependency.** The new tests use `@pytest.mark.asyncio`. Confirm `pytest-asyncio` is already listed as a test dependency in `setup.cfg` / `pyproject.toml` — it's not obvious from the diff.

4. **`_stop_grace_seconds` becomes an implicit API.** It's single-underscore, directly mutated in the example script and tests. That's fine, but worth a brief note in the docstring so consumers know it's intentional and what the units are.

5. **`stop_listening()` resets the grace clock on each call.** If called twice in quick succession (unlikely but possible), the second call cancels the first timer and starts a fresh 5-second window. In practice harmless, but worth being aware of.

---

## Smaller issues

- The `stop_lock` is initialized lazily inside `async_stop()` (same pattern as `start_lock`). Fine for asyncio's single-threaded model, but inconsistent with moving toward explicit initialization in `__init__`.
- The verification script accesses `listener._stop_grace_task` directly to assert internal state. Fine for a verification tool, but it's testing implementation details — fragile if the attribute is renamed.

---

## Bottom line

High-quality fix for a genuine bug. Ready to merge once: (a) the 5-second shutdown delay is documented in the changelog/docstring, (b) test teardown of `subscriptions_map` state is made robust, and (c) `pytest-asyncio` dependency is confirmed.

schuettc:
Thanks @pwt — addressed in 076b45d.

**CI green now.** Root cause was the missing `aiohttp` and `pytest-asyncio` in `requirements-dev.txt` (which also covers your concern about `pytest-asyncio` not being declared). Belt-and-suspenders: also added `pytest.importorskip("aiohttp")` at the top of the test module so a consumer running pytest on a system without the optional asyncio stack gets a clean skip rather than a collection error.

**Test isolation.** Added a `restore_subscriptions_map` fixture that snapshots `subscriptions_map.subscriptions` + `_pending` before each test that needs to mutate them, and restores in a `try/finally` regardless of test outcome. The three affected tests now opt in via parameter; the ad-hoc `clear()` calls and the trailing cleanup are gone.

**Shutdown-delay docs.** Extended docstrings on `__init__`, `stop_listening`, and `async_stop` to cover (a) the 5 s deferred-teardown semantics, (b) the prompt-shutdown escape hatch (`await event_listener.async_stop()` directly), and (c) `_stop_grace_seconds` as a tunable. Noted SoCo's per-version release notes seem to live as GitHub release-page links (the `doc/releases/0.31.0.rst` file is a one-line pointer) so I left that file alone — happy to add a `0.32.rst` stub if you'd like.

**Minor items.** The `_stop_grace_seconds` knob now has docstring coverage (issue 4). Behavior on consecutive `stop_listening()` calls (issue 5) — the timer-reset is intentional and now reflected in the docstring. The `stop_lock` lazy init (smaller issue 1) is deliberately consistent with the existing `start_lock` pattern in the same class; making one explicit-init and leaving the other lazy felt worse than the consistency, but I'll change it if you'd prefer. The verification script accessing `_stop_grace_task` (smaller issue 2) is by design — it's a structural-invariant test.

Re-running `pytest tests/` locally: 251 passed, 67 skipped (integration), 0 failures.

pwt:
Hi @schuettc : With a couple of minor tweaks this is merged now and will be included in the next patch release. This should happen shortly -- I'm just waiting on #999.

Thanks again for the contribution!

pwt:
Released in SoCo v0.31.1.

schuettc:
Awesome!  Thanks for the great work here.  It was super useful for what I was working on.  Will be releasing soon.

PR #1003, opened 2026-07-02
Relax strict checking for missing parentID in DIDL elements

relevitt:
### Description
Some upstream music services and specific Sonos metadata streams (such as `enqueued_transport_uri_meta_data`) may omit the required UPnP `parentID` attribute. Currently, SoCo raises a strict `DIDLMetadataError` when this attribute is missing. In the case of an Event , the relevant field is then populated with an `EventParseException`.

### Fix
This change relaxes that strict constraint by defaulting `parent_id` to an empty string (`""`) if it is absent. This prevents background exception and allows downstream applications to access the data.

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -576,7 +576,8 @@ class DidlObject(metaclass=DidlMetaClass):
         item_id = really_unicode(item_id)
         parent_id = element.get("parentID", None)
         if parent_id is None:
-            raise DIDLMetadataError("Missing parentID attribute")
+            # Relax strict checking for non-compliant DIDL metadata streams
+            parent_id = ""
         parent_id = really_unicode(parent_id)
 
         # CAUTION: This implementation deviates from the spec.


pwt (commented):
Hi @relevitt . Thanks for the contribution, and sorry for the delay (due to work & travel).

Below is Claude's review -- please take a look and let me know what you think.

---

Thanks for this — the direction is sound, and the case is a real one. CI is green across 3.8–3.13. I have two things I'd like to see addressed before merge, plus one question.

## Correctness

The choice of `""` is right, and relaxing the check here is consistent with existing practice:

- `""` already has precedent as the in-library "no parent" value — `core.py:2363` and `music_library.py:449` both construct `DidlObject(..., parent_id="")` directly, so this doesn't introduce a novel state.
- The very next block in the same function does exactly this kind of relaxation for `restricted` and `title`, under a `# CAUTION: This implementation deviates from the spec` comment naming Spotify Direct.
- Nothing asserted the old raise, which is consistent with CI passing.

One small correction to the description: the current behaviour isn't an unhandled exception. `events_base.py:103` already wraps this in `try/except SoCoException`, logs at debug, and substitutes a `SoCoFault`. So the change isn't preventing an escape — it's converting a deliberately-degraded `SoCoFault` into usable data. Still worthwhile (any attribute access on `SoCoFault` raises, so the field is effectively dead today), just a different thing from a crash fix.

## Main suggestion: put this in `data_structure_quirks.py`

That module exists for exactly this, and its docstring anticipates this change:

> A quirk, in this context, means that a specific music service does not follow a specific part of the DIDL-Lite specification. **In order not to clutter the primary implementation of DIDL-Lite for SoCo (in `soco.data_structures`) up with all these service specific exception**, they are implemented separately in this module. […] this has the added advantage of making it easier to track how many quirks are out there.
>
> The implementation of the quirks at this point is just a single function which applies quirks to the DIDL-Lite resources, **with the options of adding one that applies them to DIDL-Lite objects.**

That last clause reads as an open invitation to add an `apply_object_quirks(element)` next to the existing `apply_resource_quirks(element)`. As written, the PR inlines the quirk in `data_structures.py`, which is the specific outcome that module was created to avoid.

Worth copying one more thing from the existing quirk: `apply_resource_quirks` emits `_LOG.debug("Resource quirk applied for missing protocolInfo, ...")`. This change is silent, so anyone whose metadata is quietly getting `parent_id=""` has nothing to grep for when they go looking.

If you'd rather keep it inline, could you at least expand the comment to match the surrounding style? The neighbouring deviations name the offending service. `# Relax strict checking for non-compliant DIDL metadata streams` doesn't record *which* stream or *why* — the `enqueued_transport_uri_meta_data` detail from your description is exactly what a future reader will want, and it belongs in the code.

## Test coverage

This is the gap I'd most like closed — there are no tests in the PR. `from_element` is well covered by `TestDidlObject` in `tests/test_new_datastructures.py`, and the case is only a few lines: an `<item>` carrying `id` and `<upnp:class>` but no `parentID`, asserting `parent_id == ""` rather than a raise. Without it there's nothing to stop the strict check being reinstated by a later change.

## Question: why `parentID` but not `id`?

Two lines above the change:

```python
item_id = element.get("id", None)
if item_id is None:
    raise DIDLMetadataError("Missing id attribute")
```

If the real-world `enqueued_transport_uri_meta_data` payload omits `parentID`, it seems plausible it might omit `id` too — in which case this only half-fixes the case you hit, and the next person gets `Missing id attribute` instead. Could you post the actual offending DIDL from the event? If `id` is reliably present then the asymmetry is fine, but it's worth a comment saying that's deliberate.

## One risk worth recording

`to_element()` writes `"parentID": self.parent_id` unconditionally (`data_structures.py:756`), and `to_didl_string()` output does get sent to speakers — `core.py:2381` (`add_to_queue`), `2405`, `2415`, `2652`. So a quirked object that gets re-serialised emits `parentID=""`, which is still non-compliant, just differently. Compare `ms_data_structures.py:259-261`, which handles it the other way round:

```python
# Only add the parent_id if we have it
if self.parent_id:
    item_attrib["parentID"] = self.parent_id
```

I went looking for a path where an event-sourced object gets re-serialised and sent back, and couldn't find one — `snapshot.py:142` stores `media_info["CurrentURIMetaData"]` as a raw string and never parses it to a `DidlObject`, so restore is unaffected. I think this is theoretical rather than live, but it's the first thing I'd check if anything odd turns up later.

## Minor

Wherever the code ends up, the three-line form collapses to one — `.get` with a default already covers the `None` case, and `really_unicode("")` is a no-op:

```python
parent_id = really_unicode(element.get("parentID", ""))
```

---

🤖 Review assisted by [Claude Code](https://claude.com/claude-code)

PR #1006, opened 2026-07-29
Prevent unhandled exception when processing favorite missing DIDL class

PeteRager:
When a favorite is returned from the speaker that does not have a DIDL class; the library was raising an unhandled type exception. The change checks for this condition and raises a DIDLMetadataError

https://github.com/home-assistant/core/issues/177460

```
  File "/usr/src/homeassistant/homeassistant/components/sonos/favorites.py", line 133, in update_cache
    if fav.reference.resources:

  File "/usr/local/lib/python3.14/site-packages/soco/data_structures.py", line 987, in reference
    ref = _FROM_DIDL_STRING_FUNCTION(getattr(self, "resource_meta_data"))[0]

  File "/usr/local/lib/python3.14/site-packages/soco/data_structures.py", line 79, in didl_class_to_soco_class
    if separator in didl_class:

TypeError: argument of type 'NoneType' is not a container or iterable
```

PR #1007, opened 2026-07-29
Guard DidlFavorite.reference against unusable resMD metadata

pwt:
Follow-up to #1006.

That PR converted a `TypeError` into a `DIDLMetadataError` when a favorite's DIDL item had no `<upnp:class>`. Two sibling paths in `DidlFavorite.reference` were still unguarded, and both fail the same way for the same users:

```python
ref = _FROM_DIDL_STRING_FUNCTION(getattr(self, "resource_meta_data"))[0]
```

| Case | Input | Exception before this PR |
|---|---|---|
| 1 | Favorite with no `<r:resMD>` element | `AttributeError: 'DidlFavorite' object has no attribute 'resource_meta_data'` |
| 2 | `resMD` present, but parsing to zero DIDL items | `IndexError: list index out of range` |

Case 1 is reachable because `resource_meta_data` is an *optional* entry in `DidlFavorite._translation`, and `DidlObject.from_element` only sets attributes for elements actually present in the XML.

Case 2 is reachable because `from_didl_string` parses with `ET.XMLParser(recover=True)`, which silently discards malformed or truncated content — a damaged `resMD` yields an empty tree rather than a parse error.

Neither `AttributeError` nor `IndexError` subclasses `SoCoException`. Home Assistant's `sonos/favorites.py::update_cache` wraps the reference access in `except SoCoException`, so both escape it and one malformed favorite aborts the entire favorites cache update — the same user-visible failure #1006 addressed, one layer up.

## Change

Both cases now raise `DIDLMetadataError`, naming the offending favorite so a user with one bad favorite among many can identify it:

```
Favorite 'Broken Fav' has no resource_meta_data (resMD) element
Favorite 'Broken Fav' has resource_meta_data containing no DIDL item
```

Verified against HA's exact call pattern — both are now caught by its existing handler and the favorite is skipped and logged, with no change needed downstream.

The emptiness check deliberately lives in `reference` rather than in `from_didl_string`: `music_library.browse()` legitimately returns `[]` for an empty container, so raising there would break normal operation. The "exactly one item expected" assumption belongs to `reference`.

Returning `None` was considered and rejected — callers do `fav.reference.resources`, so it would relocate the unhandled crash rather than fix it.

## Tests

`DidlFavorite.reference` had no test coverage. The new `TestDidlFavorite` adds four tests covering the contract, not just the bug:

- missing `resMD` raises `DIDLMetadataError`
- `resMD` parsing to zero items raises `DIDLMetadataError`
- happy path returns the referenced object with the favorite's `resources` copied across
- the `reference` setter round-trips

The first two were confirmed to fail with the `AttributeError` / `IndexError` above before the fix was written. The last two pass either way and are there as regression guards.

## Out of scope

- The `reference` **setter** and its unguarded `value.resources` access — a caller-error path, not a malformed-device-data path.
- Missing `Raises:` docstrings on `from_didl_string` / `didl_class_to_soco_class`.

## Checks

`pytest`: 266 passed, 67 skipped · `black --check soco tests`: clean · `flake8 soco`: clean · `pylint soco`: 10.00/10

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/soco/data_structures.py
+++ b/soco/data_structures.py
@@ -975,7 +975,12 @@ class DidlFavorite(DidlItem):
     # object available via the 'reference' property.
     @property
     def reference(self):
-        """The Didl object this favorite refers to."""
+        """The Didl object this favorite refers to.
+
+        Raises:
+            DIDLMetadataError: If the favorite has no ``resMD`` element, or if
+                its ``resMD`` element holds no DIDL item.
+        """
 
         # Import from_didl_string if it isn't present already. The import
         # happens here because it would cause cyclic import errors if the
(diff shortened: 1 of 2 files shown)


PeteRager:
excellent - thank you!

Issue #557, opened 2018-01-06
META ISSUE Music service authentication issue


KennethNielsen:
As explained by @lawrenceakka in #428:

> Sonos has changed the way in which music service account data can be accessed. It used to be possible to get the necessary info from the players, but it is now encrypted using AES-CBC. Although other third party Sonos controllers have worked out how to get round this and decrypt the info (see here for a discussion), I am not sure that it would be right for us to do so in an open source project (though I haven't decided yet).
>
> BUT:
>
> I have today managed to successfully register a python soco script with Google Play Music. The script requires user interaction (you have to go to a web page and input your Google Play credentials). It may therefore be possible to create a workaround where you run a registration script and then use the returned details to allow SoCo to use third party media services. I suppose the details will change from time to time so it might be necessary to run the registration script again.
>
> Research continues ....

This issue has resulted in a number of issue raised: #433, #428, #446 and #505. I have tried to compile a list of affected music services from the issues:

 * Google Play Music
 * Apple Music
 * Amazon Music
 * Spotify

I have created this meta issue for the problem, to make it easier to gather information in one place. Please, keep this issue clear of "Me too" comments, and just use the +1 (thumbs up) feature on this comment. I will close all the other issues, but of course keep them listed here for information search purposes. If you have more music services for the list, comment below.

DPH:
I also use Napster, but not sure how many others do. Suggest we add to list but lower priority.

jimbofreedman:
Depending on your use case, the beta Connect endpoint on the Web API may help for workarounds in the interim - I'll be experimenting with it this afternoon: https://developer.spotify.com/web-api/web-api-connect-endpoint-reference/

olivierjc:
I have the problem with Deezer, too.

pfischi:
TuneIn service is also not working.

ghcs27:
@pfischi That's strange, it was specifically mentioned as still working [in #433](https://github.com/SoCo/SoCo/issues/433#issuecomment-250156052).

pfischi:
Can't confirm that. The function call 'my_service = MusicService("TuneIn")' responds with an empty result.

Greetings,

pfischi

jimbofreedman:
[...]

PR #1010, opened 2026-08-11
Add modern music service browsing and account management

BookCatKid:
## Summary

This PR introduces a new implementation for working with Sonos music services, primarily targeting the newer APIs and authentication mechanisms used by current Sonos controllers.

The existing `MusicService` API is kept intact. The new functionality is provided alongside it through `MusicServiceBrowser`, `MusicServiceAccountManager`, and `ConfiguredMusicServiceAccount`.

The main goal is to make authenticated music services such as Apple Music, Amazon Music, Spotify, etc. usable again without requiring users to manually extract or provide authentication tokens.

This is still a draft and I would especially appreciate testing against other Sonos households/services and feedback on the API design.

Related to #557.

### `MusicServiceBrowser`

Adds a new account-aware music-service browsing API which can:

* Automatically retrieve the credentials already stored in the Sonos household
* Browse music-service containers with paging
* Search services, including services exposing multiple search variants
* Retrieve media and extended metadata
* Retrieve playable media URIs
* Generate Sonos URIs from service item IDs
* Read service manifests and presentation maps
* Parse additional presentation-map data such as `NowPlayingRatings` and `QuickSkips`
* Support sorting where exposed by the service
* Query last-update information
* Handle both traditional SMAPI services and newer manifest/content-home-page based services

Credential refresh is enabled by default when supported.

A fair amount of compatibility/error handling is also included for provider-specific behavior and malformed/non-standard responses.

### Account discovery

`ConfiguredMusicServiceAccount` can inspect the accounts currently configured in a Sonos household.

This is also used by `MusicServiceBrowser` so callers normally do not need to manually provide credentials. A particular account can still be selected when multiple accounts for the same service are configured.

### Account onboarding and management

`MusicServiceAccountManager` implements the account-management flow used by Sonos and can:

* Add DeviceLink/AppLink accounts
* Add username/password or anonymous accounts where supported
* Re-link existing accounts
* Rename accounts
* Remove accounts
* Refresh account credentials

The authorization flow is deliberately split so beginning an OAuth/device-link operation does not change the household until `commit_link()` is called.

For example:

```python
from soco.music_services import MusicServiceAccountManager

manager = MusicServiceAccountManager("Spotify")

link = manager.begin_link()

# User visits link.registration_url and completes authorization.

account = manager.commit_link(link)
```

### Existing `MusicService`

The existing `MusicService` API remains available and its existing browsing behavior has not been replaced.

A few generally useful SMAPI operations discovered while implementing the new browser have also been added to it, including browse sorting, `get_scroll_indices`, and `set_played_seconds`.

The old authentication flow is deprecated in favor of `MusicServiceAccountManager`.

## Example / testing application

This PR also adds `examples/music_service_explorer`, a small Flask application intended to expose and test the music-service APIs against a real Sonos household.

It can be used to:

* Browse using either the existing or new API
* Search services and individual search variants
* Inspect configured accounts
* Inspect service descriptors, manifests and presentation maps
* Inspect item and extended metadata
* Retrieve stream/media URIs
* Test account onboarding, removal and renaming

Most panels also show the equivalent SoCo Python code, so it should hopefully be useful as both a testing tool and an example of the new APIs.

## Implementation notes

Sonos music-service account information is encrypted when retrieved from the household. This implementation decrypts that data locally in order to use the credentials already configured on the system.

This requires either the Python `cryptography` package or an available `openssl` executable.

No playback behavior is added to `MusicServiceBrowser` itself. It handles discovery, browsing, searching and metadata; playback/queue manipulation continues to use the existing SoCo APIs.

## Testing

Tests have been added for the browser, account onboarding/management, music-service data structures, and related changes to the existing music-service implementation.

Real-world testing would be particularly useful because music-service providers vary substantially in their SMAPI implementations and authentication behavior.

I have primarily been testing this against my own Sonos system, so **testing with other households, account configurations, and music services would be extremely appreciated.**

## Feedback wanted

Since this is a fairly large addition and is still a draft, feedback is especially welcome on:

* The public API design/naming
* How this should coexist with or eventually relate to the existing `MusicService` API
* Account onboarding/management behavior
* Provider compatibility
* Anything that should be split into separate PRs before this is ready to merge


goldyfruit (commented):
Thanks for taking this on — the side-by-side API design, explicit account selection, staged account linking, and player-resolved Sonos URIs all look like strong foundations for restoring modern music-service support.

I exercised the new/affected tests from the source tree on Python 3.14.6: all 136 passed. I also built and installed the wheel and found a release-blocking packaging problem:

- `soco/music_services/__init__.py` imports `.browser`.
- `[tool.setuptools]` explicitly packages only `soco`, `soco.plugins`, and `soco.music_services`.
- Consequently, the built wheel omits `soco/music_services/browser/**` and fails on import.

Reproduction from commit `618cc64d9e884e6838f135f05d59a909c5939126`:

```console
python -m build --wheel
python -m pip install --no-deps dist/soco-0.32.0.dev0-py3-none-any.whl
python -c "import soco.music_services"
# ModuleNotFoundError: No module named 'soco.music_services.browser'
```

Please include the new subpackage—preferably through setuptools package discovery, for example by including `soco*`—and add CI that builds the wheel, installs it into a clean environment, and imports every public package. Source-tree tests cannot detect this failure.

A few additional suggestions before this leaves draft:

- Consider declaring `cryptography` as a documented optional extra and testing both the Python implementation and OpenSSL fallback. That makes credential-decryption requirements predictable for downstream applications.
- Document the migration boundary clearly: the new result/item models, explicit account selection when a service has multiple accounts, and `sonos_uri_from_id()` playback are intentionally not drop-in equivalents for every legacy `MusicService` call.
- Maintain a real-system compatibility matrix covering provider, auth type, one versus multiple accounts, speaker generation, and tested operation (discover/browse/search/metadata/playback/onboarding). Unit fixtures are valuable, but provider-specific SMAPI behavior means they cannot establish that every advertised service works.
- It would also help to call out prominently that existing Apple Music accounts can be used, while initial Apple linking still requires the Sonos mobile app.

This is a substantial and very useful direction. With the installed-artifact issue fixed and a provider-validation matrix in place, it should be much safer for downstream projects to adopt.

BookCatKid:
Thank you for your review. I will fix the issues/suggestions soon. Have you tested it on your system, does it work properly for you?

edit: Can you test if its fixed now?

goldyfruit:
## Retest report: current head `368af884`

Thanks for the quick follow-up. I retested the new head and also exercised the music-service code against a real Sonos household.

### 1. Packaging fix: verified

The new commit fixes the release-blocking wheel problem from my earlier review.

Tested with Python 3.14.6:

```console
python -m pytest -q \
  tests/test_music_service_browser.py \
  tests/test_music_service_onboarding.py \
  tests/test_music_service_data_structures.py \
  tests/test_musicservices.py
# 136 passed

python -m build --wheel
python -m pip install --no-deps --target /tmp/wheel-test \
  dist/soco-0.32.0.dev0-py3-none-any.whl

PYTHONPATH=/tmp/wheel-test python -c \
  "import soco, soco.plugins, soco.music_services, soco.music_services.browser"
# success
```

I verified that the wheel now contains every `soco/music_services/browser/*.py` module. The package-discovery configuration, `music-services` optional extra, and clean-wheel CI job address the previous feedback well.

### 2. Live-test environment

One Sonos household, six standalone speakers spanning five model families:

| Room | Model | PR Spotify URI | Spotify control | NAS control |
|---|---|---:|---:|---:|
| Cozy | Era 100 | FAIL: UPnP 402 | PASS | PASS |
| Dining Room | Play:5 | FAIL: UPnP 402 | PASS | PASS |
| Library | Play:1 | FAIL: UPnP 402 | PASS | PASS |
| Living Room | Playbar | FAIL: UPnP 402 | PASS | PASS |
| Office | Play:1 | FAIL: UPnP 402 | PASS | PASS |
| Portable | Roam | FAIL: UPnP 402 | PASS | PASS |

All audible checks were capped at volume 3. Before testing I captured volume, mute, transport state, current URI/metadata, and group topology. After every test, all of those values were restored and verified.

The six-model matrix was initially run against `618cc64`. Commit `368af884` changes only the workflow, README, and packaging configuration; it does not change the browser/playback code. I nevertheless reconfirmed the PR URI failure on Office directly from current head `368af884`.

### 3. Configured-account and browse results

`ConfiguredMusicServiceAccount.get_accounts()` worked through the OpenSSL fallback and recovered four configured services without manual tokens:

- Sonos Radio
- Spotify
- Amazon Music
- Plex

Read-only behavior:

- **Sonos Radio:** content-home root and child browsing passed.
- **Spotify:** content-home root, personalized sections, child browsing, track search, and `get_media_metadata()` passed. Returned tracks had `mimeType=audio/x-spotify`.
- **Amazon Music:** SMAPI root and child browsing passed. A specific `tracks / Miles Davis` search returned zero items for this account/tier; I am not classifying that as a PR defect without a provider-side comparison.
- **Plex:** account recovery passed, but Plex returned HTTP 500 saying it could not contact the Plex Media Server and that Remote Access should be enabled. This appears environmental rather than a SoCo failure.

The Sonos music-library baseline also found the configured NAS share and 25,291 indexed tracks. NAS playback worked on every speaker.

### 4. Remaining blocker: Spotify playback URI is invalid on real players

For a Spotify search result such as:

```text
item_id = spotify:track:7azylXFRsebfrIoAtwfjaB
mimeType = audio/x-spotify
service id = 12
account serial = 2
```

`MusicServiceBrowser.sonos_uri_from_id()` returns:

```text
x-sonosapi-stream:spotify%3Atrack%3A7azylXFRsebfrIoAtwfjaB?sid=12&sn=2
```

Minimal reproduction on current head:

```python
accounts = ConfiguredMusicServiceAccount.get_accounts(speaker)
account = next(a for a in accounts if a.service_id == 12)
browser = MusicServiceBrowser("Spotify", account=account, device=speaker)
track = browser.search("tracks", "Miles Davis So What", count=3).items[0]
uri = browser.sonos_uri_from_id(track.item_id)

speaker.play_uri(uri, start=False)
```

Result:

```text
SoCoUPnPException: UPnP Error 402: Invalid Args
```

This same error occurred on the Era 100, Play:5, both Play:1 units, Playbar, and Roam. `start=False` confirms the failure happens in `SetAVTransportURI`, before playback begins.

Additional isolation checks on Office:

- PR URI exactly as returned: UPnP 402.
- PR URI plus `flags=8224`: UPnP 402.
- PR URI plus the legacy `0fffffff` item prefix: UPnP 402.
- PR URI plus full DIDL metadata/account UDN: UPnP 800.
- Legacy `soco://...`: UPnP 714, Illegal MIME-Type.
- `x-sonos-spotify` plus Spotify DIDL metadata: **PLAYING**, with the expected title and artist reported by the player.

The working control resource was equivalent to:

```text
x-sonos-spotify:spotify%3Atrack%3A7azylXFRsebfrIoAtwfjaB?sid=12&flags=8224&sn=2
```

with DIDL containing:

- `object.item.audioItem.musicTrack`
- resource protocol info `sonos.com-spotify:*:audio/x-spotify.*`
- the configured account UDN in `desc`
- track title, artist, and album

That control played successfully on all six models. This isolates the problem from credentials, account serial, item ID, grouping, connectivity, and speaker generation.

### 5. Likely root cause

`x-sonosapi-stream` is a radio/smart-radio transport and is not a universal player-resolved URI for arbitrary music-service tracks. Spotify search results advertise `audio/x-spotify` and require the Spotify transport/resource metadata.

The current method receives only `item_id`, so it has already discarded information needed for correct playback:

- provider MIME type
- item type and `canPlay`
- provider-specific URI scheme/flags
- resource protocol info
- account UDN for DIDL `desc`
- title/artist/album metadata

This makes a universally correct `sonos_uri_from_id(item_id)` difficult to implement. Changing only the scheme for every provider would likely fix Spotify while regressing radio, Amazon, or other SMAPI services.

### 6. Suggested implementation direction

I suggest making playback resolution item-aware rather than ID-only. For example, one of:

```python
resource = browser.playback_resource(item)
# resource.uri
# resource.metadata / queueable DIDL object

speaker.play_uri(resource.uri, meta=resource.metadata)
```

or:

```python
queueable = browser.to_didl_item(item)
speaker.add_to_queue(queueable)
```

Useful model additions would be:

- `MusicServiceBrowseItem.mime_type`
- `MusicServiceBrowseItem.can_play`
- a playback resource or queueable DIDL conversion
- a provider/MIME-aware URI resolver

For Spotify, the resolver can use `x-sonos-spotify` and the `sonos.com-spotify:*:audio/x-spotify.*` protocol information. Radio items can continue using `x-sonosapi-stream`. Other providers should be mapped from their returned MIME/item metadata rather than all sharing a single scheme. The `flags=8224` value worked here, but I would avoid treating that empirical value as universal without checking the relevant provider/item capabilities.

This does not require restoring controller-side `getMediaURI`; player-side resolution can remain the design, but the player needs the correct provider resource and DIDL metadata.

### 7. Regression tests recommended

1. Spotify `audio/x-spotify` track maps to a Spotify resource, not `x-sonosapi-stream`.
2. Generated Spotify DIDL includes the configured account UDN and Spotify protocol info.
3. Radio/station items retain `x-sonosapi-stream` where appropriate.
4. Provider track and radio mappings are separate and MIME/item-aware.
5. A returned playable item can be used with normal SoCo playback or queue APIs without downstream code reconstructing private/raw metadata.
6. Multiple Spotify accounts preserve the selected account serial and account UDN.
7. Installed-wheel test remains in CI.

### 8. Grouping result

Office and Library were grouped at volume 3. Both the working Spotify control resource and a NAS track played on both members. They were then ungrouped and returned to their original sources, volumes, mute states, and transport states. Grouping is therefore not the cause of the PR URI failure; the coordinator rejects that URI before group playback begins.

Overall: the packaging issue is fixed, account recovery and modern browsing are working impressively well on this household, but Spotify playback through the public helper remains a reproducible cross-model blocker.

BookCatKid:
Working on this. Should be done relatively soon.

edit: Sorry but this is turning out to be a lot more complicated than I originally expected, it may take a few days to get this working perfectly. I will update asap. Draft pr after all 🤷. I can't believe I didn't think of testing actual playback lol.

BookCatKid:
The Spotify playback blocker should be fixed now.

`x-sonosapi-stream` was never a valid transport for `audio/x-spotify` tracks,
so `sonos_uri_from_id()` now maps Spotify tracks to `x-sonos-spotify:` with
the `sonos.com-spotify:*:audio/x-spotify:*` protocol info and account UDN in
the DIDL — the exact resource you verified plays on all six models. In
addition, `browser.play()` for Spotify, Pandora and Audible now uses the
desktop controller's DirectControl path (enter the `x-sonos-vli:` session,
then `loadContainer` on the speaker's control API), which is what works on
households where the direct resource is rejected against the AppLink account
UDN.

Could you test playback again on your household? The Plex HTTP 500 and the
empty Amazon "Miles Davis" search look environmental/provider-side, so the
main thing to re-check is Spotify (and, if convenient, the rest of your music services. I added a script to examples that checks all of the playback and searchand browse capabilities in your household, if you share that result here that would be great.).

This pr is getting really complicated and I don't really like the direction its going so more refinement and fixes to come...

goldyfruit:
More than 17000 lines changed!

BookCatKid:
A lot of it is tests and the json file :)

goldyfruit:
## Live retest of `ea8716d1`

I retested the latest head on the same household. The MIME-aware playback work is a substantial improvement, but the new DirectControl selection still leaves an important argument-dependent failure.

### Packaging and unit checks

Python 3.14.6:

```text
196 affected tests passed
```

Covered browser, playback, DirectControl, onboarding, music-service data structures, and legacy music-service tests. I also built the wheel, installed it into a clean target, and successfully imported the new `playback` and `direct_control` modules. No GitHub checks are currently reported on the PR branch.

### Browse/search/metadata verifier

I ran `examples/verify_music_services.py --services configured --no-play` with temporary output files:

| Service | Browse | Search | Metadata | Discovered playable types |
|---|---:|---:|---:|---|
| Spotify | PASS | PASS | PASS (`audio/x-spotify`) | track |
| Amazon Music | PASS | PASS | PASS | program, track |
| Sonos Radio | PASS | PASS | PASS | program |
| Plex | FAIL | FAIL | skipped | none |

Plex still returns its explicit HTTP 500 “Unable to contact the Plex Media Server / enable Remote Access” response, so that remains environmental.

### Spotify: resource fix passes

The new MIME-aware mapping works:

```python
uri = browser.sonos_uri_from_id(item)
speaker.play_uri(uri)
```

The URI is now `x-sonos-spotify:` and reached `PLAYING` on Office at volume 3. Because this call only has the URI, the player reported no title; the helper cannot supply the DIDL that its documentation discusses.

The full regular-resource path also works:

```python
browser.play(item.item_id)
```

This reached `PLAYING` with title `So What` and artist `Miles Davis` on the current head for:

- Play:5
- Play:1 (both tested units through Office/Library coverage)
- Playbar
- Roam

All tests were capped at volume 3 and restored their original URI, metadata, state, mute, volume, and group.

### Remaining blocker: `browser.play(item)` selects a broken path

For the exact same Spotify result:

```python
browser.play(item)          # FAIL
browser.play(item.item_id)  # PASS
```

Passing the `MusicServiceBrowseItem` selects `_play_direct_control()`. Passing its string ID bypasses DirectControl and builds the working `x-sonos-spotify` resource plus DIDL.

`browser.play(item)` failed from a clean state on both an older Play:1 and a current Roam:

```text
SoCoUPnPException: UPnP Error 402: Invalid Args
```

The traceback ends in:

```text
MusicServiceBrowser._play_direct_control()
SoCo.play_direct_control()
SoCo.play_uri()
AVTransport.SetAVTransportURI(...)
```

The failure therefore happens while entering the synthetic `x-sonos-vli:` source, before `wait_for_direct_control()` or `loadContainer()`.

Office had `music_source=WEB_FILE` and no existing DirectControl session before the clean reproduction, so this is not a stale-session conflict.

### Comparison with a real Spotify VLI session

A separate Era 100 already had a real, user-created paused Spotify Connect/VLI session. I inspected it read-only and did not replace its cloud context.

The live session differs from the generated transport:

| Field | Real Spotify session | PR-generated session |
|---|---|---|
| URI label | non-empty 32-character hex session value | empty |
| DIDL `vli` cookie | `1` | `7` |
| DirectControl client | `spotify.connect.adapter` | never becomes active |

Trying arbitrary non-empty labels and cookies `1`/`7` still produced UPnP 402 on Office. That suggests the label is genuine provider/session state, not an opaque caller-selected value. A valid DirectControl session likely cannot be created merely by synthesizing an `x-sonos-vli` URI.

### Recommended behavior

At minimum, Spotify tracks should not unconditionally enter DirectControl when the item object is passed. The ordinary MIME-aware resource is verified working on this household.

Suggested order:

1. Use the regular provider-aware resource for Spotify tracks.
2. Use DirectControl only when a valid matching session already exists, or when there is a verified authenticated mechanism for acquiring the real VLI session state.
3. If DirectControl entry returns UPnP 402 or never becomes active, fall back to regular resource playback.
4. Make `browser.play(item)` and `browser.play(item.item_id)` choose equivalent playback for the same playable track.
5. Consider returning a playback-resource object containing both URI and DIDL; a bare URI cannot carry Spotify title/account/protocol metadata.

I did not exercise `loadContainer()` against the existing Era session because doing so would replace a user-owned paused Spotify container that cannot be perfectly reconstructed afterward.

### Other playback results

On Office at volume 3:

- Amazon Music program: PASS
- Amazon Music track: PASS
- Sonos Radio program: PASS
- Office + Library grouping: PASS with Spotify regular-resource playback; both members reached `PLAYING`

Grouping was then removed and both players were fully restored.

### Verification-script safety

The new verifier is useful, but playback mode needs safety controls before asking users to run it broadly:

- no volume cap/`--volume` option;
- no snapshot and restoration of the prior URI, metadata, state, position, mute, or group;
- calls `speaker.stop()` at the end even if playback was initially active;
- does not stop between item types or services, so one service can keep playing while the next service is browsed or times out.

I recommend a `try/finally` snapshot restore, `--volume` defaulting to a very low level, a short `--play-seconds`, and an explicit stop between playback attempts.

### Verdict

The original Spotify transport blocker is fixed at the resource level, and Amazon/Sonos Radio playback also works here. The public high-level Spotify path is not yet reliable because `browser.play(item)` routes into a VLI session that real players reject, while `browser.play(item.item_id)` succeeds.

goldyfruit:
@BookCatKid any update on this?

BookCatKid:
Thanks for testing this across your setup — the difference you found between play(item) and play(item.item_id) was especially helpful.

I’ve reworked the playback side based on that feedback. The DirectControl/VLI implementation has been removed from this PR for now. It’s a substantially different playback/session mechanism from ordinary music-service URI playback, and trying to handle both through the same high-level play() path was making the API more complicated and unpredictable than it needed to be.

browser.play(item) and browser.play(item.item_id) now both go through the same normal provider-aware resource path, including the Spotify x-sonos-spotify URI, protocol info, and selected account metadata that worked in your live testing.

I also tightened up the real-system verifier based on your safety comments, added explicit coverage for both the cryptography and OpenSSL crypto backends, and removed the old verification results that depended on the DirectControl experiment.

goldyfruit:
## Live retest of `48c96922`

I retested the latest head after removal of the DirectControl/VLI playback path. The main Spotify regression from my previous report is fixed: the public `MusicServiceBrowser.play()` path now works consistently for both a browsed item and its raw item ID.

### Automated checks

- Python 3.11.15: 298 affected tests passed
- Python 3.12.13: 298 affected tests passed
- Python 3.13.14: 298 affected tests passed
- Python 3.14.7: 298 affected tests passed
- Full suite on Python 3.14.7: 408 passed, 67 hardware-only integration tests skipped
- Wheel builds successfully, installs with dependencies into a clean environment, and imports `soco`, `soco.plugins`, `soco.music_services`, and `soco.music_services.browser`
- Black, Flake8 on `soco`, byte compilation, and the Sphinx documentation build pass
- No GitHub checks are currently reported for the PR branch

### Live service results

All audible tests used a maximum volume of 3 and a one-second playback window.

On Office, the verifier passed:

- Spotify: browse, search, metadata, and track playback
- Amazon Music: browse, search, metadata, program playback, and track playback
- Sonos Radio: browse, search, metadata, and program playback
- Plex: browse/search still return the provider's HTTP 500 saying the Plex server cannot be reached or Remote Access is disabled; this is environmental rather than a playback regression

Spotify resource playback also passed with different tracks across these models:

- Play:5 — `So What`
- Play:1 — `Take Five`
- Playbar — `Blue in Green`
- Era 100 — `Human Nature (2022 Remaster)`

A separate live equivalence check used `Teardrop` by Massive Attack:

- `sonos_uri_from_id(item) == sonos_uri_from_id(item.item_id)`
- both resolved to the ordinary `x-sonos-spotify:` resource path
- `browser.play(item.item_id)` reached `PLAYING` and exposed the correct title and artist
- the original Office state was restored exactly

Temporary Office + Library grouping also passed: Office coordinated Spotify playback for the two-member group at volume 3, after which both speakers' original standalone topology, source, transport state, volume, and mute state were restored exactly.

The Roam was already playing an `x-sonos-vli:` Spotify Connect session. The new verifier correctly skipped it as non-restorable and left it unchanged.

### Remaining verifier restoration defect

The new safety work is a substantial improvement, but I found one reproducible false-success case.

Starting state on a standalone Play:1:

```text
transport=STOPPED
media URI=""
metadata=""
volume=14
```

After successful Spotify playback through `test_playback()`:

```text
transport=STOPPED
media URI=x-sonos-spotify:...
metadata=<the tested track>
volume=14
```

`test_playback()` returned `status=ok` with no restore warning even though the original empty transport was not restored. `Snapshot.restore()` has no URI to set in this case, so the tested resource remains loaded while stopped.

I restored the speaker manually with `SetAVTransportURI` using empty URI and metadata, then confirmed the entire household matched its pre-test state.

Suggested fix: when the coordinator snapshot's original `media_uri` is empty, explicitly clear `CurrentURI` and `CurrentURIMetaData` during restoration. Please add verifier-focused tests for empty URI, stopped resource, paused resource, grouped playback, playback failure, and VLI/cloud-queue skip behavior; the verifier safety helpers currently have no direct unit coverage.

### Packaging and CI findings

`twine check` currently fails on the built wheel:

```text
ERROR `long_description` has syntax errors in markup and would not be rendered on PyPI.
Unknown interpreted text role "meth".
```

The raw Sphinx roles in `README.rst` (`:meth:` on the new `play`/`play_uri` references) work in the Sphinx documentation build but not in the PyPI README renderer. Plain RST links or literals would fix this.

The new wheel-import CI check is also vulnerable to source-tree shadowing. Its import command runs from the checked-out repository, so Python can import `./soco` instead of the package installed into `/tmp/wheel-venv`. Running the import after `cd /tmp`, or using an equivalent isolated invocation, will make the check actually validate the wheel. The current wheel itself does contain and cleanly import the browser package when tested outside the checkout.

Finally, `pylint soco` exits nonzero on warnings in the new modules (despite a 9.97/10 score), including missing method docstrings, protected `_request` access, missing exception chaining, and a few style warnings. Since the workflow invokes `pylint soco` directly, these should be resolved or intentionally configured before CI is enabled.

### Verdict

The important live playback issue is resolved: normal Spotify item playback is now reliable across the tested speaker generations, raw IDs behave consistently, metadata is correct, and grouping works. I would address the empty-URI restoration false success plus the README/wheel-CI issues before taking the PR out of draft.

goldyfruit:
## Follow-up retest of `49ae9c6b`

Thanks — this commit resolves most of the findings from my previous report.

### Confirmed fixes

- Empty-source restoration now works live. A standalone Play:1 began with `STOPPED`, an empty URI/metadata, and volume 14; after Spotify playback at volume 3 through the verifier, it returned exactly to the original empty transport and audio state.
- `twine check` now passes for the built wheel.
- The wheel import check now runs outside the checkout, eliminating source-tree shadowing.
- A clean wheel installation imports `soco`, `soco.plugins`, `soco.music_services`, and `soco.music_services.browser` successfully.
- `pylint soco` now passes at 10.00/10.
- The new Spotify example successfully played the expected track with correct title/artist when I ran it inside an external volume/state-restoration wrapper.

### Automated results

- Python 3.11.15: 299 affected tests passed
- Python 3.12.13: 299 affected tests passed
- Python 3.13.14: 299 affected tests passed
- Python 3.14.7 full suite: 409 passed, 67 hardware-only integration tests skipped
- Wheel build, clean installed-wheel import, `twine check`, Flake8, `git diff --check`, and Pylint passed
- No GitHub checks are currently reported for the branch

### Live playback results

All audible tests were capped at volume 3 with short playback windows.

Spotify passed with different tracks on:

- Play:5 — `Dayvan Cowboy`
- Play:1 — `Roygbiv`
- Playbar — `Olson`
- Era 100 — `Memory Death`
- Office — `Never Gonna Give You Up` through the new example

The Roam was already in an active Spotify Connect `x-sonos-vli:` session. The verifier correctly skipped it and left it unchanged.

Temporary Office + Library grouping also passed using `Avril 14th`: Office coordinated playback, both members were capped at volume 3, and the original standalone topology, sources, transport states, volumes, and mute states were restored exactly.

Amazon Music browse/search/metadata plus program and track playback passed. Sonos Radio browse/search/metadata passed; one initial program attempt returned UPnP 501, but the same item and four additional stations all passed immediately afterward, so that appears transient rather than a regression. Plex remains unavailable because its provider reports that the Plex server/Remote Access cannot be reached.

A final household audit confirmed every speaker and group matched its pre-test state.

### Additional configured-service retest after account linking

I repeated the live household sweep after Deezer, Apple Music, Audible, and CBC Radio & Music were added. All audible checks used volume 3 and short playback windows, and a final audit confirmed exact restoration of every source, transport state, volume, mute state, and group.

- **Deezer:** browse, search, metadata, Flow/program playback, and FLAC track playback pass. Live playback passed on Play:5 and Era 100. Grouped Office + Library playback also passed.
- **Apple Music:** browse, search, metadata, and AAC track playback pass. Live playback passed on Playbar.
- **CBC Radio & Music:** browse and metadata pass; the service advertises no search categories. Live `stream` playback passes on Play:1, but support is partial: five different `WebRadio-program:*` items consistently fail with UPnP 701 through the current `program -> x-sonosapi-radio` mapping. Diagnostic attempts using stream, HLS, and HTTP URI variants were also rejected, so these objects need provider-specific investigation rather than a simple generic scheme substitution.
- **Audible:** after authorization completed, the Canadian marketplace (`refmarketplace:ca`) exposes `Le Rouge et le Noir` as an `audiobook`, and playback passes through the generated `x-rincon-cpcontainer:` URI. It passed on Play:5, Play:1, Playbar, Era 100, Office, and an Office + Library group. The active Roam Spotify Connect/VLI session was correctly skipped unchanged.

The Audible verifier result is now:

```text
browse=ok search=ok metadata=error playback={audiobook=OK}
```

Audible still has two provider limitations: exact-title/author searches complete without a protocol error but return zero results, and `getMediaMetadata` for the `reftitle:*` audiobook returns `Server.ServiceUnknownError: unsupported Sonos entity tag reftitle` (HTTP 500). Playback remains successful because `resolve_item()` falls back to the valid type/title already present on the browsed item.

This confirms that Audible playback itself works after account linking. The remaining service-level compatibility gap found in this expanded run is CBC `program` playback; CBC live streams work.


### Remaining actionable findings

#### 1. Black still fails

With the currently resolved development dependencies:

```console
black --check soco tests examples/play_spotify_track.py examples/verify_music_services.py
```

Black reports that these files need reformatting:

- `soco/music_services/browser/catalog.py`
- `soco/music_services/browser/util.py`
- `examples/play_spotify_track.py`

The first two are included by the workflow's `black --check soco tests`, so CI will fail once checks run. (`tests/test_musicservices.py` is also flagged by the latest Black but is unchanged by this commit.)

#### 2. Private `_BrowseSoapFault` can escape and abort verification

During the complete configured-service run, Spotify passed and restored successfully, then Amazon's `refreshAuthToken` returned a transient HTTP 504. The exception escaped as:

```text
soco.music_services.browser.transport._BrowseSoapFault:
HTTP: Unexpected status 504 (HTTP 504)
```

That private exception is not a `MusicServiceException`, so `test_browse()` did not catch it and the entire verifier terminated instead of recording an Amazon error and continuing to Plex/Sonos Radio. Retrying Amazon passed, confirming the 504 itself was transient; the exception leak is still deterministic for that failure path.

The path is `_content_root()` calling `_client.refresh_auth_token()`, whose direct `_request()` can raise `_BrowseSoapFault` without converting it at the public boundary. Please translate it to `MusicServiceException`/`MusicServiceAuthException` as appropriate and add a regression test for a non-auth refresh failure such as HTTP 504. The verifier should also consider a final defensive per-service catch so one provider cannot abort the household matrix.

#### 3. Spotify example safety

`examples/play_spotify_track.py` works, but when run directly it starts playback at the speaker's current volume and leaves that source playing. For a user-facing hardware example, consider a `--volume` option (or at least a prominent warning) and optionally state restoration/short playback duration. It should also report a clear error when no Spotify account is configured instead of allowing `next(...)` to raise `StopIteration`.

### Verdict

The live playback and empty-URI restoration fixes are successful. Once the changed files are formatted and the refresh-token fault is normalized instead of escaping, this will be in substantially better shape for merge.

rahga:
I've reimplemented this PR's `ThirdPartyMediaServersX` decrypt in a separate project (Rust CLI) — it works as described. Observations from two S2 households, both on 18.8 (`97.1-80312`), tested across a number of services and devices:

### What the payload carries that `from_element` doesn't read

Every record also carries `NumAccounts`, `Md0` and `Flags0`.

- **`NumAccounts`** — always `"1"` in 13 of 13 records. A service with two accounts arrives as **two records** with the same encoded type, one per account, each `NumAccounts="1"` — same as the legacy `/status/accounts` format, one `<Account>` per account. The `0` suffix is probably vestigial and your per-record iteration is correct. I walk `0..NumAccounts` anyway (four lines, stops on an index with no `Token`/`SerialNum`) because it makes the question moot; I've never seen it exceed 1. *I'm still testing this inference.*
- **Credential-less records exist**: `UDN="SA_RINCON228871_"`, `SerialNum0="9"`, empty `Username0`, no `Token0`/`Key0` attribute at all. Worth filtering.
- **`Username0`** was `X_#Svc<type>-<key>-Token` in all 13; the `user@example.com` shape in your docs never appeared here (presumably username/password services — I have none). Anything splitting the synthetic form needs a fallback to the whole string.

### The legacy format, flattened

| `/status/accounts` | blob |
|---|---|
| `UN` / `MD` / `NN` / `Key` | `Username0` / `Md0` / `Nickname0` / `Key0` |
| `SerialNum` (attr) | `SerialNum0` |
| `Type` (attr) | encoded type in `SA_RINCON<type>_` |

`Token0`, `Flags0`, `Tier0` are new; `OADevID` has no counterpart. `Md0` was empty in every record.

### `Flags0`, 13 for 13

| `Flags0` | `Username0` | seen on |
|---|---|---|
| `4` | `X_#Svc<type>-<hex>-Token` | Saavn, Qobuz, YouTube Music, iHeartRadio ×2, Deezer (household A) |
| `0` | `X_#Svc<type>-0-Token` or empty | Sonos Radio, TIDAL, Deezer (household B), the credential-less record |

Deezer is on both rows — not a property of the service. Still researching.

### `Username0`'s key is what the app names a second account by

```
Username0="X_#Svc1543-81dee58d-Token"   Nickname0="iHeartRadio"
Username0="X_#Svc1543-885ebbcc-Token"   Nickname0="iHeartRadio 885ebbcc"
```

Same household, two iHeartRadio accounts. Handy for telling an app-generated nickname from a user-renamed one. The key is **not** secret: it's the cdudn in every favorite from that account, in plaintext over ContentDirectory — it has to be, the player resolves content with it.

### `SerialNum0` is the household counter `match` answers in

Same `sn_N` namespace as `musicServiceAccounts:1 match` returns after a device link — useful for keying an account reached both ways once. Household-wide (a new service lands at the next number, not 1); allocation is **highest live serial + 1**, so it *falls* when the top account is removed and means nothing across time. Observed sparse: `1,9,10,11,13,14,15` and `10,23,24,25,26,27`.

### The event subscription is the only route (I tried to avoid the callback)

- `ThirdPartyMediaServersX` is in **1 of 14** SCPDs (`ZoneGroupTopology1.xml`), `sendEvents="yes"`.
- `GetZoneGroupState` and `GetZoneGroupAttributes` — the service's only readers — don't carry it.
- `QueryStateVariable` → UPnP `401`.

### `/status/accounts` is an empty envelope on 18.8

111 bytes of `<ZPSupportInfo></ZPSupportInfo>` — as are `topology`, `perf`, `upnp`; `zp` (1547 B) and `ifconfig` (2600 B) still serve. A batch of diagnostics was emptied; this decrypt is the only route.

### Two services not in this thread

**Qobuz (sid 31)** — `getDeviceAuthToken` refuses every device-link attempt; the stored token decrypts and search/browse work. This mechanism is the only way in.

**YouTube Music (sid 284)** — the token isn't the obstacle, and I chased down what is. The stored token decrypts cleanly and `sendRequest` still answers `403 SERVICE_DISABLED` for **"YouTube Music API (Partner)"**, method **`google.music.sonos.v1.Sonos.SendRequest`** — a partner API keyed to *Sonos's own Google Cloud project*. A self-service API key hits that gate; so does an OAuth token from a self-service client (with full `auth/youtube` scope — `youtube.readonly` returns `SCOPE_INSUFFICIENT` first, so scope isn't the wall, the project is). Only a *user* identity (a login cookie) clears the project gate, and then bounces off Google's cross-domain check. The API isn't in the Cloud Library and no non-Sonos project can enable it — control, same key/project: YouTube Data API v3 → `200`. So a `403` after a clean decrypt is Google's project gate, not a broken decrypt — worth a note, since it reads exactly like one.

### Two checks for other testers

- **Control the decrypt** against tokens you already obtained through the ordinary device-link flow: three services here were byte-identical to what was on disk.
- **Tokens rotate.** Same household, 15 hours apart, no re-linking: five of six identical, YouTube Music's changed *at the same length*. Re-read rather than cache; length doesn't detect a change.

My implementation is at https://github.com/rahga/x2rock (0BSD), which started as an attempt at Sonos control from Android TV with YTM search. The decrypt is a pure function with synthetic test vectors, so nothing from a real household reaches CI.

goldyfruit:
@BookCatKid any update on your work?

BookCatKid:
Thanks for the detailed findings, @rahga! Glad to hear the decryption works independently across both households.
Also, thanks for following up, @goldyfruit. I thought everything was finished after the last round of fixes and testing. Is there anything else you need me to change before this is ready to merge/mark as ready for review?

goldyfruit:
> Thanks for the detailed findings, @rahga! Glad to hear the decryption works independently across both households.
> Also, thanks for following up, @goldyfruit. I thought everything was finished after the last round of fixes and testing. Is there anything else you need me to change before this is ready to merge/mark as ready for review?

Yeah few items still remains, check my last (Codex) review for all the details.

Thanks again.

Issue #975, opened 2025-02-23
Groups not updating on zoneGroupTopology event


jadencodes:
I have been struggling with this issue for a few days now. I use eventlet as my events package since I am working on using this in a flask app and it works great for AV and rendering events. However, when I listen to zoneGroupTopology events, I get them, but no matter what I do the `groups` object doesn't update and will always return the original groups from when the code was started.

For example:
```python
class Foo:
	def zone_group_event(self, event: Event):
		print(f"new group event: {vars(event)}")
		device = soco.discover().pop()
		print(f"STATE: {device.zone_group_state}")
		print(f"STATE: {device.all_groups}")
		print(f"STATE: {device.all_zones}")

	def run(self):
		print(f"Starting group manager listener")
		# subscribe to zone change events
		self.device = soco.discover().pop()
		self.zone_group_sub = self.device.zoneGroupTopology.subscribe(auto_renew=True)
		self.zone_group_sub.callback = self.zone_group_event

		try:
			while True:
				time.sleep(1)
		except Exception as e:
			print(f"Exception in {__name__}: {e}")
			pass
```

I have read the code for `zonegroupstate` and from my understanding it should be keeping a single object up to date that has all the groups. I even tried things such as:

```python
device.zone_group_state.clear_cache()
```

I am grouping/ungrouping from my phone and I see the events come through as expected.

And still nothing. The print statements always print out the same thing.

Any guidance would be appreciated!

PR #1011, opened 2026-08-11
Fix ZoneGroupState not updating on zone_group_state events (#975)

BookCatKid:
The shared ZoneGroupState was only refreshed by explicit polls, so a zone_group_state event arriving via a ZoneGroupTopology subscription never updated it: callbacks reading SoCo.all_groups from inside an event handler kept seeing stale groups.

Route zone_group_state events into the shared ZoneGroupState through the documented Service._update_cache_on_event() hook, which every event backend (events, events_asyncio, events_twisted) calls before delivering an event:

* ZoneGroupTopology now overrides _update_cache_on_event() to feed the payload to ZoneGroupState.process_payload(), so subscribers see the fresh topology as soon as the event is delivered. A malformed payload is logged and never blocks delivery.
* The asyncio-specific inline ZGS handling is removed: all backends now share the hook instead of double-processing events.

ZoneGroupState is now written from event-handler threads as well as from polls, so make it concurrency-safe:

* The whole process_payload() update (parse, deduplicate, mutate) runs under a single RLock acquisition; an RLock is required because update_soco_instances() re-enters via clear_zone_groups().
* get_groups()/get_all_zones()/get_visible_zones() return snapshots taken under the lock, and SoCo.all_groups/all_zones/visible_zones delegate to them, so callers never copy the sets mid-update.
* An in-flight poll whose result would supersede event-driven updates is discarded: after GetZoneGroupState() returns, the subscription state is re-checked under the lock and the result dropped if eventing became active while the request was on the wire.

Tests cover the hook, its failure path, end-to-end delivery through the real event handler, reader/writer exclusion, stale-poll discard, and the synchronized public properties. Each test fails without its fix.

### DIFF
--- a/reproduce_issue_975.py
+++ b/reproduce_issue_975.py
@@ -1,0 +1,191 @@
+"""Live verification of the issue #975 fix (branch soco-fixes-1).
+
+Drives the Sonos system itself - no manual interaction needed. It subscribes
+to ZoneGroupTopology events, programmatically joins and unjoins two speakers,
+and verifies that `soco.all_groups` tracks the changes *purely from events*:
+while the subscription is active, the poll path is patched to raise, so any
+topology update visible via `all_groups` must have come from a
+zone_group_state event (via ZoneGroupTopology._update_cache_on_event).
+
+Usage: python reproduce_issue_975.py
+"""
+
+import sys
+import time
+
+from soco import discovery
+from soco.events import event_listener
+
+FAILURES = []
+
+
+def wait_for(predicate, what, timeout=20.0):
+    """Wait up to `timeout` seconds for predicate() to return True."""
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        if predicate():
+            return True
+        time.sleep(0.2)
+    print(f"  TIMEOUT waiting for: {what}")
+    return False
+
+
+def groups_label(groups, device):
+    """Human-readable description of the groups containing `device`."""
+    labels = [f"    {g.label}" for g in groups if device in g]
+    return "\n".join(labels) if labels else "    (not in any group)"
+
+
+def main():
+    print("== Step 1: discover Sonos system ==")
+    devices = discovery.discover()
+    if not devices:
+        print("FAIL: no Sonos devices found on the network")
+        sys.exit(1)
+    print(f"  Found {len(devices)} device(s)")
+
+    # Use the largest household
+    households = {}
+    for d in devices:
+        households.setdefault(d.household_id, []).append(d)
+    household = max(households.values(), key=len)
+    print(f"  Using household with {len(household)} device(s):")
+    for d in household:
+        print(f"    {d.ip_address} {d.player_name}")
+    if len(household) < 2:
+        print("FAIL: need at least 2 speakers in one household to test grouping")
+        sys.exit(1)
+
+    # Populate the shared ZoneGroupState with one ordinary poll
+    zgs = household[0].zone_group_state
+    zgs.clear_cache()
+    baseline = household[0].all_groups
+    print(f"\n== Step 2: baseline topology ({len(baseline)} group(s)) ==")
+    for g in sorted(baseline, key=lambda g: g.label):
+        print(f"    {g.label}")
+
+    # Event source and join target must be usable speakers
+    usable = [
+        d
+        for d in household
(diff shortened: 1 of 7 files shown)


reproduce_issue_975.py
```diff
@@ -0,0 +1,191 @@
+"""Live verification of the issue #975 fix (branch soco-fixes-1).
```
pwt: This is a useful live verification script, but it shouldn't be committed to the repo root (it requires real hardware, mutates the user's grouping, and references a personal branch name). Could you drop it from the PR? The steps/results would be great to keep in the PR description instead.

soco/zonegroupstate.py
```diff
@@ -200,7 +222,20 @@ def poll(self, soco):
...
         try:
             zgs = soco.zoneGroupTopology.GetZoneGroupState()["ZoneGroupState"]
-            self.process_payload(payload=zgs, source="poll", source_ip=soco.ip_address)
+            # A subscription may have become active while the request was in
+            # flight; discard the result so it can't supersede event updates.
+            with self._lock:
+                if self.has_subscriptions:
+                    self.total_requests += 1
+                    _LOG.debug(
+                        "Subscription became active during poll for %s, "
```
pwt: Reasonable: once eventing is live, the subscription's initial NOTIFY carries the full topology. One edge case worth being aware of: if this is the very first poll (sets still empty) and the initial event hasn't been processed yet, the caller gets an empty `all_groups`. Probably acceptable, but it might be worth a comment.

soco/services.py
```diff
@@ -823,6 +823,22 @@ class ZoneGroupTopology(Service):
     """Sonos zone group topology service, for functions relating to network
     topology, diagnostics and updates."""
 
+    def _update_cache_on_event(self, event):
+        """Keep the shared ZoneGroupState in sync with ZGS events (issue #975)."""
```
pwt: Nit: I'd drop the "(issue #975)" references from docstrings/comments here and in the tests — the commit/PR already links the issue. Also consider calling `super()._update_cache_on_event(event)`; the base is a no-op today, but it keeps the override safe if the base ever gains behaviour.

tests/test_zonegroupstate.py
```diff
@@ -0,0 +1,163 @@
...
+    def read_groups():
+        reader_started.set()
+        result.append(zgs.get_groups(None))
+        reader_finished.set()
+
+    reader = threading.Thread(target=read_groups)
+    reader.start()
+    try:
+        assert reader_started.wait(timeout=5), "reader never started"
+        # The reader is executing; it must not be able to finish.
```
pwt: Nice use of events and `finally` to avoid hung threads. Note these short "still blocked" windows (0.1s here, 0.2s at line 137) can only produce false *passes* on a slow runner, never false failures — so fine as-is, just flagging that they're a weaker guarantee than they look.

pwt (commented):
Thanks for this — the fix is correct and the approach is the right one. Routing `zone_group_state` through `ZoneGroupTopology._update_cache_on_event()` fixes the real bug for the threaded `events` backend (an active ZGT subscription made `poll()` short-circuit, but nothing ever refreshed the state), and it removes the asyncio-only special case. The locking also looks sound: the RLock is needed for the `clear_zone_groups()` re-entry, it's never held across the `GetZoneGroupState()` network call, and constructing SoCo instances under it doesn't create a lock-order inversion with the singleton metaclass.

I've pushed two commits to your branch: a merge of `master` (resolving an additive conflict in `tests/test_events.py` with the `get_listen_ip` tests from #1012) and a `black` fix to `tests/test_zonegroupstate.py`.

Remaining requests:

1. **Please remove `reproduce_issue_975.py`** from the repo root (see inline comment).
2. **Redundant `process_payload()` in `update_zgs_by_event_default()`** (`soco/zonegroupstate.py`, ~line 306). With the new hook, the event received there has already been processed by the handler before it reaches `sub.events`, so the explicit call is now a no-op thanks to dedup. Suggest dropping it so it matches the asyncio fallback, which already relies on the handler. The docstring on `update_zgs_by_event_asyncio` ("the events_asyncio notify handler will call 'process_payload'") should also now refer to the `_update_cache_on_event` hook.

Minor / optional:

3. The docstring on the base `Service._update_cache_on_event` warns implementers that they must not touch shared state without locks. The override satisfies that via the ZoneGroupState lock — a brief note there (or in the override's docstring) would help future readers.
4. Small nits inline.

Issue #978, opened 2025-03-16
get_listen_ip uses a blocking socket with asyncio


bdraco:
```
2025-03-15 23:56:51.679 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved (None)
Traceback (most recent call last):
  File "/Users/bdraco/home-assistant/homeassistant/components/sonos/__init__.py", line 502, in _async_handle_discovery_message
    await self.async_subscribe_to_zone_updates(discovered_ip)
  File "/Users/bdraco/home-assistant/homeassistant/components/sonos/__init__.py", line 227, in async_subscribe_to_zone_updates
    sub = await soco.zoneGroupTopology.subscribe()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bdraco/home-assistant/venv/lib/python3.13/site-packages/soco/events_asyncio.py", line 373, in _async_wrap_subscribe
    await self.event_listener.async_start(self.service.soco)
  File "/Users/bdraco/home-assistant/venv/lib/python3.13/site-packages/soco/events_asyncio.py", line 209, in async_start
    ip_address = get_listen_ip(any_zone.ip_address)
  File "/Users/bdraco/home-assistant/venv/lib/python3.13/site-packages/soco/events_base.py", line 776, in get_listen_ip
    sock.connect((ip_address, config.EVENT_LISTENER_PORT))
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bdraco/home-assistant/venv/lib/python3.13/site-packages/blockbuster/blockbuster.py", line 109, in wrapper
    raise BlockingError(func_name)
blockbuster.blockbuster.BlockingError: Blocking call to socket.socket.connect
```

Since its closed right away its likely fine to setblocking for False for all use cases
```diff
diff --git a/soco/events_base.py b/soco/events_base.py
index 8c7e6a6..805e102 100644
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -772,6 +772,7 @@ def get_listen_ip(ip_address):
     if config.EVENT_LISTENER_IP:
         return config.EVENT_LISTENER_IP
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    sock.setblocking(False)
     try:
         sock.connect((ip_address, config.EVENT_LISTENER_PORT))
[...]

PR #1012, opened 2026-08-11
Use a non-blocking socket in get_listen_ip() (#978)

BookCatKid:
get_listen_ip() called sock.connect() on a blocking UDP socket, which can stall (e.g. while waiting on ARP resolution) and is unsafe to call from asyncio event loops - it broke the asyncio event backend inside Home Assistant.

Set the socket non-blocking before connecting; the local address is assigned either way, so getsockname() still returns the correct interface IP even if the connect is left in progress (EINPROGRESS), which is handled explicitly. Genuine OSErrors still return None.

Tests cover the non-blocking connect (including that setblocking(False) is used), the EINPROGRESS path, and the failure path; each fails without its part of the fix.

### DIFF
--- a/soco/events_base.py
+++ b/soco/events_base.py
@@ -772,7 +772,12 @@ def get_listen_ip(ip_address):
         return config.EVENT_LISTENER_IP
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     try:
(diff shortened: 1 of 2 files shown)


pwt:
Thanks for the contribution, @BookCatKid

Issue #621, opened 2018-06-07
occasional snapshot() error with "cross_fade" property


rudyvan:
I sometimes have and error when calling snapshot().   On 40 calls a day the error appears less than 2 times.  I am using version 0.14

Maybe the error comes from other things my script does, but the code leading up to the error is really streightforward:

```
zones = list(soco.discover())
for zone in zones:
	resume[zone.ip_address]=SoCoCam(zone)
	if resume[zone.ip_address].snapshot(): # it returns True when the snapshot zone is a coordinator
		coordinators.append(zone); volumes[zone.ip_address]=zone.volume
		try: zone.pause() # can not pause TV, so could throw off an error
		except: pass
		if not zone.is_playing_tv: zone.volume=self.volume
	        else: memberships[zone.ip_address]=zone.group.coordinator
```

error:

```
Traceback (most recent call last):
  File "/home/pi/voice.py", line 123, in sonos_thread
    if resume[zone.ip_address].snapshot(): # it returns True when the snapshot zone is a coordinator
  File "/usr/local/lib/python3.5/dist-packages/soco/snapshot.py", line 130, in snapshot
    self.cross_fade = self.device.cross_fade
  File "/usr/local/lib/python3.5/dist-packages/soco/core.py", line 104, in inner_function
    raise SoCoSlaveException(message)
soco.exceptions.SoCoSlaveException: The method or property "cross_fade" can only be called/used on the coordinator in a group
```

Any reason this could happen?
Thanks in advance for feedback!

DPH:
Hi @rudyvan 
useful to know there are some occasional issues on long running processes. 
Is this the same error we had back in #519?
We did a fix (although not sure it really made a difference - but used SoCo functions) in #529 

Somehow we are identifying a player as coordinator when it is not. Either it is stale information or a logic problem somewhere.

Will try to have a look at this, but am not here much for the next couple of weeks.
Shout if you have any brain waves, or if anyone else has ideas.

Regards Daivd

rudyvan:
Indeed, is the same issue, it did not get away.

Kind Regards,
Rudy

ghcs27:
It seems like device.cross_fade is queried even if the device is not a coordinator, but the getter raises an exception in that case.

PR #1013, opened 2026-08-11
Don't read cross_fade in snapshot() for non-coordinators (#621)

BookCatKid:
Snapshot.snapshot() reads device.cross_fade whenever the device plays from its local queue, but the getter is decorated @only_on_master: if the ZoneGroupState is stale (e.g. the device just became a coordinator after a group change), it raises SoCoSlaveException and the snapshot crashes.

Only read cross_fade when the device is a coordinator, using the flag already captured at the top of snapshot(); the restore path was already gated on the same flag. Tests cover both the coordinator and non-coordinator paths; the non-coordinator test fails with the exact SoCoSlaveException from the issue if the guard is removed.

### DIFF
--- a/soco/snapshot.py
+++ b/soco/snapshot.py
@@ -127,7 +127,9 @@ class Snapshot:
         if self.is_playing_queue:
             # playing from queue - save repeat, random, cross fade, track, etc.
             self.play_mode = self.device.play_mode
-            self.cross_fade = self.device.cross_fade
+            if self.is_coordinator:
+                # cross_fade is only readable on the coordinator (#621)
+                self.cross_fade = self.device.cross_fade
 
             # Get information about the currently playing track
             track_info = self.device.get_current_track_info()
--- a/tests/test_snapshot.py
+++ b/tests/test_snapshot.py
@@ -1,9 +1,26 @@
 """Tests for the snapshot module."""
 
-from unittest.mock import MagicMock, call
+from unittest import mock
+from unittest.mock import MagicMock, PropertyMock, call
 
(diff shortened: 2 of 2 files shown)


pwt (commented):
Thanks for this, it's a small, correct fix. It also matches the restore side: `self.device.cross_fade = ...` only runs inside `_restore_coordinator()`, which is itself gated on `self.is_coordinator`, so skipping the save for non-coordinators loses nothing. `play_mode` is still read unconditionally, which is fine because it isn't `@only_on_master`.

**Needs fixing**
- `black --check` fails on `tests/test_snapshot.py`: there's an extra blank line after the imports, before `QUEUE_URI`. Running `black tests/test_snapshot.py` fixes it.

**Minor / optional**
- The guard uses the flag captured at the start of `snapshot()`, but `only_on_master` calls `is_coordinator` again, which re-polls the zone group state. If the ZGS cache refreshes in between and the device flips from coordinator to non-coordinator, it would still raise. That window is very narrow and I don't think it blocks this PR, but catching `SoCoSlaveException` around the read would close it if you want. Also, the description's example ("just became a coordinator") doesn't quite match the case the guard handles, which is the ZGS saying *not* coordinator while the transport still reports the local queue.
- Nit: the tests use both `from unittest import mock` and `from unittest.mock import MagicMock, PropertyMock, call`. Importing `patch` alongside the others would be tidier.
- Nit: `make_snapshot_playing_queue` returns a patch context manager rather than a snapshot, so a name like `playing_local_queue(moco)` might read better.

The snapshot tests pass locally and flake8/pylint are clean.

PR #1014, opened 2026-09-06
Support multi-port players such as Sonos Amp Multi

EliSchleifer:
## Problem

Sonos Amp Multi can expose up to four independently controlled virtual players from one physical device. Each player shares the same IP address but advertises a different UPnP port in ZoneGroupState, typically 1400, 1500, 1600, and 1700.

SoCo currently discards the port from each ZoneGroupMember Location, creates singleton instances using only the IP address, and sends all requests to port 1400. As the topology is processed, the virtual players overwrite one another's UUID and zone metadata. Consumers therefore see one incorrectly identified player instead of every Amp Multi zone.

## Change

- Add an optional UPnP port to SoCo, defaulting to 1400.
- Identify singleton instances by IP address and port.
- Preserve non-default ports from ZoneGroupState Location URLs.
- Use the player's base URL for UPnP services, device descriptions, battery status, album art, and account requests.
- Preserve existing construction and representation behavior for normal port-1400 players.

## Tests

The new regression fixture models four virtual players sharing one IP address. It verifies that all four remain distinct by port, UUID, and name and that their base URLs use the correct ports.

Local checks:

- pytest: 270 passed, 67 hardware integration tests skipped
- flake8 soco: clean
- flake8 critical-error scan: clean
- pylint soco: 10.00/10
- black --check soco tests: clean
- Sphinx documentation build: successful

The change was also tested read-only against two physical Sonos Amp Multi units, exposing eight virtual zones in total. SoCo discovered all eight and independently retrieved each zone's device information, volume, mute state, and transport state. The test did not change playback or device state.

## Out of scope

- SSDP still discovers the physical device on its standard endpoint; the additional virtual players are learned from ZoneGroupState.
- Consumer-specific device-registry presentation for several virtual players sharing one physical MAC address.

### DIFF
--- a/soco/core.py
+++ b/soco/core.py
@@ -120,9 +120,14 @@ class _ArgsSingleton(type):
         key = cls._class_group if hasattr(cls, "_class_group") else cls
         if key not in cls._instances:
             cls._instances[key] = {}
-        if args not in cls._instances[key]:
-            cls._instances[key][args] = super().__call__(*args, **kwargs)
-        return cls._instances[key][args]
+        instance_key = (
+            cls._singleton_key(*args, **kwargs)
+            if hasattr(cls, "_singleton_key")
+            else args
+        )
+        if instance_key not in cls._instances[key]:
+            cls._instances[key][instance_key] = super().__call__(*args, **kwargs)
+        return cls._instances[key][instance_key]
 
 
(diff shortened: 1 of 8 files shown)


pwt:
Thanks for this! How does it behave if network scan discovery is used instead of SSDP?

PR #1015, opened 2026-09-19
Add Sonos Beam Ultra as a soundbar

baptiste-devana:
## Summary

The Sonos Beam Ultra (model S59) was missing from `SOUNDBARS`, so the soundbar-only
features silently degraded on that device: `night_mode`, `dialog_mode`, `audio_delay`,
the surround settings and `soundbar_audio_input_format` returned `None` or raised
instead of talking to the speaker.

This adds `"beam ultra"` to `SOUNDBARS` and generalises the Arc Ultra special case so
it covers both Ultra soundbars.

## Details

Like the Arc Ultra, the Beam Ultra exposes a 0-4 `DialogLevel` and a separate
`SpeechEnhanceEnabled` EQ type, rather than the on/off `DialogLevel` of the older
soundbars. That distinction was previously hard-coded to the Arc Ultra alone, so this
introduces a shared predicate instead of duplicating the model check:

- new `ULTRA_SOUNDBARS = (ARC_ULTRA_PRODUCT_NAME, BEAM_ULTRA_PRODUCT_NAME)` and a
  matching `SoCo.is_ultra_soundbar` property;
- the `speech_enhance_enabled` getter and setter now gate on `is_ultra_soundbar`
  instead of `is_arc_ultra_soundbar`, and the `NotSupportedException` message was
  reworded accordingly;
- `is_arc_ultra_soundbar` is left in place unchanged, so any external caller relying
  on it keeps working. It is simply no longer used internally.

The change is additive and backwards compatible: no existing public name is removed or
changes meaning.

## Testing

`tests/test_core.py` gains a case asserting that `speech_enhance_enabled` is served for
a `model_name` ending in "beam ultra", alongside the existing Arc Ultra case.

Ran the full CI check set locally on Python 3.11:

- `black --check soco tests` — clean
- `flake8 soco` — clean
- `pylint soco` — 10.00/10
- `pytest` — 261 passed, 68 skipped (the skips are the integration tests, which need a
  real device, plus `test_events_asyncio.py` for a missing `aiohttp`)

### Verified against real hardware

Checked on an actual Sonos Beam Ultra, with a Sonos Arc (non-Ultra soundbar) and a
Sonos One (not a soundbar) as controls. Read-only: `GetSpeakerInfo` and `GetEQ` only,
nothing was set and playback was not touched.

| | Beam Ultra | Arc (control) | One (control) |
|---|---|---|---|
| `model_name` | `Sonos Beam Ultra` | `Sonos Arc` | `Sonos One` |
| `is_soundbar`, master | **`False`** | `True` | `False` |
| `is_soundbar`, this branch | **`True`** | `True` | `False` |
| `is_ultra_soundbar` | `True` | `False` | `False` |
| `GetEQ DialogLevel` | **`4`** | `0` | 402 Invalid Args |
| `soundbar_audio_input_format` | `'No input'` | `'No input'` | `None` |

Two things this pins down:

- **The bug reproduces on the device.** `"sonos beam ultra"` does not end with `"beam"`,
  it ends with `"beam ultra"`, so on master `is_soundbar` is `False` for this speaker and
  all five soundbar properties return `None`. With this branch they all answer.
- **`DialogLevel` reads back as `4`.** Under the on/off semantics of the older soundbars
  that value is not representable, which is the concrete evidence for treating the Beam
  Ultra like the Arc Ultra here. Note this is an observed value: I did not issue `SetEQ`
  calls, so the full 0-4 range is inferred rather than exercised.

Happy to adjust the naming (`is_ultra_soundbar`, `ULTRA_SOUNDBARS`) if you would rather
keep per-model predicates.


pwt:
Thanks for contributing this @baptiste-devana. I expanded the tests a little.
Will be released shortly in a patch release.

